PR #188, opened 2014-09-30
Refactor Metadata and Relationship docs

chuckbutler:
This was a huge refactoring, which does a few things:

- Strips metadata down to just the bare essentials of what a metadata.yaml is, and why its important.
- Introduces large swaths of changes to the relationship documentation that in my reading takes you from understanding what a relationship is, to how to author one. I kept quite a bit of the old documentation, as it was solid but needed to be re-structured so its introduced incrementally on the same page.

Please pull the branch and run make + look at the source html to really grok the layout and what changes were made (as theres going to be a lot of red/green on this merge)


chuckbutler:
Ask and thou shalt receive... over a month later @marcoceppi :)

PR #216, opened 2015-01-06
Add juju logging settings to troubleshooting for container/kvm creation

chuckbutler:
It's often handy to understand why container/vm creation on a unit fails, and presently there is no way to capture that output that's very apparent to the user. By tweaking the juju logging environment settings
we can capture the TRACE of those commands yielding in more meaningful debugging sessions.


howbazaar:
:+1:

marcoceppi:
I think logging should exist elsewhere possibly its own section. This is a good start though.

chuckbutler:
I agree, I thought about starting the logging section - but its a fairly significant topic and the possibility of me writing all that tonight was close to nil, so I figured this was the best bet. Get it in troubleshooting and move it later. I'll file a bug for that.

chuckbutler:
See issue #217

PR #220, opened 2015-01-13
Actions doc fleshed out based on latest content

binary132:
This PR still needs work in the user docs for Actions.  Currently, the charm author section has been fleshed out and better navigation bits and pieces added.

### DIFF
--- a/src/en/actions.md
+++ b/src/en/actions.md
@@ -1,0 +1,317 @@
+# Juju Actions (DRAFT)
+
+Juju charms can describe actions that users can take on deployed services. 
+Actions are parametrized events that can be addressed and delivered to a 
+service or unit. Parameters for an action take the form of a YAML document,
+limited to things we can validate with JSON Schemas.
+
+The following juju commands are specified:
+
+    actions
+    do
+    wait
+    queue
+    kill
+
+In addition, the `status` command grows the ability to show status for 
+an action.
+
+
+
+
+## User experience
+
+A user can list the actions that a service or unit exports, as well as 
+the actions associated with interfaces to a service:
+
+    $ juju actions mysql
+    backup         benchmark      dump           ps             restart        
+    restore        start          stop           test
+    $ juju actions mysql/2
+    pause          resume         sync
+
+A user can take an action from the command-line passing an optional YAML
+file as parameters.
+
+    $ juju do mysql backup
+    action: UUID
+    result: <text display of yaml result>
+    $
+
+    $ juju do mysql tpcc
+    mysql has no 'tpcc' action
+    $
+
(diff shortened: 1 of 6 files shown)


src/en/actions.md
```diff
@@ -0,0 +1,317 @@
+# Juju Actions (DRAFT)
```
marcoceppi: This seems largely duplicated with charms-actions.md I don't see where this is referenced anywhere in nav or others?

src/en/authors-charm-actions.md
```diff
@@ -0,0 +1,233 @@
...
+ - JSON-Schema defines special keys such as `required` and `additionalProperties`,
+which may be given for the whole action at the same level as `description` and
+`params`, or within nested schemas at the usual level.
+
+ - At this time, the `$schema` key is not supported by Juju, as it may trigger
+resolution of remote objects.
+
+### Example Schema
+
+ > Charm dir
```
marcoceppi: We'd typically use a simple tree output of the charm directory here to keep with current styling

src/en/authors-charm-actions.md
```diff
@@ -0,0 +1,233 @@
...
+          minimum: 0
+          maximum: 9
+  required: [filename]
+  additionalProperties: false
+```
+
+This schema would support a call such as:
+
+```bash
+$ juju action do mysql/0 snapshot filename=out.tar.gz compression.type=gzip
```
marcoceppi: No `$` in executable examples

src/en/charms-actions.md
```diff
@@ -0,0 +1,317 @@
...
+
+
+
+
+## User experience
+
+A user can list the actions that a service or unit exports, as well as 
+the actions associated with interfaces to a service:
+
+    $ juju actions mysql
```
marcoceppi: Command should follow a call and response style, so this would be something like

A user can list the actions that a service exports:

```bash
juju actions mysql
```
which would yield:

```
backup   ...
```

In addition to a service, actions can be unit specific

Executing the following:

```
cmd
```

yields

```
...
```

src/en/charms-actions.md
```diff
@@ -0,0 +1,317 @@
...
+    backup         benchmark      dump           ps             restart        
+    restore        start          stop           test
+    $ juju actions mysql/2
+    pause          resume         sync
+
+A user can take an action from the command-line passing an optional YAML
+file as parameters.
+
+    $ juju do mysql backup
```
marcoceppi: See above, there are examples for longer blocks of commands, I'll find an example.

binary132:
Request additions from @johnweldon in

PR #222, opened 2015-01-14
Updating policy around testing

marcoceppi:
https://lists.ubuntu.com/archives/juju/2015-January/004763.html


castrojo:
+1, though it would be nice to link each mention of the type of test to an example.

evilnick:
examples would be cool. for everything...

PR #224, opened 2015-01-15
reworked link refs to be relative

evilnick:
changed all the applicable references to be relative rather than absolute. This allows for greater portability, removes possible conflicts with languages/versions and compatibility for a printed version.

A few minor reformats included as a bonus.


chuckbutler:
:+1:  LGTM

PR #235, opened 2015-01-28
Joyent secure

evilnick:
changed note in the Joyent config to urge use of an individual key for use with Joyent


chuckbutler:
:+1:  LGTM

PR #238, opened 2015-01-29
old file cleanups

### LINKED ISSUE #225: clear out old files
there seem to be old template files etc which no longer need to be in the repo.

evilnick:
fixes #225


marcoceppi:
I got real nervous when I saw build.py, but build-from-source is the new script. +1

PR #239, opened 2015-01-29
Added 1.21.1 stable release notes.

sinzui:


evilnick:
Looks good, thanks. I will have a proper go through it in the morning - there is a lot there!

PR #263, opened 2015-02-11
howto-privatecloud.md: completed some sentences

mattyw:
These sentences were incomplete, they should now make sense

### DIFF
--- a/src/en/howto-privatecloud.md
+++ b/src/en/howto-privatecloud.md
@@ -264,7 +264,7 @@ Now generate relevant metadata for the tools by running the command:
 
     juju metadata generate-tools -d <tools_dir>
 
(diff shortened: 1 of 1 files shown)


src/en/howto-privatecloud.md
```diff
@@ -264,7 +264,7 @@ Now generate relevant metadata for the tools by running the command:
 
     juju metadata generate-tools -d <tools_dir>
 
-Finally, the contents of  can be uploaded to a location in the Juju metadata
+Finally, the contents of <tools_dir> can be uploaded to a location in the Juju metadata
```
marcoceppi: This won't work. It will be interpreted as HTML. Use a tick to wrap as code blocks

mattyw:
@marcoceppi good catch, thanks!

PR #266, opened 2015-02-13
Support versioning the built docs.

kadams54:


kadams54:
## QA

**Note:** If this PR is merged, we'll need a follow up PR to setup the proper versions in `build.yml`. Right now the *only* version that can be built on this branch is this branch; anything else will throw an error. Due to this limitation, you may need to edit the versions setting in `build.yml` and set it to whatever branch name (i.e., `qa-versioned-docs`) you're using to QA. This change will need to be committed to your local branch to ensure it isn't wiped when the script checks out the specified version.

Two scenarios to test:

1. `make` - Ensure that a straight make runs and generates versioned docs under `htmldocs`.
2. `make watch` - Make sure the watch runs and updates a file when you change it.

evilnick:
cool. this may take a while to test adequately!

mitechie:
@evilnick definitely, we'll do some review/QA on our end but this is part of our goal to support multi-versioned docs and your input and feedback is greatly appreciated. Once this is good to go it's one more branch on the main site to allow this to work on jujucharms.com/docs.

src/en/charms.md
```diff
@@ -1,4 +1,5 @@
 # Introducing Charms
+## Missing Words
```
marcoceppi: I'm confused why this is here?

build.py
```diff
@@ -0,0 +1,229 @@
...
+        '''
+        try:
+            self.repo = Repo('.')
+        except:
+            logging.exception('unable to find git repository')
+        original_ref = self.repo.head.reference
+        try:
+            # Grab the correct version.
+            assert not self.repo.bare
+            assert not self.repo.is_dirty()
```
jaycee: Why asserts? Not saying it's wrong--I just usually wonder if introducing our own exceptions doesn't give us better information when something goes wrong.

We can probably move them out of the try block as well, can't we? An assert that fails will halt the program regardless and we should try to keep as little as possible inside a try block as we can.

jaycee:
QA notes:

I updated the build.yaml to the branch name I checked this out into, and had to commit that change b/c otherwise make complained about unsaved changes. With the commit, a straight `make` worked fine.

Running `make watch` *did* show an attempt at update every time I edited the files, however it failed each time with the unsaved changes message as above. An example:

```
jc@jaas:~/Code/canonical/jujudocs$ make watch
Requires watchdog module to be installed. - to install pip install watchdog.
+ watchmedo shell-command '--patterns=*.md;' --recursive '--command=make doc=$(basename ${watch_src_path})' ../
make[1]: Entering directory `/home/jc/Code/canonical/jujudocs'
PYTHONPATH="/home/jc/Code/canonical/jujudocs/tools/plugins:/home/jc/Code/canonical/jujudocs/tools/plugins/gfm:" ./build.py config-joyent.md
make[1]: Entering directory `/home/jc/Code/canonical/jujudocs'
PYTHONPATH="/home/jc/Code/canonical/jujudocs/tools/plugins:/home/jc/Code/canonical/jujudocs/tools/plugins/gfm:" ./build.py config-joyent.md
make[1]: Entering directory `/home/jc/Code/canonical/jujudocs'
PYTHONPATH="/home/jc/Code/canonical/jujudocs/tools/plugins:/home/jc/Code/canonical/jujudocs/tools/plugins/gfm:" ./build.py config-joyent.md
INFO:root:Building docs for kadams54-versioned-docs
ERROR:root:error while building docs
Traceback (most recent call last):
  File "./build.py", line 48, in build
    assert not self.repo.is_dirty()
AssertionError
INFO:root:skipping checkout; the repo has unsaved changes
make[1]: Leaving directory `/home/jc/Code/canonical/jujudocs'
INFO:root:Building docs for kadams54-versioned-docs
ERROR:root:error while building docs
Traceback (most recent call last):
  File "./build.py", line 48, in build
    assert not self.repo.is_dirty()
AssertionError
INFO:root:Building docs for kadams54-versioned-docs
INFO:root:skipping checkout; the repo has unsaved changes
ERROR:root:error while building docs
Traceback (most recent call last):
  File "./build.py", line 48, in build
    assert not self.repo.is_dirty()
AssertionError
INFO:root:skipping checkout; the repo has unsaved changes
make[1]: Leaving directory `/home/jc/Code/canonical/jujudocs'
make[1]: Leaving directory `/home/jc/Code/canonical/jujudocs'
```

I'm not sure if this is expected behavior or not. If it is, can watchdog be configured to watch for just changes to committed files? This is a lot of noise for the make watch feature otherwise.

src/en/charms.md
kadams54: Me too. I may have put it in there while testing watches or whatnot. I'll remove - good catch!

build.py
kadams54: I agree; there's already validation of repo status (dirty or not) in the `checkout` function, so these asserts aren't really needed here. I'll fix.

kadams54:
After having the weekend away from this code to think about it… having a build script that changes the code around (multiple checkouts, etc.) of it's own repo seems very dangerous to me. I'd rather move the versioning stuff (basically most of build.py) outside of this repo. In the end, what that would mean for this PR is that it would just be the file moves; the old build shell script would be restored. Thoughts?

marcoceppi:
I feel much safer with that option; however, I don't think it should be moved outside the repo I think it'd be nice to keep them in the repo but default action shouldn't be to build all the versions of the doc/checkout stuff, possibly enable that with a flag for production stuff.

bac:
@kadams54 Can you confirm this change gets rid of the top-level directory called 'build'? If not, it needs to to avoid pip pitfalls.

kadams54:
@marcoceppi FWIW, the default behavior has been updated to skip checking out any versions if the repo is dirty; I think that should handle most cases where you don't want to build all the versions. I still think the `build.py` should be moved outside this repo, to avoid the situation where we want to build a version of the docs that doesn't have `build.py`.

@bac Was there ever a top level `build` directory? Nothing in this PR deals with a `build` dir…

bac:
@kadams54 My bad. Wrong project.

fabricematrat:
QA Notes:
Mostly working but when I did undo my changes I did get
ERROR:root:unable to checkout "versioned-docs" due to error
Traceback (most recent call last):
  File "./build.py", line 169, in checkout
    origin.fetch()
  File "/usr/lib/python2.7/dist-packages/git/remote.py", line 593, in fetch
    return self._get_fetch_info_from_stderr(proc, progress or RemoteProgress())
  File "/usr/lib/python2.7/dist-packages/git/remote.py", line 539, in _get_fetch_info_from_stderr
    assert len(fetch_info_lines) == len(fetch_head_info), "len(%s) != len(%s)" % (fetch_head_info, fetch_info_lines)

fabricematrat:
Also I did modify 6 months ago the markdown extension for jujucharms.com and only did the modification in the download cache of the web site juju charms
Today we have modified the way the docs are build, there might be a conflict here.

kadams54:
@fabricematrat Hmm, that makes me really nervous. Is it possible to get the changed extension in this repo as well?

fabricematrat:
@kadams54 yes as soon as the docs are redirected to jujucharms.com  we could but for today I am not sure.

evilnick:
Hey guys, I have been off for a few days, but am looking at this again now. @fabricematrat can we see the code for the extension that has been changed? 

I also think it may be better for the build tool to live somewhere else

kadams54:
After talking with @mitechie and @marcoceppi, I think a separate tool would be best. I'm going to close this PR and get to work on pulling the versioning logic into a Python library/command line tool. The intent is to focus very narrowly on the versioning aspects to maximize flexibility.

PR #269, opened 2015-02-24
Added 1.21.3 release notes.

sinzui:


evilnick:
thanks. I will add a versions page soon and let you know

PR #270, opened 2015-02-25
Fetch multiple versions of the docs into a source tree.

kadams54:


kadams54:
Take two at doing versioned docs. This script is much smaller and simpler. The idea is to *only* create a source tree with multiple versions of the repo. Since each version is created within a gitignored directory, we sidestep the issues created by checking out different versions within the local repo. Theoretically, the existing build tools (e.g., `build-from-source`) could be modified to generate the final HTML from the generated source tree instead of the original. That is, it would use (perhaps with the help of something like `xargs`) `_versioned_docs/X.Y.Z/source` instead of `source`.

kadams54:
**QA**

1. Run `tools/version.sh`. You should see a usage message.
2. Run `tools/version.sh 1.18.0 master`. You should see messages about fetching each version.
3. Verify that each version is present and accounted for in the `_versioned_src` directory.
4. Run `tools/version.sh -d foobar 1.18.0 master`.
5. Verify that each version is present and accounted for in the `foobar` directory.

evilnick:
cool, i will give that a try. I think this is a more promising route for now, I will start testing!

jrwren:
:+1: QA instructions worked exactly.

hatched:
:+1: Thanks for this!

PR #271, opened 2015-02-26
Releases

sinzui:
Hi Nick.

I am co-opting this PR to discuss text and formatting of a page to list current releases and how get them. The goal is a single page that explain the 3 kinds of releases we do, and help the user get the version suited to them. There was a recent change to the ubuntu/juju site that is pointing windows users to
    https://launchpad.net/juju-core/+download
Which is promoting devel releases. I can see 10 or more Windows users got a devel release instead of a working stable release.

1. I think the text has the right information, but needs love.

2. I experimented with several layouts. for the instructions for ubuntu/osx/win. The subheader and list layouts were too loose and repetitive. They were hard to scan. I tried tables which almost works, but I had to fight the Ubunutu CSS. I don't like the table. So I offer additional support for definition lists. I like this. Even if you don't like the layout, you may want to keep definition lists for the glossary. The dashes for the dd elements did not play well with the code blocks, so I added attribute support to remove them.

### DIFF
--- a/htmldocs/css/main.css
+++ b/htmldocs/css/main.css
@@ -551,10 +551,17 @@ hr {
   font-weight: 600;
 }
 
-dl { border: 1px dotted #aaa; padding: 0.5em; background: #eeeeee; } 
-dt { float: left; clear: left; width: 100px; text-align: left; font-weight: bold; color: #000; } 
-dd { margin: 0 0 0 110px; padding: 0 0 0.5em 0; }
+dl {
+  border: 1px dotted #aaa;
+  padding: 0.5em;
+  background: #eeeeee;
+  margin-bottom: .75em;
+} 
+dt { float: left; clear: left; width: 130px; text-align: left; font-weight: bold; color: #000; } 
+dd { margin: 0 0 0 140px; padding: 0 0 0.5em 0; }
 dd:before { content: "- "; }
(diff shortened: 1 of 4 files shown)


mitechie:
Once this releases page is in and landed and part of the jujucharms.com/docs please file a bug on https://github.com/CanonicalLtd/jujucharms.com/issues for us to point the https://jujucharms.com/get-started page to the new releases document vs a direct download link for windows users.

evilnick:
The tables need some work, the default style is rubbish, and we lost our custom style last time the stylesheets got changed.
The definition list is better for this sort of thing, mainly because editing tables with links in is prone to disaster! I will need to fiddle with the styles for these - I don't like OS X Homebrew breaking onto two lines for example. 
And it is a bit of a faff having to use HTML in the middle for the code. I miss the old tab display! I think that may be the long term solution - create a MD plugin/syntax for tabs, but we should go with the definition list for now. If you want to remove the table stuff for now, I will land this and fiddle with it.

sinzui:
Hi Nick.

I switch the DL presentation. I tweaked the CSS for DLs to help with the spacing of this case. I am sure the css needs to be revisited for all DLs.

Note that devel version is listing the 1.22-beta4 version , The announcement of that will happen shortly.

PR #275, opened 2015-02-27
versions support part 2

evilnick:
I messed around somewhat with the versions support from @kadams54 

this adds a further script in tools, and a makefile target which

  * creates a _build directory
  * reads the (new) versions file in master and fetches those branches
  * adds them to the build directory, processes all subfolders for markdown
  * removes non-web-facing files
  * lifts the images from the master branch and puts it top-level in _build so all versions can reference the same graphics
  
essentially, _build should be good to go

@mitechie @fabricematrat  would appreciate you taking a look at this. As j.u.c/docs is now defunct, may be worth putting jujucharms.com templates in for base.tpl and nav.tpl?
@marcoceppi I subsumed the original build script as part of the new one. We may still want to keep the old one though?

As these are all additions, merging this should not effect anything currently munging docs content


evilnick:
P.S. i just tried a clean version of this and got some errors because I didn't have local versions of the branches in versions (thought this was in the version.sh script?). For now if you 'git checkout' the branches in the versions file, it should work (will fix it up properly on Monday)

kadams54:
Yeah, `version.sh` has a `git fetch` command in it that should have prevented this. I'm curious why it did not…

evilnick:
yeah, I don't know git *that* well, but it seems that  ```git pull``` and ```git fetch``` etc only operate on tracked branches, and if you haven't checked them out or tracked them it doesn't add a local copy.
I think adding something like:
```git branch --track $version```
in there somewhere may fix it.

evilnick:
I added a line to track the git branches, so should work without any extra checkouts now

kadams54:
:+1: QA looks good. One suggestion: add git to the `sysdeps` make target. That bit me in the VM I used to QA. I'd like to get this landed as I have a branch elsewhere that'll make use of it.

evilnick:
Cool, well, as we are both happy and as I said, it doesn't interfere with anything else, consider it landed. (i added git also - nice catch)

PR #276, opened 2015-03-02
Remove Charm Quality Rating, we're not using it any more.

castrojo:


evilnick:
I did a quick search and can't find anything that even links to this page anymore, so LGTM

PR #277, opened 2015-03-03
azure-config: storage location nitpick

mattyw:
I've had errors when specifying storage in a different location to the instances, this change brings it inline with the default environments.yaml settings


chuckbutler:
+1 LGTM, thanks @mattyw

PR #278, opened 2015-03-04
hook tools additions for ports

evilnick:
apart from some formatting adjustments - the changes are in the open-port, close-port sections (which now support ranges) and the new opened-ports command


evilnick:
:metal:

PR #295, opened 2015-03-16
Fix 404 link.

castrojo:
These two URLs broke in the docs, fixing them.


evilnick:
cool. I guess there may be some more like this, I will do a search

PR #307, opened 2015-03-23
Force branch creation.

kadams54:


kadams54:
## QA

1. Run `tools/version.sh 1.20`.
3. Run `tools/version.sh 1.20` a second time; there should be no errors about an existing branch.

PR #309, opened 2015-03-23
Devel release-1.23-beta1 is the current devel release.

abentley:
This updates the text and URLs to reflect the 1.23-beta1 release.


evilnick:
thanks!

PR #310, opened 2015-03-23
Fix markdown anchor extension.

fabricematrat:


fabricematrat:
QA:
make on master if currently failing.
make with this PR runs.
This checks the presence of the text before adding the id for anchor.

marcoceppi:
Thanks for this fix!

PR #335, opened 2015-04-09
added rough support for local building of jujucharms-like docs

evilnick:
This is just a quick fix so that locally built docs look similar to how they appear on jujucharms.com
It uses the css and code from jujucharms, but various parts of the displayed page will not work or display as they would as it obviously lacks all the JS jiggery-pokery.
However, the main body of the docs appear as they will on jujucharms.com, which is useful for checking changes and additions.


evilnick:
addendum - this only works with the 'make build' target at present. the multi-doc builder doesn't copy css stuff.

PR #369, opened 2015-04-15
getting-started: I think this helps make config file path clearer

mattyw:
When I was trying to find the environments.yaml on a windows box I felt that having the file path exploded out like this would have made things easier for me. Just my opinion - feedback appreciated.


evilnick:
I agree - thanks!

PR #391, opened 2015-04-27
Changes to language used in the explanation of what Juju is

evilnick:


src/en/about-juju.md
```diff
@@ -1,15 +1,15 @@
 ##What is Juju?
-Juju is a state-of--the--art, open source, service orchestration tool. Juju allows you to configure, manage, maintain, deploy, and scale cloud services quickly and efficiently on public clouds, as well as on physical servers, OpenStack, and containers. You can use Juju from the command line or through its beautiful GUI.
+Juju is a state-of-the-art, open source, universal model for service oriented architecture and service oriented
+deployments. Juju allows you to configure, manage, maintain, deploy, and scale cloud services quickly and efficiently on public clouds, as well as on physical servers, OpenStack, and containers. You can use Juju from the command line or through its beautiful GUI.
```
marcoceppi: Seems weird that deploy isn't the first word in the sentence. Logically it seems "deploy, configure, manage, and scale" would be a better progression here (even though it's unchanged from the original docs)

marcoceppi:
LGTM, just had a small suggestion

evilnick:
fair point, i'll change that then roll it out across the versions

PR #398, opened 2015-04-29
Storage wip

evilnick:
adds a new section in nav for work in progress docs
adds the storage wip.

### DIFF
--- a/src/en/wip-storage.md
+++ b/src/en/wip-storage.md
@@ -51,22 +51,30 @@ Juju will use a default size of 1G, unless the charm itself has specified a mini
 
 When deploying on a provider which supplies storage, the supported storage pool types may be used in addition to ‘loop’ and ‘rootfs’. For example, on using Amazon’s EC2 provider, we can make use of the default ‘ebs’ storage pool
 
-    juju deploy cs:~axwalk/postgresql --storage data=ebs,10G
+```
+juju deploy cs:~axwalk/postgresql --storage data=ebs,10G
+```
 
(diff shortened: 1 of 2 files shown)


src/en/wip-storage.md
```diff
@@ -0,0 +1,212 @@
...
+
+We can also merely specify the size, in which case Juju will use the default pool for the selected environment.  E.g. 
+
+```
+juju deploy cs:~axwalk/postgresql --storage data=10G
+```
+
+Which, on the EC2 provider, will create a 10 gigabyte volume in the ‘ebs’ pool.
+
+Charms may specify as many different types of storage as they require, in which case they may all be specified using the constraint, or some or all can be omitted to accept the default values:
```
bz2: Instead "Charms may declare multiple types* of storage..."

Don't really like types, but okay for now.

src/en/wip-storage.md
```diff
@@ -0,0 +1,212 @@
...
+juju deploy cs:~axwalk/postgresql --storage data=ebs,10G cache=ebs-ssd
+```
+
+### Persistence (provisional)
+
+Some providers have the option to detach storage from the lifespan of the instance(s) which created/used it. This means that even after services have been removed, the storage and its contents still exist in your cloud, which may be useful for backup, recovery or transport purposes. Juju will not allow you to completely destroy an environment which contains such storage (except by using the --force option). If this feature is enabled for storage pools, you must remember to decommission the storage manually.
+
+### Provider-specific options
+
+AWS/EC2
```
bz2: Also openstack at some point, and maas.

bz2:
Looks like a good start, thanks!

evilnick:
:monkey_face:

PR #404, opened 2015-05-01
Add auth-file to the GCE config notes.

ericsnowcurrently:
"auth-file" was added in 1.23-beta2 (9d28cdfd0230ae9ef7c120a95adf8d53d2c4bfd8).

This patch also cleans up the GCE config notes a little.


ericsnowcurrently:
Note that this change applies to juju 1.23.

PR #411, opened 2015-05-06
Various updates to storage docs

axw:


src/en/wip-storage.md
```diff
@@ -31,19 +31,21 @@ juju storage volume list
 
 As previously mentioned, a charm which requires storage will automatically allocate the default storage (unit filesystem) by default. It is possible to instead specify the storage to be used when the service is deployed, using constraints.
 
-The constraints can specify the type/pool and size of the storage required.
+The constraints can specify the type/pool and size, and optionally a count, of the storage required.
```
anastasiamac: From what I have seen, type/pool and size have reasonable defaults as well. So, they could be "optionally" specified as well :)

src/en/wip-storage.md
```diff
@@ -31,19 +31,21 @@ juju storage volume list
...
 
 ```
-juju deploy <charm> --storage <label>=<pool>,<size>
+juju deploy <charm> --storage <label>=<pool>,<size>[,count]
 ```
 
-For example, to deploy the postgresql service and have it use the unit’s local filesystem for 10 Gigabytes of storage for its ‘data’ storage requirement:
+Unless specified, the count is 1; this means that only one instance of the specified store will be created.
```
anastasiamac: Well, if store in charm specifies that minimum count is 0, then, when not specified, the count would be 0 not 1.

src/en/wip-storage.md
```diff
@@ -84,16 +86,27 @@ Charms may declare multiple types of storage, in which case they may all be spec
 juju deploy cs:~axwalk/postgresql --storage data=ebs,10G cache=ebs-ssd
 ```
 
-### Persistence (not working!)
+### Persistence (incomplete!)
+
+Some providers have the option to detach storage from the lifespan of the instance(s) which created/used it. This means that even after services have been removed, the storage and its contents still exist in your cloud, which may be useful for backup, recovery or transport purposes. Juju will not allow you to completely destroy an environment which contains such storage (except by using the --force option). To create persistent volumes, create a storage pool with the "persistent" attribute set to "true". Persistent volumes are currently supported by the EBS and Cinder storage providers. Currently, all Cinder volumes are considered persistent, whether the pool is configured as such.
```
wallyworld: s/whether the pool/regardless of whether the pool

src/en/wip-storage.md
```diff
@@ -84,16 +86,27 @@ Charms may declare multiple types of storage, in which case they may all be spec
...
+ - tmpfs: filesystem-type, creates a tmpfs
 
-### Provider-specific options
+Additionally, native storage providers exist for the EC2 (ebs) and OpenStack (cinder).
 
-AWS/EC2
-OpenStack
-MAAS
+The EC2/EBS provider currently supports the following pool configuration attributes:
+ - volume-type: specifies the EBS volume type to create (gp2 (ssd), io1 (provisioned-iops), standard (magnetic)). By default, magnetic/standard volumes will be created. An 'ebs-ssd' pool is created in all EC2 environments, which defaults the volume-type to ssd/gp2.
```
wallyworld: it's not 100% clear to a new user what the EBS option is vs the Juju one ie what does a user type into the juju create pool command gp2 or ssd

src/en/wip-storage.md
axw: will remove "optionally"

src/en/wip-storage.md
```diff
@@ -84,16 +86,27 @@ Charms may declare multiple types of storage, in which case they may all be spec
...
+ - iops: the number of IOPS for provisioned-iops volume types.
```
wallyworld: i think we could/should add a note that iops restrictions exist, see http://blah for more details

src/en/wip-storage.md
axw: It's definitely 1. If you don't specify any storage constraints for a store with min=0, then it won't create anything. But if you do specify constraints without count, it'll default to just creating 1.

src/en/wip-storage.md
```diff
@@ -107,29 +120,45 @@ Storage requirements _may_ be added to the 'metadata.yaml' of the charm as follo
 storage:
   data:
     type: filesystem
+    description: junk storage
+    shared: false
```
wallyworld: if we add shared and read-only can we add a note here that they're not supported yet

src/en/wip-storage.md
axw: fixed, thanks

src/en/wip-storage.md
```diff
@@ -107,29 +120,45 @@ Storage requirements _may_ be added to the 'metadata.yaml' of the charm as follo
...
+    read-only: false
+    minimum-size: 100M
     location: /srv/data
 ```
 
-Here the charm is asking for storage it is calling 'data', and it further defines a type and location. It is possible to specify as many entries as desired for storage, and all of the related keys are optional. 
+Here the charm is asking for storage it is calling 'data', and it further defines a type and location. It is possible to specify as many entries as desired for storage, and all but the 'type' key are optional. The 'type' attribute specifies the type of the storage: filesystem or block (i.e. block device/disk). The 'minimum-size' attribute specifies the minimum size of the store, overriding the default of 1G if the user does not specify a size. The location specifies the path at which to mount filesystem-type storage. The 'read-only' and 'shared' attributes are currently not handled. Support will be added in a future version of Juju.
```
wallyworld: instead of 1G, can we say either 1GiB or 1GB (I can't recall which OTTOMH) just to be 100% explicit. or does 1G mean the same as 1GB

src/en/wip-storage.md
```diff
@@ -107,29 +120,45 @@ Storage requirements _may_ be added to the 'metadata.yaml' of the charm as follo
...
-@hooks.hook('data-storage-attached')
-def data_storage_attached():
-    hookenv.log("Storage ready and mounted", hookenv.DEBUG)
-    config_changed(mount_point=external_volume_mount)
+mountpoint=$(storage-get location)
+sed -i /etc/myservice.conf "s,MOUNTPOINT,$mountpoint"
 ```
-(this example assumes you are writing hooks in python using the charm helpers)
 
+The storage-attached hooks will be run before the install and upgrade-harm hooks, so that installation and upgrade routines may use the storage. The storage-detaching hooks will be run before storage is detached, and always before the stop hook is run, to allow the charm to gracefully release resources before they are removed and before the unit terminates.
```
wallyworld: harm->charm

src/en/wip-storage.md
```diff
@@ -185,33 +209,16 @@ IMPLEMENTED FEATURES
...
 
-- Storage destruction. Unit and machine destruction are prevented if they have attached storage, so if you're playing with storage expect to have to "destroy-environment --force".
-- Unit agent will not yet wait for required storage before installing.
+- Persistent storage destruction. If you use persistent storage, you must use "destroy-environment --force" and manually destroy the storage through your cloud's management UI.
 - Unit/machine placement is currently disabled if storage is specified.
-- Doesn't matter at the moment due to above point, but charm deployment currently does not check for mount-point conflicts.
+- Charm deployment currently does not check for mount-point conflicts.
 - Charm upgrade does not currently check for incompatible changes to storage requirements in deployed charms.
-- No X-storage-detach(ing|ed) hook yet; it hopefully goes without saying, but anyway: since storage destruction isn't yet done, you won't be notified when storage is destroyed.
 - storage-add command: this is being worked on now, and will hopefully be ready soon.
```
wallyworld: will be ready for juju 1.25

src/en/wip-storage.md
```diff
@@ -185,33 +209,16 @@ IMPLEMENTED FEATURES
...
- - rootfs: filesystem-kind, creates a sub-directory in the agent's data-dir for the unit/charm to use
-
-We have implemented support for creating volumes in the ec2 provider, via the "ebs" storage provider. By default, the ebs provider will create cheap and nasty magnetic volumes. There is also an "ebs-ssd" storage pool provided OOTB that will create SSD (gp2) volumes. Finally, you can create your own pools if you like; the parameters for ebs are:
-  - volume-type: may be "magnetic", "ssd", or "provisioned-iops"
-  - iops: number of provisioned IOPS (requires volume-type=provisioned-iops)
-
-Some storage providers also support a "persistent=<bool>" pool attribute. By using this, Juju will not tie the lifetime of storage entities (volumes, filesystems) to the lifetime of the machines that they are attached to. In EC2/EBS terms, a persistent volume is one without any attachments having the DeleteOnTermination flag set. Juju will not allow you to cleanly destroy an environment with persistent volumes; you may use "--force" to subvert this as usual, but please be aware that this will leak the resource.
 
 UNIMPLEMENTED/CAVEATS
```
wallyworld: also mention charm upgrades

src/en/wip-storage.md
axw: We define 1G to be 1GiB. I've changed instances of 1G in descriptions to 1GiB, but left as 1G in the example CLI invocations.

src/en/wip-storage.md
axw: added #comments

src/en/wip-storage.md
axw: done

src/en/wip-storage.md
axw: reworded

src/en/wip-storage.md
axw: fixed

src/en/wip-storage.md
axw: done

src/en/wip-storage.md
axw: already did:
"Charm upgrade does not currently check for incompatible changes to storage requirements in deployed charms."

wallyworld: and maybe explicitly shared and read only again

evilnick:
Excellent. Thanks for the updates on this!

PR #426, opened 2015-05-26
adding title to maas page as example

evilnick:


evilnick:
@petermatulis You can merge this one first!

PR #431, opened 2015-05-28
vagrant howto code snippets updated for 14.04

andrewdmcleod:
install hook example snippet changed to use rubygems-integration in 14.04, and not to set alternatives (this errors when using the current jujubox as per the howto, so removed completely)

port set to 80 in start hook example snippet - perhaps should be rewritten to be defaulted in the config

### DIFF
--- a/src/en/howto-vagrant-workflow.md
+++ b/src/en/howto-vagrant-workflow.md
@@ -89,9 +89,8 @@ hooks.
 
     #!/bin/bash
     set -ex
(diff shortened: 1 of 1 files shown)


marcoceppi:
This is a candidate for backporting

evilnick:
@andrewdmcleod Thanks for the contribution!

evilnick:
I am going to merge this and add the config.yaml change at a later date. Thanks

Issue #378, opened 2015-04-20
Documentation is confusing for Juju Charm Actions


mitechie:
ported from https://github.com/CanonicalLtd/jujucharms.com/issues/88

https://jujucharms.com/docs/1.20/authors-charm-actions

About 1/3 of the way down the page.

The example schema for a Juju Action specifies that the actions directory should contain the 2 executables (report and run). 

Yet the schema specified below specifies that the name of the script should be "snapshot." 

As well as the example command provided:

juju action do mysql/0 snapshot filename=out.tar.gz compression.type=gzip 

Example Schema

Charm dir
├── actions
│   ├── report  <--------------------------------- Should this be snapshot ????
│   └── run
├── actions.yaml
...

actions.yaml example
# actions.yaml

report:
snapshot: 
  description: Take a snapshot of the database.
  params:
    filename: 
      type: string
      description: The name of the snapshot file.
    compression:
      type: object
      description: The type of compression to use.
      properties:
        kind:
          type: string
          enum: [gzip, bzip2, xz]
        quality:
          description: Compression quality
          type: integer
          minimum: 0
          maximum: 9
  required: [filename]
  additionalProperties: false

copying @puravidahope

evilnick:
@binary132 is there any clarification on this?

binary132:
Yes.  Good catch.  Should be `report` and `snapshot`.  :grimacing:

PR #437, opened 2015-05-31
Clarify discrepancy between Actions schema and example

binary132:
cf. https://github.com/juju/docs/issues/378


evilnick:
Ace. Thanks for the fix

PR #438, opened 2015-05-31
added step to enable google compute api - 4a

andrewdmcleod:
as per subject, step to enable google compute api missing from instructions

### DIFF
--- a/src/en/config-gce.md
+++ b/src/en/config-gce.md
@@ -126,6 +126,11 @@ private key. You can however generate a new key by returning to the credentials
 page and clicking on the `Generate new JSON key` button, after which you may
 want to delete the original fingerprint entry from the list on the same page.
 
+## 4a. Enable Google Compute Engine API
+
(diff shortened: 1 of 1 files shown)


evilnick:
Hi, thanks for this.
The gce API was enabled by default on the two projects I created whilst testing this, but I take it this is not always the case?
I will rework step 4 to include a check for this.

andrewdmcleod:
I spun up a stack yesterday and it wasn't enabled by default - maybe its a 'might be needed sometimes' step?

evilnick:
Mine was brand new too, I just opened an account to do the instructions, so it seems there is a lack of consistency. I ran through it twice actually, to check the different with methods, and both projects enabled the API.

I think you are right, a might-be-needed step it is. I will take care of it. And thanks again for pointing this out, I'm sure others would have been tripped up by it.

evilnick:
this was fixed with #445 .
Thanks!

PR #440, opened 2015-06-01
Subordinate services: Broken links

schkovich:
Fixed broken links to the pages having detailed information on scoped and implicit relations. While there is a page titled [Implicit relations](https://jujucharms.com/docs/devel/authors-implicit-relations) the _**Scoped relations**_ page does not exist. However, scope of relations (global, container) is discussed in details on page [Relations in depth](https://jujucharms.com/docs/devel/authors-relations-in-depth). I considered that linking _**Relations in depth**_ page is better than 404.


evilnick:
Hi, many thanks for taking the time to do this. the broken links were fixed by another branch that was in the queue for review already (and also fixed some problems with page names and out of date pages being linked.
I/we do appreciate the effort though!

PR #459, opened 2015-06-05
code fencing src/en/tools-foo.md

petermatulis:
* I didn't fence command output.
* I added the 'python' language fence.

Will wait for verification of this commit before continuing with the other files.

### DIFF
--- a/src/en/tools-amulet.md
+++ b/src/en/tools-amulet.md
@@ -388,8 +388,10 @@ Return files and directories of directory
 
 Example of output
 
-    {'files': []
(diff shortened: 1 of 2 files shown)


evilnick:
Yes to python,
Do fence output though, you can use the "no highlight" type. Apparently the auto syntax highlighting isn't very good at auto detection, so it is best to be explicit!

evilnick:
You can modify this same branch, so you don't need to resubmit the PR.

evilnick:
:+1:

PR #462, opened 2015-06-09
add Vmware config

evilnick:
basic instructions for configuring vSphere

fixes #462 #457

### DIFF
--- a/src/en/config-vmware.md
+++ b/src/en/config-vmware.md
@@ -1,0 +1,51 @@
+
+Title: Configuring Juju for use with vSphere
+
+# Configuring the VMWare (vSphere) Provider
+
+Juju supports VMWare's vSphere ("Software-Defined Data Center") installations
+as a targetable cloud. In order to use the vSphere provider, you will need to
+have an existing vSphere installation, which supports VMWare's Hardware Version
+8 or better.
+
+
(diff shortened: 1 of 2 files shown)


src/en/config-vmware.md
```diff
@@ -0,0 +1,51 @@
...
+ vsphere:
+    type: vsphere
+    host: <api-endpoint>
+    user: <some-user>
+    password: <some-password>
+    datacenter: <datacenter-name>
+    external-network: <external-network-name>
+```
+
+The values contained within angle brackets (`<` and `>`) need to be replaced
```
petermatulis: Sentence strikes as odd to me. Also seen, albeit a different variation, on the Helion instructions. Newbies may also  just replace the values within the brackets (and keep the brackets).

Maybe standardize and clarify the wording. Suggestion:

"Enter your local settings above (ex: "password: secret")."

It's not completely generic as each provider has its own parameters so some editing is required.

petermatulis:
Otherwise good to go.

src/en/config-vmware.md
evilnick: yes. part of the problem is that this isn't standardised in the boilerplate files themselves, which is something I guess we will have to fix

evilnick:
ok

PR #474, opened 2015-06-11
Many fixes to the Glossary:

mbruzek:
- Corrected "boostrap"
- Alphabetized the terms
- Truncated the strings to < 80 characters
- Added a definition for "Bundle"
- Made the Charm URL parts an indented list
- Reworded the "Relation" definition


mbruzek:
This change started with the intent to remove "boostrap" and I could not leave well enough alone!

evilnick:
Good! The glossary was crap

PR #475, opened 2015-06-12
change outfile.name to outfile for consistency

jrwren:
Using the .name suffix in one place but not throughout the entire document was confusing. I have removed that. I hope this is more clear.


petermatulis:
:+1:

Issue #444, opened 2015-06-02
Code examples need specified type for highlighting to work


evilnick:
- [x] about-juju.md	
- [x] actions.md	
- [x] authors-charm-actions.md	
- [x] authors-charm-best-practice.md	
- [x] authors-charm-components.md	
- [x] authors-charm-config.md	
- [x] authors-charm-hooks.md	
- [x] authors-charm-icon.md	
- [x] authors-charm-metadata.md	
- [x] authors-charm-policy.md	
- [x] authors-charm-store.md	
- [x] authors-charm-upgrades.md	
- [x] authors-charm-writing.md	
- [x] authors-hook-debug-dhx.md
- [x] authors-hook-debug.md	
- [x] authors-hook-environment.md	
- [x] authors-hook-errors.md	
- [x] authors-implicit-relations.md	
- [x] authors-intro.md	
- [x] authors-relations-in-depth.md	
- [x] authors-relations.md	
- [x] authors-service-config.md	
- [x] authors-subordinate-services.md	
- [x] authors-testing.md	
- [x] charm-review-process.md	
- [x] charm-unmaintained-process.md	
- [x] charms-bundles.md	
- [x] charms-config.md	
- [x] charms-constraints.md	
- [x] charms-deploying.md	
- [x] charms-destroy.md	
- [x] charms-environments.md	
- [x] charms-exposing.md	
- [x] charms-ha.md	
- [x] charms-relations.md	
- [x] charms-relationships.md	
- [x] charms-scaling.md	
- [x] charms-service-groups.md	
- [x] charms-upgrading.md	
- [x] charms-working-with-units.md	
- [x] charms.md	
- [x] commands.md	
- [x] config-LXC.md	
- [x] config-aws.md	
- [x] config-azure.md	
- [x] config-digitalocean.md	
- [x] config-gce.md	
- [x] config-general.md	
- [x] config-hpcloud.md	
- [x] config-joyent.md	
- [x] config-local.md	
- [x] config-maas.md	
- [x] config-manual.md	
- [x] config-openstack.md	
- [x] config-scaleway.md	
- [x] config-vagrant.md
- [x] config-vmware.md	
- [x] contributing.md	
- [x] getting-started-keygen-win.md	
- [x] getting-started.md	
- [x] glossary.md	
- [x] howto-authorised-keys.md	
- [x] howto-backup-restore.md	
- [x] howto-gui-management.md	
- [x] howto-harvesting.md	
- [x] howto-highavailability.md	
[...]

PR #476, opened 2015-06-12
code fencing + general review: src/en/g_files.md

petermatulis:
More code/output fencing and general review of

src/en/g*.md

#444 updated

### DIFF
--- a/src/en/getting-started.md
+++ b/src/en/getting-started.md
@@ -35,8 +35,8 @@ brew install juju
 
 We also recommend trying Juju in [our Vagrant box](config-vagrant.html).
 
-For more installation information and what versions are available, see [the
-releases page](reference-releases.html).
+For more installation information and what versions are available, see
+[the releases page](reference-releases.html).
 
 ## Windows
 
@@ -55,9 +55,12 @@ following file:
     ~/.juju/environments.yaml
 
 ## Windows
-    %LOCALAPPDATA%/Juju
+
+    %APPDATA%\Juju\environments.yaml
     
-Juju can automatically generate this file in this way:
+    %APPDATA% typically == C:\Users\<user>\AppData\Roaming
+
+Juju can automatically generate the file in this way:
 
 ```bash
 juju generate-config
@@ -69,17 +72,18 @@ services.  Edit it to provide specific information for your chosen cloud
 provider. For more specifics on what needs to be changed, see the relevant
 sections in the left pane (under *Install & Configure*).
 
-**Note:** Juju's command line interface includes documentation, doing `juju
-help` will show you the topics. You can also look at the [Juju command
-cheatsheet](https://github.com/juju/cheatsheet) if you are looking for a handy
-command guide.
+**Note:** Juju's command line interface includes documentation, running `juju
+help` will show you the topics. You can also look at the
+[Juju command cheatsheet](https://github.com/juju/cheatsheet) if you are
+looking for a convenient command guide.
 
 # Testing your setup
 
 The first step is to create a bootstrap environment. This is a cloud instance
 that Juju will use to deploy and manage services. It will be created according
-to the configuration you have provided, and your SSH key will automatically be
-injected so that Juju can communicate securely with the bootstrap instance.
+to the configuration you have provided, and your public SSH key will be
+uploaded automatically so that Juju can communicate securely with the
+bootstrap instance.
 
 <iframe style="margin-left: 20%;" class="youtube-player" type="text/html"
 width="420" height="350" src="//www.youtube.com/embed/0AT6qKyam9I"></iframe>
--- a/src/en/getting-started.md
+++ b/src/en/getting-started.md
@@ -1,12 +1,14 @@
 # Introduction
 
-This tutorial will show you how to get started with Juju, including installing,
-configuring and bootstrapping a new Juju environment. Before you start you will
-need:
+This tutorial will explain how to get started with Juju, including installing,
+configuring, and bootstrapping a new Juju environment. Prerequisites include:
 
-  - An Ubuntu, OSX or Windows machine to install the client on.
-  - An environment which can provide a new server with an Ubuntu cloud operating system image on-demand. This includes services such as Amazon EC2, HP Cloud, an OpenStack installation, or your local machine
-  - An SSH key-pair. If you have not already set one up, you need to generate an appropriate keypair. On Linux and Mac OSX: `ssh-keygen -t rsa -b 2048` On Windows: See the [Windows instructions for SSH and PuTTY](getting-started-keygen-win.html)
+  - An Ubuntu, OSX, or Windows machine to install the client upon.
+  - An environment which can provide a new server with an Ubuntu cloud operating
+  system image on-demand. This includes services such as Amazon EC2, HP Cloud,
+  an OpenStack installation, or your local machine
+  - An SSH key-pair. On Linux and Mac OSX: `ssh-keygen -t rsa -b 2048` On Windows:
+  See the [Windows instructions for SSH and PuTTY](getting-started-keygen-win.html)
 
 # Installation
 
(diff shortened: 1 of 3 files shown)


src/en/getting-started.md
```diff
@@ -49,52 +57,40 @@ particular cloud provider. This is done by generating and editing the following
 ## Windows
     %LOCALAPPDATA%/Juju
```
petermatulis: June 12 @ EDT on #juju
14:33 <    hazmat> pmatulis: on windows it goes to the user's HOMEDRIVE + HOMEPATH env var then .juju
14:42 <  pmatulis> hazmat: so the online docs are wrong then. they currently show
14:42 <  pmatulis> %LOCALAPPDATA%/Juju
*silence*

src/en/getting-started.md
```diff
@@ -15,20 +17,26 @@ need:
...
 
-    brew install juju
+```bash
+brew install juju
+```
 
 We also recommend trying Juju in [our Vagrant box](config-vagrant.html).
 
-For more information on installing and the current versions available, see [the releases page](reference-releases.html).
+For more installation information and what versions are available, see [the
```
evilnick: I tend to avoid at pretty much all costs, breaking links across lines - all markdown parsers seem to be particularly bad at  catching these, plus it makes them far harder to spot in the raw text

src/en/getting-started.md
evilnick: cool. Well, I don't have a Windows setup to test this, so...

src/en/getting-started.md
```diff
@@ -49,52 +57,40 @@ particular cloud provider. This is done by generating and editing the following
...
-  - [Configuring for LXC local provider (Linux)](config-local.html)
-  - [Configuring for Vagrant local provider (OSX, Linux, and Windows)](config-vagrant.html)
+This action will not overwrite an existing file but merely dump the information
+onscreen (STDOUT). It will contain sample profiles for different types of cloud
+services.  Edit it to provide specific information for your chosen cloud
+provider. For more specifics on what needs to be changed, see the relevant
+sections in the left pane (under *Install & Configure*).
 
-**Note:** Juju's command line interface includes documentation, doing `juju help` will show you the topics. You can also check out the [Juju command cheatsheet](https://github.com/juju/cheatsheet) if you are looking for a handy command guide.
+**Note:** Juju's command line interface includes documentation, doing `juju
```
evilnick: running juju help would be better

src/en/getting-started.md
```diff
@@ -49,52 +57,40 @@ particular cloud provider. This is done by generating and editing the following
...
+help` will show you the topics. You can also look at the [Juju command
+cheatsheet](https://github.com/juju/cheatsheet) if you are looking for a handy
```
evilnick: see previous note about breaking links

src/en/getting-started.md
```diff
@@ -49,52 +57,40 @@ particular cloud provider. This is done by generating and editing the following
...
-
-The first thing to do is set up a bootstrap environment. This is an instance in
-the cloud that Juju will use to deploy and control other services with. It will
-be created according to the configuration you have provided, and your SSH key
-will automatically be uploaded so that Juju can communicate securely with the
-bootstrap instance.
+The first step is to create a bootstrap environment. This is a cloud instance
+that Juju will use to deploy and manage services. It will be created according
+to the configuration you have provided, and your SSH key will automatically be
+injected so that Juju can communicate securely with the bootstrap instance.
```
evilnick: I don't like some of this rewording. Why have we changed 'uploaded' to 'injected' for example?

src/en/glossary.md
```diff
@@ -39,26 +36,24 @@ following format and restrictions:
 
 **Endpoint** - The combination of a service name and a relation name.
 
-**Environment** - An Environment is a configured location where Services can
-be deployed onto. An Environment typically has a name, which can usually be
-omitted when there's a single Environment configured, or when a default is
-explicitly defined. Depending on the type of Environment, it may have to be
-bootstrapped before interactions with it may take place (e.g. EC2). The local
-environment configuration is defined in the ~/.juju/environments.yaml file.
+**Environment** - A configured location where services can be deployed onto. An
```
evilnick: I would lose the 'onto'

src/en/glossary.md
```diff
@@ -39,26 +36,24 @@ following format and restrictions:
...
+Environment typically has a name, which can usually be omitted when there's a
+single Environment configured, or when a default is explicitly defined.
+Depending on the type of Environment, it may have to be bootstrapped before
+interactions with it may take place (e.g. EC2). The local environment
+configuration is defined in the ~/.juju/environments.yaml file.
```
evilnick: not sure this holds true for all platforms.

src/en/glossary.md
```diff
@@ -1,26 +1,23 @@
 # Glossary of terms
 
-**Bootstrap** - Initialize a Juju Environment so that Services may be deployed.
+**Bootstrap** - Initialize a Juju environment so that services may be deployed.
```
evilnick: you changed case on some words here, but not in the doc as a whole. I think there should be more consistency, but possibly capitalising them is not the answer. Maybe backticking them so they show up as highlighted in a monospaced font. My guess is they are to call out other terms which are defined on this page.

src/en/glossary.md
```diff
@@ -39,26 +36,24 @@ following format and restrictions:
...
-terminating machines in an Environment, as necessary for the requested
-configuration.
+terminating machines in an Environment, as necessary for the requested configuration.
 
-**Relation** - Relations are the way in which juju enables Services to
-communicate to each other, and the way in which the topology of Services is
-assembled. The Charm defines which Relations it offers to other Services
-(provides), and what kind of Relations it can make with other Services
-(requires).
+**Relation** - The way in which juju enables Services to communicate with each
```
evilnick: Juju

src/en/getting-started.md
petermatulis: I'm trying to be more concise (4 lines instead of 10).

The term ["injected"](https://goo.gl/4LOYRC) is used very often when authentication material is automatically introduced during a setup process.

src/en/glossary.md
petermatulis: I was debating this too. The terms we're using are tricky. In some contexts a Juju environment is not a unique/proper thingy but in others it is (the Juju Environment). Ditto for "Juju service".

src/en/glossary.md
petermatulis: I didn't change that but ok.

src/en/getting-started.md
evilnick: I understand. I think we want to worry less about being concise and more about using natural language which expresses things clearly, especially in the pages aimed at less experienced users. It is certainly okay to cut out unnecessarily verbose passages, but I think you might want to rein that in a little bit when it reduces the conversational nature of introductory text. Indeed, 'injected' is a word that is commonly used in tech docs around this subject, but "uploaded" is actually what happens, is arguably easier to understand, and certainly easier to translate.

src/en/glossary.md
evilnick: Sure, I think it was in the additions matt made a few days back, but as we are checking it over... I think it would be okay to simply say the environments.yaml file and maybe link to whereits location is explained

src/en/glossary.md
evilnick: I would also be fine using lower case for everything and not treating them as special unless they are being used in a particularly special way, which I haven't noticed so far

src/en/getting-started.md
petermatulis: Understood.

src/en/getting-started.md
petermatulis: Note: this file (getting-started) is bundled into the Windows installer (as a README) and pops up upon installation.

petermatulis: yaml file ends up here:

C:\Users\{user}\AppData\Roaming\Juju\environments.yaml

where

%APPDATA% = C:\Users\{username}\AppData\Roaming

Not LOCALAPPDATA and not .juju

Also assuming Windows is installed on the C: drive.

evilnick: good to know!

src/en/getting-started.md
petermatulis: What about changing the name of this file from "getting-started.md" to "install-configure.md"? The title in the TOC (left pane) shows "Install & Configure".

There is an entry in the TOC that is called "Getting started" whose corresponding filename is "authors-intro.md".

evilnick: Ah, there is a whole history there. Short answer is, it would break too many things which link here, without setting up a redirect which I can no longer actually do

src/en/glossary.md
petermatulis: Alright, "lowercasing" the terms.

evilnick:
:+1:

PR #478, opened 2015-06-16
Code fencing+general review actions+contributing

petermatulis:
More code/output fencing and general review of

src/en/actions.md
src/en/contributing.md

#444 updated


petermatulis:
Somehow my previous work on src/en/g*_files.md got put in here.

evilnick:
that's what happens if you don't keep your branches separate. when you have finished one thing, you need to checkout master again before you checkout and create a new branch.

petermatulis:
Strange, I have yet to checkout master and this never happened before. I think what might have occurred here is that I created a new branch before committing the changes from an existing branch. Can we proceed here or do I need to do something special?

evilnick:
i think since the changes in the previous branch were merged, it should resolve itself

evilnick:
Hmm, except, it seems to have the changes you made to the other files  before they were corrected. I think the easiest way to proceed is for you to make a new local clone, create a new branch on it and copy  the files from this branch into it, then commit  that and do a PR.

PR #487, opened 2015-06-18
action feature flag should be 'action'

natefinch:
See relevant code here: https://github.com/juju/juju/blob/1.22/cmd/juju/action/action.go#L20


marcoceppi:
Thanks for this fix

PR #490, opened 2015-06-18
Review  config osv files.md

petermatulis:


petermatulis:
Still something wrong with my workflow. Commits from PR 489 are in here. I closed the latter.

PR #491, opened 2015-06-19
 status command usage example

tyloch:
status command usage examples: simple and very useful  option with watch command


evilnick:
:+1:  thanks

Issue #300, opened 2015-03-19
juju block needs docs


evilnick:
### Blocking changes to the environment

The 'block' and 'unblock' commands control which changes can be made to
a running environment. Use the 'block' command to lockdown the
environment to prevent accidental changes. Three levels of restrictions
can be applied, 'destroy-environment', 'remove-object', and
'all-changes'.

The 'destroy-environment' level blocks just the 'destroy-environment'
command. You can prevent an environment from being destroyed like this:

    juju block destroy-environment

The 'remove-object' level blocks the 'destroy-environment',
'remove-machine', 'remove-relation', 'remove-service', and 'remove-unit'
commands from being run. You can prevent the machines, services, units
and relations from being removed thusly:

    juju block remove-object

The 'all-changes' level blocks commands that add, alter, or remove
things in the running environment. These commands are blocked:
'add-machine', 'add-relation', 'add-unit', 'authorised-keys add',
'authorised-keys delete', 'authorised-keys import', 'deploy',
'destroy-environment', 'ensure-availability', 'expose', 'resolved',
'remove-machine', 'remove-relation', 'remove-service', 'remove-unit',
'retry-provisioning', 'run', 'set', 'set-constraints', 'set-env',
'unexpose', 'unset', 'unset-env', 'upgrade-charm', 'upgrade-juju',
'user add', 'user change-password', 'user disable', and 'user enable'
You can lockdown the environment to prevent all changes like this:

    juju block all-changes

When a command is blocked, and you are certain you need to alter the
environment, you can remove the block using the 'unblock' command. For
example, to permit the remove-relation command blocked by the
'remove-object' level, run:

    juju unblock remove-object

You can restore the block after you complete your changes.

Note that some commands have a '--force' option bypasses the
restrictions set by 'block'. Only use the '--force' option *after* the
command was run without it to ensure you are aware of the restrictions
set on the environment by the 'block' command.

For more information run 'juju help block' and 'juju help unblock'.

jamesbeedy:
Additional docs would be a huge help!!

evilnick:
@petermatulis If you are bored with code-fencing, perhaps you would like to take a look at this? It requires a new page!

PR #492, opened 2015-06-19
New page  howto juju block.md

petermatulis:
As per #300

### DIFF
--- a/src/en/howto-juju-block.md
+++ b/src/en/howto-juju-block.md
@@ -85,7 +85,8 @@ juju block all-changes
 
 ## Adding restrictions via environments.yaml
 
-You have the option of hardcoding restriction levels via environments.yaml.
+You have the option of setting restriction levels via environments.yaml instead
+of using the 'block' command.
 
 The three (Boolean) parameters corresponding to the three levels are:
 
--- a/src/en/howto-juju-block.md
+++ b/src/en/howto-juju-block.md
@@ -1,3 +1,5 @@
+Title: Using Juju block to prevent accidental changes
+
 # Restricting changes to the running Juju environment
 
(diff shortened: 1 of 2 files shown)


src/en/howto-juju-block.md
```diff
@@ -0,0 +1,105 @@
+# Restricting changes to the running Juju environment
```
evilnick: You should add a title. The metadata gets picked up by the build and is used as the actual page title in the browser. The first line just needs to be something like :

Title: Using Juju block to prevent accidental changes

src/en/howto-juju-block.md
```diff
@@ -0,0 +1,105 @@
...
+
+Typically you would restore the block after having made the change.
+
+!!! Note: The '--force' option recognized by some Juju commands bypasses any
+restriction level that would otherwise apply. If your policy is to use
+restrictions then the immediate use of the '--force' option should not be part
+of your workflow. If you must use it, do so after having first run the Juju
+command without it to ensure you are aware of any possible restrictions.
+
+For more information run ```juju help block``` and ```juju help unblock```.
```
evilnick: would be cool to also include adding these options to environments.yaml. the key pairs are detailed in https://jujucharms.com/docs/devel/config-general#alphabetical-list-of-general-configuration-values

src/en/howto-juju-block.md
petermatulis: Dunno, my tab (Chrome, Firefox) shows this for all pages:

Introducing Charms | Documentation | Juju

petermatulis: Disregard. I only see the above with my local web browser.

PR #497, opened 2015-06-22
fences for units

evilnick:
I have started doing these one at a time. I need to finish up some status stuff now, so this is the last one I was working on today at least. it would be cool to get all the boxes ticked before we make a 1.24 version as it avoids unpleasant backporting, which is going to get messy as some of the files may have to change


petermatulis:
:+1:

PR #498, opened 2015-06-22
review: src/en/config-lm_files.md

petermatulis:
updated #444


evilnick:
Cool. I think at some point we want to go back and rationalise the titles, so they are more consistent.

PR #501, opened 2015-06-22
Review  charms s files.md

petermatulis:
updated #444

### DIFF
--- a/src/en/charms-relations.md
+++ b/src/en/charms-relations.md
@@ -1,3 +1,4 @@
+Title: Managing Relationships
 # Managing Relationships
 
 Few services you might want to run can do so completely independently - most of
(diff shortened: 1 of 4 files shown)


petermatulis:
So files (those from r_files) are included in this commit. I think I know what happened. Other PR closed.

petermatulis:
Comment from other PR still stands. Added below.

> These 2 files refer to

> charms-relations.html
> charms-relationships.html

> But I cannot find the latter on the website. I performed a review therefore at a high level since I'm not sure about these files.

evilnick:
:+1: 

I think the relationships file was superceded, but I was unsure about information of value in there, so I left it at the time. I opened #508 to sort this, but it is more important we get stuff shored up for 1.24 now

PR #513, opened 2015-06-23
review: src/en/charms-bundles.md

petermatulis:
updated #444


evilnick:
cool, thanks for sorting out the lists too

:palm_tree:

PR #514, opened 2015-06-23
Benchmarks documentation

marcoceppi:
We would like to merge this into the docs, for 1.23 and 1.24 and devel. This is the documentation on writing benchmarks for Juju charms. We're not sure the page it should reside, guidance on that would be great.

### DIFF
--- a/src/en/draft-benchmarking.md
+++ b/src/en/draft-benchmarking.md
@@ -107,10 +107,18 @@ benchmark-finish
 # of your benchmark.
 
 # Raw results
-action-set results.raw="`cat /opt/load-gen/cpu/results/$run/load-gen.log`"
+benchmark-raw /opt/load-gen/cpu/results/$run/load-gen.log
 
 # Parse the output from load-gen
 score=`awk '{print $1}' /opt/load-gen/cpu/results/$run/load-gen.log`
+availability=`awk '{print $2}' /opt/load-gen/cpu/results/$run/load-gen.log`
+responsetime=`awk '{print $3}' /opt/load-gen/cpu/results/$run/load-gen.log`
+throughput=`awk '{print $4}' /opt/load-gen/cpu/results/$run/load-gen.log`
+
+benchmark-data 'score' '${score}' 'requests/sec' 'desc'
+benchmark-data 'availability' '${availability}'
+benchmark-data 'throughput' '${throughput}' 'bytes/sec' 'desc'
+benchmark-data 'response-time' '${responsetime}' 'secs' 'asc'
 
 # Set the composite score
 benchmark-composite ${score} "requests/sec" "desc"
--- a/src/en/draft-benchmarking.md
+++ b/src/en/draft-benchmarking.md
@@ -20,6 +20,21 @@ If you bundle a benchmarking service as part of your service, or if the benchmar
 suite for your application runs on the same machine, this will likely be the path
 for you.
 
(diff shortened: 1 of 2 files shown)


evilnick:
Cool, this will be a nice addition. I will think about placement and do a proper review in the morning. In the meantime, would you mind wrapping the text to 80 columns?  it makes it a lot easier to comment in the right places!

evilnick:
As for *where* to put it, it depends on whether this is something we want to push on, in which case I would suggest under the 'actions and actions yaml' link, or whether it is something we merely want people to know about if they care to find it, in which case putting it above charm testing would suffice.

Could you also rename it 'authors-charm-benchmarks`

src/en/draft-benchmarking.md
```diff
@@ -0,0 +1,147 @@
+# Why benchmarking
```
evilnick: Could you put in a metadata title as the first line please:

Title: Adding benchmarks to your charm

src/en/draft-benchmarking.md
```diff
@@ -0,0 +1,147 @@
...
+
+```yaml
+load-gen:
+  description: Generate load against the service
+  params:
+    size:
+      type: number
+      description: size of load to generate
+      default: 1
+```
```
evilnick: I get that this is just like actions, but a brief description of what this example does would probably be useful

src/en/draft-benchmarking.md
```diff
@@ -0,0 +1,147 @@
+# Why benchmarking
+
+Before proceeding make sure to read and understand the [Juju Actions](./authors-charm-actions)
+section of the docs as it will be referenced heavily in the upcoming sections
+
+# Types of benchmarks in charms
+
+Two types of benchmark: specific and generic.
```
evilnick: this isn't really mentioned elsewhere - is it needed/relevant?

src/en/draft-benchmarking.md
```diff
@@ -0,0 +1,147 @@
+# Why benchmarking
+
```
evilnick: It could probably do with a bit more explanation here - what are the benefits of adding a benchmark to your charm (testing against different architectures/ cloud providers, identifying bottlenecks and optimum config values  or whatever). In fairness, it should probably make it obvious that benchmarks are a 'nice-to-have' optional feature.

AdamIsrael:
Thanks for the feedback, Nick! I've done some cleanup that will hopefully clear up your concerns.

petermatulis:
I'm looking this over. A few questions:

1. Why specify this?
 
> Two types of benchmark: specific and generic. The former tests application-specific functionality,
> such as measuring message queue latency or database query performance are specific to an
> application, whereas the later test things like memory, cpu, and disk i/o.

It can read like users are limited in some way.

2. The docs for `charm-benchmark` are empty but why point to an obscure site? Is that meant as a placeholder?

3. `Benchmark GUI` is mentioned without any supporting explanation.

marcoceppi:
The first is meant to describe scenarios, we'll reword.

Charm Benchmark docs aren't empty, they're hosted on read the docs I'll make sure @AdamIsrael updates to point to that documentation.

Benchmark GUI is being released this week, links and more information will be supplied when that is available

petermatulis:
I managed to create a new PR at #541 . Let me know what you guys want to do.

PR #518, opened 2015-06-24
Update incorrect tools mirror URL

NotBobTheBuilder:
Previously incorrect url, must have become outdated at some point during development.


marcoceppi:
This LGTM. @evilnick this will need to be backported to all released versions of the docs

PR #531, opened 2015-06-25
review: src/en/authors-relations.md

petermatulis:
Not so fast...

I usually tried to update #444 when I began work on a file, to avoid collisions.


evilnick:
hurrah again then !

PR #544, opened 2015-07-01
added docs on 1.24 tagging

### LINKED ISSUE #495: Add new section on "resource tags"
The 1.24 release introduced some changes to the way we tag cloud resources. Initially this only covers AWS and OpenStack. Below is a copy & paste from the release notes. This should be added to the docs somewhere, but I'm not sure where.

----

Juju now tags instances and volumes created by the EC2 and OpenStack providers with the Juju environment UUID, and related Juju entities. Instances will be tagged with the names of units initially assigned to the machine. Volumes will be tagged with the storage-instance name, and the owner (unit or service) of said storage.

Juju also adds any user-specified tags set via the "resource-tags" environment setting to instances and volumes. The format of this setting is a space-separated list of key=value pairs:

    resource-tags: key1=value1 [key2=value2 ...]

These tags may be used, for example, to set up chargeback accounting. Any tags that Juju manages will be prefixed with "juju-"; users must avoid modifying these.

Instances and volumes are now named consistently across EC2 and OpenStack, using the scheme `juju-<env>-<resource-type>-<resource-ID>`, where `<env>` is the human-readable name of the environment as specified in environments.yaml; `<resource-type>` is the type of the resource ("machine" or "volume") and `<resource-ID>` is the numeric ID of the Juju machine or volume corresponding to the IaaS resource.

evilnick:
fixes #495


evilnick:
@petermatulis I am :+1: myself on this as i want to push a 1.24 release now we have the 1.24 stuff more or less covered, but do let me know if you find anything and we can fix in situ

PR #546, opened 2015-07-01
Update commands.md

evilnick:
update the commands reference for 1.24 (updates #543)


evilnick:
this may need some fiddling with formatting, it is tricky to tell until it is merged

PR #547, opened 2015-07-07
New HTML for better collapsing nav.

kadams54:
Tidy up the nav HTML and make it more consistent with the production site. Then change the HTML structure to better support having multiple levels of collapsing nav.


evilnick:
Looks good :+1:

PR #548, opened 2015-07-07
Revert "New HTML for better collapsing nav."

kadams54:
This reverts commit b2618a4ae9d79d0b3fe08c9e9feed3dafd7f9cd4. We need to revert until corresponding changes are made in the hosting app, otherwise the `devel` docs will be broken in production.


evilnick:
oops.

Issue #247, opened 2015-01-30
new add-machine options need to be added


evilnick:
e.g. juju add-machine lxc:1 -n 2

PR #552, opened 2015-07-10
Update charms-deploying.md

wallyworld:
Doc changed to describe the new support for placement directives with the --to option as used with deploy and add-unit.


evilnick:
fixes #247

thanks!

petermatulis:
How does this fix #247 again?

Issue #100, opened 2014-05-15
local provider needs some docs for KVM


evilnick:

petermatulis:
Fixed in #553.

PR #553, opened 2015-07-13
issue #100; local provider KVM

petermatulis:
I still need to add "using LXC containers within a KVM guest". Suggestions on where to put that?


evilnick:
Hmmm. i guess that should go on the config-KVM page, as it only applies to local KVM (and possibly MAAS I guess)

evilnick:
otherwise :+1:  :custard:

PR #558, opened 2015-07-16
Add some howto docs for systems and users.

howbazaar:
This is a first cut of some docs around the new Juju Environment System changes that will be behind a feature flag for 1.25.

### DIFF
--- a/src/en/howto-systems.md
+++ b/src/en/howto-systems.md
@@ -1,0 +1,203 @@
+Title: How to configure Juju to hold multiple environments
+
+# Enable the Juju Environment System feature flag
+
+In order to use the multiple environment features of JES, you need to enable
+a development feature flag:
+
+```bash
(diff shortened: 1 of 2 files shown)


src/en/howto-systems.md
```diff
@@ -0,0 +1,203 @@
...
+new environment has no machines, and no services.
+
+```bash
+$ juju status
+environment: test
+machines: {}
+services: {}
+```
+
+# Colaboration with other users
```
axw: Collaboration

evilnick:
This looks good - thanks very much

There are a few things that need tidying up, but I think it is easier if I merge it and then tidy (I'll copy you in to the PR). I will do that tomorrow, as i also want to move it to the new 'Experimental' section of the docs and re-jig a few other things. 

thanks for all the words!

PR #566, opened 2015-07-17
reorg of JES additions

evilnick:
slight rework of @howbazaar 's uncustomarily  excellent words. this is still very much a draft - some of the styling needs redoing and parts need explanation, but it's worth putting live as a draft IMO


src/en/wip-systems.md
```diff
@@ -31,9 +39,95 @@ practice to create additional environments for real workload deployment.
...
-like landscape and nagios.
+like Landscape and Nagios.
+
+
+## Creating environments in a Juju System
+
+The `juju system create-environment` command will create a new empty
+environment that has the current user as the owner. These secondary
+environments are called 'hosted environments', as they are hosted within a
+Juju System. When creating an environment, the user will need to specify cloud
```
petermatulis: System is capitalized here but not on line 29.

src/en/wip-systems.md
```diff
@@ -31,9 +39,95 @@ practice to create additional environments for real workload deployment.
...
+API server is not accessible, and replaces the older `juju destroy-environment 
+--force` command.
+
+The command will first attempt to destroy all the environments through the
+API, but will fall back to destroying the machines using direct cloud provider
+calls if the API is not accessible.
+
+If there were hosted environments running that had cloud machines allocated to
+it, and the API server was not accessible, the machines will need to be
+manually reclaimed through the cloud providers tools.
```
petermatulis: provider's

src/en/wip-users.md
```diff
@@ -1,16 +1,16 @@
-Title: How to work with users in Juju
+Title: Managing users in Juju
 
-# User overview
+# Managing users in Juju
 
 Juju has understanding of two different types of users:
```
petermatulis: Suggestion:

"Juju has two types of users:"

petermatulis:
Otherwise LGTM :sake:

evilnick:
I think we will have to revisit this extensively - the language is very confusing, mainly because the 'new thing' has been called Juju (unavoidably ubiquitous word) Environment ( has many many meanings with Juju) System (has many many meanings everywhere). Also we seem to have invented a new word for the boostrap node or state-server or (possibly other names are also used). It makes it pretty impossible to give precise instruction on anything when there isn't a simple and definitive way to refer to objects in this universe.

Issue #217, opened 2015-01-06
juju logging needs its own topic page


chuckbutler:
We have a ton of logging information under troubleshooting. While that's great - logging is a topic all unto itself and should reside on its own topic page in the user docs with proper topic headings.

The developer logging can be summed up by explaining the different logging levels and `juju log`, which I believe is already in the author docs.

petermatulis:
I'm intending on creating a new page (under Troubleshooting in TOC):

troubleshooting-debug-log.md

And transferring stuff from:

* [troubleshooting-debug-log.md](https://jujucharms.com/docs/stable/troubleshooting.html) (include link to new page)
* [authors-hook-debug.md](https://jujucharms.com/docs/stable/authors-hook-debug.html) (include link to new page)

In the sense of 'juju-log' being "developer logging" I don't find much info available, certainly no logging levels:

* [authors-charm-writing.md](https://jujucharms.com/docs/stable/authors-charm-writing.html)
* [authors-hook-environment.md](https://jujucharms.com/docs/stable/authors-hook-environment.html)

Do you mean logging levels for debug-log given on [commands.md](https://jujucharms.com/docs/stable/commands.html)?

petermatulis:
@evilnick
Comments?

evilnick:
wrt the logging levels, I believe chuck is saying we should have some charm author docs on writing the logs and using appropriate levels (?). This uses the juju-log hook tool (https://jujucharms.com/docs/stable/authors-hook-environment#hook-tools)
I'm not sure if the current usage info set there is complete (it only mentions two levels) because of course, if it changed nobody bothered to tell me about it.

petermatulis:
I've consolidated the debug-log stuff. As for juju-log, I'm not sure what else can be done. A charm can send a message to the log (default level: INFO) but can set it as level: DEBUG using --debug.

howbazaar:
On 24/07/15 08:23, Peter Matulis wrote:
> I've consolidated the debug-log stuff. As for juju-log, I'm not sure
> what else can be done. A charm can send a message to the log (default
> level: INFO) but can set it as level: DEBUG using --debug.

the charm log command can specify the log level explicitly using

 -l or --log-level

These values are case insensitive and are: trace, debug, info, warning,
error.

Also look at 'juju help logging'

Tim

petermatulis:
'juju help logging' mentions a CRITICAL level as well which is not available to 'juju debug-log', which is strange.

[...]

PR #570, opened 2015-07-23
issue #217; consolidating debug-log

petermatulis:
I could not get the examples done using markdown ordered lists without losing the code formatting. I could indent the actual text/commands but then the multi-line blocks would end up on one line.

Also, in troubleshooting-debug-log.md, the second 'Note' does not get formatted like the other two do (locally anyway).

Feel free to make up for my lackings.

### DIFF
--- a/src/en/troubleshooting-debug-log.md
+++ b/src/en/troubleshooting-debug-log.md
@@ -1,0 +1,118 @@
+Title: Troubleshooting with debug-log
+
+# Troubleshooting with debug-log
+
+When problems arise the first step in determining the cause is to look at the
+logs. The `debug-log` command shows the consolidate logs of all Juju agents
+running on all machines in the environment.
+
+The output is a fixed number of existing log lines (number specified by
+possible options; the default is 10) and a stream of newly appended messages.
+Both existing lines and appended lines are filtered by any specified options.
+The exception to the streaming is when limiting the output (option `--limit`;
+see below) and that limit is attained. In all other cases the command will need
+to be interrupted with `Ctrl-C` in order to regain the shell prompt.
+
+For complete syntax, see the [command reference page](./commands.html).
+
+You can also learn more by running `juju debug-log --help` and `juju help 
+logging`.
+
+
+## Examples:
+
+**#1** - To begin with the last ten log messages:
+
+```bash
+juju debug-log
+```
+
+**#2** - To begin with the last thirty log messages:
+
+```bash
+juju debug-log -n 30
+```
+
+**#3** - To begin with all the log messages:
+
+```bash
+juju debug-log --replay
+```
+
+**#4** - To begin with the last twenty log messages for the 'local' environment:
+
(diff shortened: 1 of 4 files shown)


src/en/troubleshooting-debug-log.md
```diff
@@ -0,0 +1,118 @@
...
+For complete syntax, see the [command reference page](./commands.html).
+
+You can also learn more by running `juju debug-log --help` and `juju help 
+logging`.
+
+
+## Examples:
+
+**#1** - To begin with the last ten log messages:
+
```
evilnick: why are we numbering the examples? As they aren't referred to anywhere?

src/en/troubleshooting-debug-log.md
```diff
@@ -0,0 +1,118 @@
...
+**#9** - To see all WARNING and ERROR messages in the entire log:
+
+```bash
+juju debug-log --replay --level WARNING
+```
+
+**Note**: The `level` option restricts messages to the specified log-level or
+greater. The levels from lowest to highest are TRACE, DEBUG, INFO, WARNING, and
+ERROR.
+
```
evilnick: Can we use the "!!! Note:" syntax. I know this way works, but !!! is the accepted markdown conventionwhereas **Note:** exists only in our docs :)

src/en/troubleshooting-debug-log.md
```diff
@@ -0,0 +1,118 @@
...
+```bash
+juju debug-log --replay --exclude-module juju.state.apiserver
+juju debug-log --replay --exclude-module juju.state
+juju debug-log --replay --exclude-module juju
+```
+
+**Note:** The `include-module` and `exclude-module` are used to select/deselect the
+type of message displayed based on a (dotted) module name. The module name can
+be truncated.
```
evilnick: I would take this out of the note and add it as an explanation above.

evilnick:
Nice job. The only serious issue i have is with all the numbered examples. The numbers don't add anything IMO, they just make the page longer. If you think the individual examples should have some sort of heading (I don't), then a meaningful name would be better, and searchable. I think overall we are aiming for 'accessible guide' rather than 'terse reference work'.

petermatulis:
Edits made.

Numbered examples can be useful for external references to the docs:

"Look at example #4 on page troubleshooting-debug-log."

evilnick:
If they are only for reference, headers serve just as well - better in fact, as they actually mean something. E.g. you could consider grouping examples under a header ("using log levels"), which then automatically has an anchor in the text to link to

petermatulis:
There are not enough examples to group. Anyway, the reference I referred to is to an individual example.

evilnick:
If there aren't enough to group then it isn't going to be difficult to find :)

petermatulis:
Are you gonna press the big green button? :santa:

evilnick:
:octocat:

PR #571, opened 2015-07-23
When using underscores in the config keys, the GCE  provider stacktraces

chuckbutler:
to the user. The intended effect was a working provider. The culprit
appears to be underscores in the example config that were copied 1:1.

The python script also generates underscores in the yaml keys - minor
warning added to convert to dashes, as my python foo of .replace did not
work.

### DIFF
--- a/src/en/config-gce.md
+++ b/src/en/config-gce.md
@@ -181,13 +181,16 @@ information is to run something like:
 python -c 'import sys, yaml, json; yaml.safe_dump(json.load(sys.stdin), sys.stdout, default_flow_style=False)' < gce-auth-file.json 
 ```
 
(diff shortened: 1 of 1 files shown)


chuckbutler:
@evilnick There may be minor formatting you want to do here. The inline python actually yields a bogus config for the user as our provider code uses hyphens not underscores.

I've got a bug to track this issue here: https://bugs.launchpad.net/juju-core/+bug/1477712

What would be ideal is for the provider keys to be updated to read either/or.

evilnick:
Hmm. The example given is exactly the version I used to get a working GCE, but good catch if there is an issue with '_'. I will test again and see if I can't fix the python

PR #574, opened 2015-07-24
add note about '-' vs '_'

### LINKED ISSUE #572: GCE instructions have a pitfall for the unwary
GCE credentials present the keys as 'client_id:' and 'client_email:' whereas Juju wants 'client-id:'
This only matters if you cut and paste verbatim from the YAMLified JSON.

evilnick:
fixes #572, obsoletes #571

### DIFF
--- a/src/en/config-gce.md
+++ b/src/en/config-gce.md
@@ -196,7 +196,10 @@ private-key: |
   E2sZPXYMT8TmQnat35f4+h4FT5MYPHpr/OS27kejHLjxpWkbz3Dziqx6upM+fMiY
   ...
 ```
-!!! Note: The JSON file presents the keys as `client_email:`, `client_id:` and `private_key`, with underscores. Juju will not accept these keys, and they must be changed to `client-email:`,`client-id` and `private-key`, with dashes, in order to be recognised. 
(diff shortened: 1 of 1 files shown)


src/en/config-gce.md
```diff
@@ -196,6 +196,7 @@ private_key: |
   E2sZPXYMT8TmQnat35f4+h4FT5MYPHpr/OS27kejHLjxpWkbz3Dziqx6upM+fMiY
   ...
 ```
+!!! Note: The JSON file presents the keys as `client_email:`, `client_id:` and `private_key`, with underscores. Juju will not accept these keys, and they must be changed to `client-email:`,`client-id` and `private-key`, with dashes, in order to be recognised.
```
petermatulis: Why are these things called "keys"? Maybe something more generic like "parameters" or "variables".

"The JSON file contains the variables `client_email:`, `client_id:` and `private_key`. Juju will not accept the underscores and will need to be changed to dashes: `client-id`, ..."

evilnick: because 'keys' are what they are called in the context of a YAML file, I guess to distinguish them from the actual value. Perhaps it would be clearer if it said 'key names'

petermatulis:
LGTM :hamburger:

PR #577, opened 2015-07-27
adds prism js for local docs syntax highlighting

evilnick:


evilnick:
@petermatulis this adds the new syntax highlighting code to local builds. it required a change to the buildtool to rewrite <code> classes in the html.

petermatulis:
tried it; didn't notice a difference to be honest

fabricematrat:
you need to make sure the generate class contains language-yaml for example
The previous syntax higlighter did not support yaml.

evilnick:
It does, that's why I made changes to the build tool. This corks fine for me on pages with code

evilnick:
*works

evilnick:
@petermatulis can we merge this then?

petermatulis:
yeah, let's do it :ok:

Issue #260, opened 2015-02-09
Using Juju link points to a topic page of "introducing charms"


chuckbutler:
This is misleading, the topic page gives an overview of charms - not so much on "using juju" - which to me would be more of an overview on how to interface with juju, a brief introduction to charms, and links to sub-topic pages with a brief overview of the page contents/intent.

petermatulis:
I guess some stuff needs to be added *before* the "Introducing Charms" header. I agree that the page is jarring as is.

petermatulis:
@evilnick 
What do you think about renaming the page 'Juju Charms' and be done with it? The underlying file is already just called 'charms'.

petermatulis:
Confirmed today that the above will be done. Stand by.

petermatulis:
See PR #592

PR #589, opened 2015-07-29
issue #260; 'Using Juju' link

petermatulis:


evilnick:
Sorry to be a pain but i have thought about this and I really like having the word 'Using' in there as it helps separate it from the charm author stuff and is also more indicative of what is going on there.
Perhaps 'Using Juju Charms' or 'Using Juju & Juju Charms'?

petermatulis:
Using PR #592 now

Issue #86, opened 2014-04-24
Document environment-specific placement directives


axw:
There's now support (on trunk) for juju bootstrap/add-machine to take an optional environment-specific placement directive. MAAS is the only provider that currently supports such directives; we use this to bootstrap/add machines based on specific nodes in MAAS.

e.g.
    juju bootstrap <maas-name>
    juju add-machine <maas-name>

petermatulis:
we have some of this in [charms-ha](https://jujucharms.com/docs/devel/charms-ha.html) and in [commands](https://jujucharms.com/docs/stable/commands.html) but there is nothing specific regarding what providers support this and no caveats (a provider may or may not support it depending on implementation).

is augmenting the above two documents good enough or should we:

- start building out provider-specific docs (beyond installation)?
- include this in a provider's install/config page? (my preference)

petermatulis:
- Consensus in issue #187 to put this stuff in provider's install/config page.
- Still need to augment charms-ha and commands

petermatulis:
Submitted PR #596

petermatulis:
PR merged. Closing.

PR #596, opened 2015-08-07
issue #86; environment-specific placement

petermatulis:
- in commands.md:
i followed existing text and added:

`juju machine add --to myhostname`

- in config-maas.md:
existing text shows:

`juju add-machine hostname`

do we want to standardize: 'machine add' or 'add-machine'?

- in config-openstack.md:
i removed the link to the community wiki:

[https://help.ubuntu.com/community/UbuntuCloudInfrastructure](https://help.ubuntu.com/community/UbuntuCloudInfrastructure)

as i do not believe the community wiki to be an appropriate resource for our official documentation. this page in particular has not been updated in the last 16 months

### DIFF
--- a/src/en/charms-ha.md
+++ b/src/en/charms-ha.md
@@ -1,6 +1,8 @@
-Title:Ensuring High Availability (HA) for deployed services
+Title: Ensuring High Availability (HA) for deployed services
 
-# Service High Availability
+
+# Service High Availability (HA)
+
 
 ## Distribution groups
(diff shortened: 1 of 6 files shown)


evilnick:
This looks like good stuff overall.
I am not sure about the add-machine / machine add. The former used to be the only way to do it and will continue to be supported. "Machine add" follows the new command style and is more logical. I think for the amount of times we will use it, machine add has more advantages.

REMEMBER!!!! If you want to change text on the commands page, you really need to change the help text in juju core - when the commands page gets regenerated, your changes will be lost!

PR #602, opened 2015-08-12
Update bundles doc for V4 bundles

makyo:
V4 bundles are now the default for the GUI, the GUI's interactions with the deployer, juju-bundlelib, and the charmstore.  This updates the docs to reflect this.


petermatulis:
LGTM, thanks!

PR #608, opened 2015-08-18
Remove references to non-existant shell interface

skylerberg:
After reading the documentation, I attempted to use the `amulet`
command. It appears that this command was never implemented beyond a
point where it fails immediately. Let's remove this from the docs to
save others a little time figuring that out.


evilnick:
@marcoceppi ?

marcoceppi:
It's true https://github.com/juju/amulet/pull/87

evilnick:
Cool. Thanks @skylerberg !

PR #609, opened 2015-08-19
Fix spelling mistakes.

bac:
This branch uses the `spell` and `ispell` programs to find obvious misspellings among the sea of false positives.

My approach was to run `make spell > badlist` and then manually look at the humongous list to pick out the non-jargon words that were misspellings then use `git grep` to find the offending file, e.g.
`git grep percieve`

The ispell program can be run on that file to see if there are misspellings:
`ispell -p custom -d british src/en/authors-charm-hooks.md`

I did not change the release notes file, even though it was full of typos, because I thought it was historic. Feel free to clean it up if you wish.

I also changed some American to British spellings if I was editing the file anyway.  Others were left. They should probably all be cleaned up.

A makefile target for `clean` was added to get rid of generated stuff.

This branch is just a start to using available tools to improve the documents our users see.


huwshimi:
Looks like a heap of work! Thanks @bac :+1: No QA.

evilnick:
Many thanks for this, it is very useful. i think ultimately this is the kind of thing we should include in the build tool, which is why the new python tool has methods for pre and post processing. Then it could emit warnings or whatever as it processed the files.

Anyhow, great work. Some caveats:

I am in two minds about changing spellings in the boilerplate that comes from Juju - I think they are better fixed there first.

Similarly, the commands page is entirely generated from the help text within Juju, so the next time it gets regenerated, those changes will be lost - I guess we should try and fix those errors there.

A few of the things are style rather than mistakes, e.g. 'lifecycle'.

bac:
Thanks for the comments @evilnick.  Yes some things were changed that perhaps should have been left.  Perhaps we need another pass and come up with a single spelling for "lifecycle" etc.  It can be manually added to the 'custom' dictionary and will then not be flagged again.

I didn't know the commands file was generated.  We can use the changes shown here to feed back to juju-core to have them fix them properly.

Still I feel it was a worthwhile effort to get rid of some of the more egregious typos.

evilnick:
Oh, it was definitely worthwhile, thanks very much!
I will have a play with it and add to the custom dictionary.

PR #619, opened 2015-08-19
Changes related to JES

bac:
Note these files cannot be merged as is.  I have some sections

> Denoted
> like this

which raise questions/make suggestions.  This branch cannot be merged with those comments intact.


src/en/wip-systems.md
```diff
@@ -30,10 +30,10 @@ A Juju Environment System (JES), also sometimes shortened to 'Juju System',
...
 underlying database.
 
-This initial environment is also called the 'system environment' 
-(or 'bootstrap environment'), and is what is
-created when the bootstrap command is used.  This System environment is a
-normal Juju environment that just happens to have machines that manage Juju.
+This initial environment is also called the 'system environment' (or
+'bootstrap environment'), and is created when the bootstrap command is used.
+This System environment is a normal Juju environment that just happens to have
+machines that manage Juju.
```
evilnick: This whole section is confusing to me, particularly with the proliferation of uses for the terms system, environment and boostrap. Are we stuck with JES? Can't we have a Juju Environment Manager or a Juju Environment Controller instead? It becomes easier to differentiate what these entities are then, and the name is more descriptive

mitechie: The big thing is that we want the terms to match the cli experience. The cli is juju system and juju environment so things have been attempting to keep in line with the user's actual expectations.

src/en/wip-systems.md
```diff
@@ -45,19 +45,33 @@ like Landscape and Nagios.
...
+> Somewhere it should be noted that an initial juju environment must be
+> bootstrapped before 'system create-environment' can be called.
+> Later in the discussion of `juju system destroy` it becomes clear this
+> original, bootstrapped environment is the `system` and it must be
+> bootstrapped before proceeding.
+
 The `juju system create-environment` command will create a new empty
 environment that has the current user as the owner. These secondary
 environments are called 'hosted environments', as they are hosted within a
 Juju System. When creating an environment, the user will need to specify cloud
```
evilnick: does this mean 'when creating a **hosted** environment'. I think it does.

src/en/wip-systems.md
```diff
@@ -66,6 +80,12 @@ created environment "test"
 staging (system) -> test
 ```
 
+If your config file is improperly formatted you'll get an error like:
+
+```no-highlight
+ERROR json: unsupported type: map[interface {}]interface {}
+```
+
 When the environment has been created, it becomes the current environment. A
```
evilnick: does this mean when the **system** environment is created

src/en/wip-systems.md
bac: I think we need to figure out the terms and stick with them in this document to allow others to think about things properly.

Thing 1: the initially bootstrapped Juju that then hosts and manages subsequent environments.  

Thing 2: The result of 'juju system create-environment'.  It is the equivalent of a normal 'juju bootstrap' but lives on top of Thing 1.  You can create as multiples of these Things.

So it sounds like thus far we've been calling Thing 1 "System" and Thing 2 things "Environments".

User management happens at the System level but users can share environments among themselves.  I think my suggestion to show the full workflow (annotated) is the right way to describe what's happening.  Perhaps with a block diagram.

src/en/wip-systems.md
bac: Yes.  You provide the --config option to 'juju system create-environment' if the cloud provider for that type of environment requires them.  If you are creating an ec2 environment then you must provide a config. If creating a local environment then you do not.  Probably anything except local will require credentials.

src/en/wip-systems.md
evilnick: I'm +1 on a diagram. 

it seems we have been calling Thing1 the "system _environment_" in places, which I think is unpleasantly confusing.

src/en/wip-systems.md
evilnick: Yes, the part about creating credentials needs a lot more explanation, with a lot more examples

evilnick:
The current document also doesn't touch on things which are no doubt changed:

High availability
juju backups
juju block

PR #639, opened 2015-08-24
nav changes

evilnick:
Introduces a new section on managing juju, distinct from the 'recipes' of the HowTo section. moves some files, renames some things.


evilnick:
I think all the things which ought to be in docs which don't relate to charms specifically - e.g. Juju sysadmin stuff, should go in the new section. Then the Howto section can be for proper HowTo things, like how to run a tomcat app and such

PR #643, opened 2015-08-26
Update storage docs for 1.25

axw:
A bunch of changes ahead of the 1.25 release:
 - storage is no longer considered experimental in 1.25. I'd like to discourage the use of storage in 1.24, so I've just called out that it is stable in 1.25.
 - there are some new storage providers (azure, gce)
 - new storage-list hook tool (I've added docs for all storage hook tools)
 - moved persistence to "future work" section. Most of the inner workings are complete, but this currently lacks a UI for attaching/detaching, so it is not usable.
 - removed "implemented features" section, which was mostly just a check list against the spec, and not meaningful to users. Moved bits out to main sections.

I've also renamed from wip-storage to storage, and moved it in the nav tree. We should probably separate user vs. charm author bits later.


evilnick:
This looks good, thanks for the extra detail. :+1: 

After I have merged this I will start work on integrating this better with existing docs - e.g. the hook tools should go with the others.

petermatulis:
@evilnick 
Re provider-specific stuff, we should link from a provider's section on the Install/Configure page to the corresponding section on the storage page.

Issue #642, opened 2015-08-25
Typo's on Private Cloud docs


chrisjohnston:
https://jujucharms.com/docs/stable/howto-privatecloud

I believe the following are both typo's:

the <path_url> value as described above in Tools Metadata Contents

the <path_url> value as described above in Image Metadata Contents

(copy and paste failed, currently on the docs website they show up as LT and GT instead of < and >)

chrisjohnston:
Could this be pushed to the stable docs as well please:

https://jujucharms.com/docs/stable/howto-privatecloud

petermatulis:
Done :sleeping: . Please confirm.

chrisjohnston:
Looks good. Thanks

PR #644, opened 2015-08-26
Update howto-privatecloud.md

evilnick:
final(!) fix for private cloud <> problem


petermatulis:
Dealt in #642 via direct edit.

PR #647, opened 2015-08-27
More lxc caching

petermatulis:


evilnick:
Jebus, seems a bit complicated (the system, not your explanation)

:+1:

PR #656, opened 2015-08-31
for issue #265; remove stuff + some nav; under manage + little on 'st…

### LINKED ISSUE #265: "Removal with Juju" documentation needs more detailed explanation
The [removal with Juju documentation](https://juju.ubuntu.com/docs/charms-destroy.html#removing-services) talks about the commands like `juju remove-service <service-name>` without actually explaining what the commands do. For example, it doesn't say that the `juju remove-service` calls the stop hook.

I have asked an askubuntu question about this and received an [excellent answer](http://askubuntu.com/a/584746/172367). I think this information should be included in the documentation about each command.

petermatulis:
…op hook'

Fixes #265 

Note: In [Managing Environments](https://jujucharms.com/docs/devel/charms-environments.html) we should probably remote the bullet with '-e' as I learned that this switch is deprecated. I wasn't sure enough to do this deletion as it may affect other things on that page.

### DIFF
--- a/src/en/charms-destroy.md
+++ b/src/en/charms-destroy.md
@@ -1,112 +1,128 @@
 Title: Removing services, units and environments in Juju
 
-# Removal within Juju
 
-Juju isn't just about magically spinning up services as and when you need them,
-it is also about quickly, sanely and efficiently removing something when you no
-longer need it. This section deals with the sort of things you can ruthlessly
-destroy, and how to go about it.
+# Removing services, units and environments
 
-## Removing Services
+Juju can sanely and efficiently removing something when you no longer need it.
+This section looks at how to remove services, units and environments.
 
-Once a service is no longer required it can be removed with a simple command.
+
+## Removing a service
+
+Once a service is no longer required it can be removed with:
 
 ```bash
-juju remove-service <service-name>
+juju destroy-service <service-name>
 ```
 
-**Warning!:** Removing a service which has active relations with another running service will break that relation. This can cause errors in both services, as such review and potentially remove any relationships first.
+**Warning!:** Removing a service which has active relations with another
+running service will break that relation. This can cause errors in both
+services. Make sure you review this aspect and, if required, remove the
+relations first.
 
-A service can take a while to "die", but if running a juju status reveals that
-the service is listed as dying, but also reports an error state, then the
-zombied service will still be hanging around. See caveats for how to manage
-services in a dying state.
+This is the order of events for removing a service:
 
-**Note:** Destroying a service removes that service, but not any nodes which may have been created for it to run on. This is juju's way of preserving data to the best of its ability. See Destroying Machines for additional details.
+1. The Juju client tells the state server that every unit (in this service) is
+to be destroyed.
+1. The state server signals to the service (charm) that it is going to be
+destroyed.
+1. The charm breaks any relations to its service by calling `relationship-broken`
+and `relationship-departed`.
+1. The charm calls its 'stop hook' which **should**:
+    - Stop the service
+    - Remove any files/configuration created during the service lifecycle
+    - Prepare any backup(s) of the service that are required for restore purposes.
+1. The service and all its units are then removed.
 
-## Removing Units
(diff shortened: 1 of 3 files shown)


src/en/charms-destroy.md
```diff
@@ -1,112 +1,128 @@
...
-# Removal within Juju
 
-Juju isn't just about magically spinning up services as and when you need them,
-it is also about quickly, sanely and efficiently removing something when you no
-longer need it. This section deals with the sort of things you can ruthlessly
-destroy, and how to go about it.
+# Removing services, units and environments
 
-## Removing Services
+Juju can sanely and efficiently removing something when you no longer need it.
```
evilnick: /removing/remove/

src/en/charms-destroy.md
```diff
@@ -1,112 +1,128 @@
...
-Once a service is no longer required it can be removed with a simple command.
+
+## Removing a service
+
+Once a service is no longer required it can be removed with:
 
 ```bash
-juju remove-service <service-name>
+juju destroy-service <service-name>
 ```
```
evilnick: please revert this, the command is 'remove-service' (as with remove-unit). The only thing which should get destroyed now is an environment

src/en/charms-destroy.md
```diff
@@ -1,112 +1,128 @@
...
+and `relationship-departed`.
+1. The charm calls its 'stop hook' which **should**:
+    - Stop the service
+    - Remove any files/configuration created during the service lifecycle
+    - Prepare any backup(s) of the service that are required for restore purposes.
+1. The service and all its units are then removed.
+
+A service can take a while to "die", but if `juju status` reveals that the
+service is listed as dying, but also reports an error state, then the removed
+service will not go away. See section 'Caveats' below for how to manage services
```
evilnick: perhaps "See the 'Caveats' section below"

src/en/charms-destroy.md
```diff
@@ -1,112 +1,128 @@
...
-## Removing Machines
+## Removing machines
 
-Instances or machines which have no currently assigned workload can be removed
-from your cloud using the following command:
+Machines (instances) can be removed like this:
 
 ```bash
-juju remove-machine <number>
+juju destroy-machine <number>
```
evilnick: please revert this also

src/en/charms-destroy.md
```diff
@@ -1,112 +1,128 @@
...
+
+Once a service is no longer required it can be removed with:
 
 ```bash
-juju remove-service <service-name>
+juju destroy-service <service-name>
 ```
 
-**Warning!:** Removing a service which has active relations with another running service will break that relation. This can cause errors in both services, as such review and potentially remove any relationships first.
+**Warning!:** Removing a service which has active relations with another
```
evilnick: Now would be a good time to update these to the proper admonition style, which is:

!!! Warning:
!!! Note: 

etc

src/en/charms-destroy.md
```diff
@@ -1,112 +1,128 @@
...
+associated instances and the bootstrap node itself:
 
 ```bash
 juju destroy-environment <environment>
 ```
 
-This will completely remove all instances running under the current environment
-profile. As there is no 'undo' capability, some further safety precautions have
-been added to this command.
+**Note:** The '-e' switch is deprecated. The environment should be specified as
```
evilnick: I would rather explain this better, if it is an issue. You reference the -e switch, but it hasn't been mentioned up until now, and indeed some people have never used a version with '-e' in the destroy-environment command.

Perhaps something like:
"!!! Note Older versions required the use of the environment switch ('-e') with this command. This switch is no longer required for this command

evilnick:
See comments!

evilnick:
P.S. The '-e' switch is not deprecated, only its use in 'destroy-environment'

petermatulis:
Edited.

evilnick:
Cool. Just FYI, the 'delete-' variants of those commands were swapped for 'remove-' versions about a year ago, I believe because remove was deemed less scary.

:+1:

PR #658, opened 2015-09-01
Document Composing a Charm from Layers

johnsca:
Charm tools 1.6.0 has been released, which includes composer.  Add documentation for using it.


src/en/authors-charm-composing.md
```diff
@@ -0,0 +1,262 @@
...
+$Configuration['Database']['Name'] = '{{ db.database() }}';
+$Configuration['Database']['User'] = '{{ db.user() }}';
+$Configuration['Database']['Password'] = '{{ db.password() }}';
+?>
+```
+
+And that's all you need to get Vanilla up and running using layers!
+The final directory structure looks like this:
+
+![directory tree](./media/author-charm-composing-01.png)
```
marcoceppi: I'd argue having the file tree as actual text wrapped in a code block instead of a picture. It'll make it easier to update overtime, but that's just an opinion of mine.

src/en/authors-charm-composing.md
```diff
@@ -0,0 +1,262 @@
...
+
+Check out the [repo][] for the complete charm layer.
```
marcoceppi: null link?

src/en/authors-charm-composing.md
```diff
@@ -0,0 +1,262 @@
...
+
+It is important to note that we didn't have to create any hooks.  Those were
+all handled by the other layers and interfaces.  This is common when using
+composition and relation stubs.
+
+It's also worth noting that there is a file for each layer in the `reactive`
+directory.  This allows the handlers for each layer to remain separate and
+not conflict.  All handlers from each of those files will be discovered and
+dispatched according to the [discovery and dispatch rules][].
```
marcoceppi: null link?

src/en/authors-charm-composing.md
johnsca: Yeah, I'd prefer that as well, but I wanted to get the tree structure and coloring included, which didn't seem possible otherwise.  This is the same in the existing walk-through charm.

petermatulis:
Thanks @johnsca , can you double-check the hyperlinks? I've found a few broken ones early on.

johnsca:
@petermatulis I fixed the authors-charm-writing link(s) that was mistakenly pointing back to the composing page, but I didn't see any other broken links.  Was that the one you were referring to?

petermatulis:
@johnsca 
There is 'hand', 'before', 'Charm Tools'. May be more. Are the first couple placeholders?

johnsca:
I fixed "hand" and "before" pointing back to the the current doc instead of the writing doc, as I mentioned, but the Charm Tools link works for me.  If you're referring to the empty brackets, I'm using implicit reference links (as documented at http://daringfireball.net/projects/markdown/syntax#link), but they all seem to be working.  (For reference, I have this branch being served at http://52.0.202.26:8000/en/authors-charm-composing.html)

PR #659, opened 2015-09-01
config-manual: Bring to the top the native support note

mattyw:
I think it would be nice to bring this "native support option" to the top of the docs rather than hide it at the bottom. Just a thought


evilnick:
Ok, Cool :cake:

PR #668, opened 2015-09-09
Bundle command line description was completely missing form the doc

tyloch:
I've  spent today a few hours trying to find, document and  explain to  the user how to deploy bundle using command line. 
GUI is important but  day to day operations will be based on command line and API calls. 

Local deployment via command line described below is a more charm developer focused. 
While for the charm user, the command line deployment from juju store is the most common case.

### DIFF
--- a/src/en/charms-bundles.md
+++ b/src/en/charms-bundles.md
@@ -20,6 +20,8 @@ If you simply want to download some you can
 [search the store for bundles](https://jujucharms.com/q/?type=bundle)
 too.
 
+
+
 ### Local import to Juju GUI
 
 The easiest way to import a bundle into the GUI is by dragging the bundle file
@@ -32,7 +34,11 @@ masthead. After clicking the button you’ll be prompted to select the bundle
 file. Once a file is selected the process is the same as the drag-and-drop
 method.
 
-### Local deploy via command-line
+### Command-line deployment 
+
+#### Deployment from Juju charm store 
+
+The easiest way to use bundle is to deploy it from command line. 
 
 A bundle file can be deployed via the command-line interface by using the `juju
 quickstart` tool.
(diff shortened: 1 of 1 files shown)


src/en/charms-bundles.md
```diff
@@ -32,7 +34,11 @@ masthead. After clicking the button you’ll be prompted to select the bundle
 file. Once a file is selected the process is the same as the drag-and-drop
 method.
 
-### Local deploy via command-line
+### Command-line deployment 
+
+#### Deployment from Juju charm store 
+
+The easiest way to use bundle is to deploy it from command line.
```
mitechie: "The easiest way to use a bundle is to deploy it from the command line."

src/en/charms-bundles.md
```diff
@@ -44,6 +50,12 @@ running the following command:
 sudo apt-get install juju-quickstart
 ```
 
+E.g this bundle https://jujucharms.com/u/jean-deruelle/mobicents-restcomm-mysql-bundle on top of the page  you will find command line command you can use to deploy this bundle.
```
mitechie: I'd suggest going with something simpler like:

https://jujucharms.com/turbo-lamp

and maybe just stick to the link and the command vs the 'top of the page' part.

evilnick:
Thanks for the effort here. I have converted this into an issue for the docs team to solve as there are a couple of other considerations to take into account (e.g. there is no quickstart for Windows, etc).

src/en/charms-bundles.md
tyloch: simpler is fine.
However mobicents-restcomm-mysql-bundle deploy  some nice WebRTC video call service / workload. 
Much better for demo.  up to you

PR #672, opened 2015-09-11
Adds initial revision of the charming with docker, layers, and reactive narrative

chuckbutler:
@evilnick  This document probably has quite a bit of feedback coming.

This is the first draft of our 'charming with docker' story that leverages quite a few new utilities, namely composer and reactive framework. 

Feedback is welcome


src/en/authors-charm-with-docker.md
```diff
@@ -0,0 +1,325 @@
+# Charming with Docker
+
+You have a Docker container and you heard about Juju.  Juju can deploy your
```
mbruzek: We should probably remove this line.  If they are in the Juju documentation they know about Juju.

mbruzek:
+1 LGTM Great start on a Docker document @chuckbutler

PR #676, opened 2015-09-15
Remove link that breaks document generator.

bac:
The removed link is an <a> with no href.  It is unclear why it was added.

Unfortunately it found a bug in the docs generator for jujucharms.com and has caused the generation to fail since its inclusion.

I am fixing the offending line so that docs can be generated again and creating a task to make the generator more robust.


chuckbutler:
+1 LGTM

Thanks for catching this @bac

PR #693, opened 2015-09-23
storage: clarify storage-get

axw:
Clarify storage-get's purpose, i.e. that it is
expected to be used in both storage-attached
and storage-detaching.

Also, update statement about ordering of hooks
relative to install/upgrade hooks. We do not
run storage hooks before upgrade anymore.


axw:
@cholcombe973 FYI

cholcombe973:
+1

evilnick:
:pear:

PR #709, opened 2015-10-22
Added release 1.24.7.

sinzui:
Note that the links to former stable 1.24.6 were pointing to the unsigned windows installer. We can see that 1.24.6 has inverted download counts for the signed an unsigned installers when we compare other releases. I am pretty certain users are using these pages to find the windows installer.


evilnick:
At least they are being read then :smiley:

Issue #713, opened 2015-10-26
backport compose/build changes to 1.24


evilnick:
https://github.com/juju/docs/commit/4f119c47bc95dc48f3a2f00dcf52ec46e4948f06

evilnick:
@johnsca we'll backport these changes to the current stable docs too

PR #714, opened 2015-10-26
Updated docs to reflect terminology change from compose to build

johnsca:
Backport to address #713


evilnick:
actually, this looks good - thanks !

PR #731, opened 2015-11-06
note change to default instance type

natefinch:
in 1.25, we changed how the default instance type was determined, so that we default to m3.medium (the previous default, m1.small, is problematic since they're being phased out by Amazon).


natefinch:
we may want to put a corresponding line in older versions of the doc, specifying that we default to m1.small.

petermatulis:
Backport: being explicit about instance type

petermatulis:
I backported to 1.24 (m1.small) and 1.25 (m3.medium).

PR #737, opened 2015-11-09
Update docs to match logging changes in 2.0

mjs:
- removed mention of all-machines.log as this is gone
- removed mention local provider differences as the local provider is gone
- prioritise debug-log over looking at log files directly
- removed irrelevant detail that made the logging docs harder to understand
- more detail about advanced log filtering
- document the new debug-log --no-tail option

### DIFF
--- a/src/en/troubleshooting-logs.md
+++ b/src/en/troubleshooting-logs.md
@@ -1,5 +1,4 @@
-Title: Viewing Juju logs  
-
+Title: Viewing Juju logs
 
 # Viewing logs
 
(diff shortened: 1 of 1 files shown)


petermatulis:
I object to the removal of so much material. It helps, in a troubleshooting context, to know about how log files and agent files get created and where they reside. If all-machines.log is no longer used we can remove mention of that.

Also, [LP #1484688](https://bugs.launchpad.net/juju-core/+bug/1484688) does not look fixed so there is no reason to remove the text that describes it.

mjs:
@pmatulis: Sorry, I hadn't noticed your comments here and had completely forgotten about this PR.

There isn't actually very much removed. The diffs make it look like I removed more than I did because I changed the ordering of the sections so that the debug-log usage information came before the more detailed explanations of how specific log files work - we hope that people will debug-log in most situations, with the hosts residing on the machines as a fallback.

The docs have changed so this PR no longer applies cleanly. I'll revise it today.

pmatulis:
@mjs 
Yes, the docs were rearrange very recently. I also just made an amendment with https://github.com/juju/docs/pull/1059.

pmatulis:
The current upgrade docs were updated a while ago and seem fine to me. I'm closing this. Please re-read the current docs and re-submit if I missed anything.

PR #738, opened 2015-11-09
Added tools/serve.py

mjs:
The default HTTP server used by SimpleHTTPServer is single threaded which can make refreshes of documentation pages a bit slow. This PR adds a little script which starts a multi-threaded HTTP server for
viewing docs. It also takes the directory to serve on the command line, and avoids the traceback when Ctrl-C is used.

The "serve" make target has been updated to use the new script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-
 build:
 	tools/mdbuild.py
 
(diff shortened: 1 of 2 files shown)


evilnick:
I was never troubled by the speed of a single-threaded server, but thanks!

PR #744, opened 2015-11-20
authors-charm-with-docker.md: Fixed broken link

mattyw:


evilnick:
Hi, I'm going to reject this one (not your fault) I just noticed that whomever wrote the original used an absolute link to a version of docs, which we try and avoid (what happens when that version gets archived, turned into 1.26 or whatever)



So, thanks for the heads up, i  will fix this bit myself!

Also, fwiw, while I was checking I found a whole load of other incorrectly expressed links!

PR #751, opened 2015-11-30
Update authors-charm-actions.md

battlemidget:
Fix the example for using the object keys and makes it much clearer on how to utilize the param properties.


chuckbutler:
:+1:   I like it

PR #756, opened 2015-12-07
Update tools-amulet.md

mattyw:
This is the best practice way of accessing units in the sentry


mbruzek:
+1 That looks correct to me.

marcoceppi:
LGTM, thanks for getting this updated

PR #764, opened 2015-12-24
Modified install hook, fixes #762.

### LINKED ISSUE #762: authors-charm-building contains incorrect code
The current documentation for composing the vanilla charm in authors-charm-building.md is broken.
Without applying the recursive chmod to the /var/www/vanilla/conf directory, the following error is displayed, and the charm does not deploy successfully.
screenshot from 2015-12-24 00 46 23

A simple fix could be to call chmod -R 777 /var/www/vanilla/conf from the install hook.

jamesbeedy:
Not setting recursive perms on /var/www/vanilla/conf causes an
error.

The error can be seen here: https://github.com/juju/docs/issues/762.

### DIFF
--- a/src/en/authors-charm-building.md
+++ b/src/en/authors-charm-building.md
@@ -146,6 +146,7 @@ react to our database going away in an intelligent manner:
 ```python
 import pwd
 import os
+from subprocess import call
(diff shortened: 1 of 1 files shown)


evilnick:
Thanks!

PR #771, opened 2016-01-08
Fix typo in vanilla layer example doc: s/layer.yaml/layers.yaml

### LINKED ISSUE #754: layers.yaml instead of layer.yaml
https://jujucharms.com/docs/1.25/authors-charm-building

Talks about the layers.yaml file. It should be layer.yaml

1stvamp:
Title says it all.

Fixes #754.


chuckbutler:
:-1:  its actually `layer.yaml`

1stvamp:
Doh, this was caused by doing [stable doc](https://jujucharms.com/docs/stable/authors-charm-building) the other evening, thinking "oh I'd better fix that", then coming back and doing the inverse against the devel doc which is already correct. :raised_hands:

PR #790, opened 2016-01-13
Developer event cycle

AdamIsrael:
Minor cleanup

### DIFF
--- a/src/en/developer-event-cycle.md
+++ b/src/en/developer-event-cycle.md
@@ -1,9 +1,9 @@
 # Event Cycle
 
- Taking a broad look at multiple services there are several common events that
-each service goes through install, configure, start, upgrade, and stop.  To
+ Taking a broad look at multiple services, there are several common events that
+each service goes through: install, configure, start, upgrade, and stop. To
 orchestrate services, Juju drives these events that a charm can respond to with
-executable files called hooks.  Juju executes the specific hook for the
+executable files called hooks. Juju executes the specific hook for the
 appropriate event.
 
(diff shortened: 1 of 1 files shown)


chuckbutler:
:+1:  LGTM

evilnick:
:strawberry:

src/en/developer-event-cycle.md
```diff
@@ -38,7 +38,7 @@ Reference section of the Juju documentation.
 
 When deploying a charm, there is a guaranteed set lifecycle events that every
 charm will run - this is the basic series of hooks executed in the initial
-deployment cycle.  
+deployment cycle.
```
mbruzek: Actually in this case the two spaces at the end of a line have a meaning to start a new paragraph.  I have had some problems with numbered lists right after text paragraphs so I use the two spaces on purpose as I did here.

src/en/developer-event-cycle.md
```diff
@@ -61,16 +61,16 @@ When a relation is added from the vanilla charm to a database charm, the first
 event is `database-relation-joined` the two units know about each other and the
 code should prepare communication between the two. After the join a
 `database-relation-changed` state is set, the two units have peer to peer
-communication.  
+communication.
```
mbruzek: Again used the trailing spaces on purpose to create a new paragraph.

mbruzek:
Hey Adam thanks for fixing this stuff I use the trailing double space on purpose here to create new paragraphs when doing the editing.  Some markdown converters are more strict on this than others.

PR #793, opened 2016-01-14
Developer layer example

AdamIsrael:
Fix some papercuts and edit for clarity

### DIFF
--- a/src/en/developer-layer-example.md
+++ b/src/en/developer-layer-example.md
@@ -1,9 +1,9 @@
 ## Writing a layer by example
 
-In this howto, we will be writing a charm layer for Vanilla. Vanilla is an open
-source themeable, pluggable, multi-lingual forum software, which enables
-communities to host a forum for discussion topics at scale. Powered by PHP and
-MySQL Vanilla is a fine example of a 3 tiered application
+In this document we will be writing a charm layer for Vanilla. Vanilla is an
+open source, themeable, pluggable, and multi-lingual forum software, which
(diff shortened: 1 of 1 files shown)


src/en/developer-layer-example.md
```diff
@@ -1,9 +1,9 @@
 ## Writing a layer by example
 
-In this howto, we will be writing a charm layer for Vanilla. Vanilla is an open
-source themeable, pluggable, multi-lingual forum software, which enables
-communities to host a forum for discussion topics at scale. Powered by PHP and
-MySQL Vanilla is a fine example of a 3 tiered application
+In this document we will be writing a charm layer for Vanilla. Vanilla is an
+open source, themeable, pluggable, and multi-lingual forum software, which
```
chuckbutler: This feels like a runaway run-on now

chuckbutler:
Minor comment left inline, otherwise :+1: LGTM

src/en/developer-layer-example.md
AdamIsrael: First sentence and the dropped comma?

chuckbutler: Yeah, its 2 logical thoughts mashed together with verbs and nouns and all that goodness. Maybe i'm being too picky, when read aloud it sounds fine.

AdamIsrael: I don't think it's technically wrong but I lean towards more commas in general.

evilnick:
:frog:

PR #796, opened 2016-01-14
Fix papercuts

AdamIsrael:
Cleanup some papercuts (spacing, case, etc) and tweak readability


mbruzek:
These changes look fine to me. +1

PR #797, opened 2016-01-14
Papercuts

AdamIsrael:
Fix a few spacing and readability issues


mbruzek:
+1 LGTM. I will have to force myself not to use double spaces after a period from here on out.

PR #801, opened 2016-01-15
2.0 lxd provider

petermatulis:


src/en/config-LXD.md
```diff
@@ -0,0 +1,204 @@
...
+ - Due to the above, superuser privileges (sudo) are no longer required.
+ - There is no more "Juju cache" for images. LXD now has its own cache
+   (`/var/lib/lxd/images`) which is separate from the traditional LXC cache
+   (`/var/cache/lxc`).
+
+Although this document does provide enough information to get you running with
+LXD itself (this document is ultimately about running LXD with Juju) it does
+so assuming you are setting up LXD for the first time.
+
+**Note:** If you are running iptables (firewall) or even an iptables frontend
```
mbruzek: Recommending that we turn off the firewall is not going to fly with some companies.  Do we know what ports to disable in ufw to get LXD to run?

petermatulis: Yeah, this was taken from the existing config-LXC. I never looked into it.

src/en/config-LXD.md
```diff
@@ -0,0 +1,204 @@
...
+such as `ufw`, the LXD local provider might not work properly. Troubleshoot
+accordingly or stop the firewall altogether.
+
+
+# Prerequisites and installation of Juju and LXD
+
+ - The LXD local provider will not work on Ubuntu 12.04 LTS and backporting to
```
mbruzek: It seems the convention for this document is NN.NN LTS (codename). I don't see the convention used here.  Can we use (Precise) and (Trusty) here?

src/en/config-LXD.md
```diff
@@ -0,0 +1,204 @@
...
+
+With standalone LXD, images are available via either a local or network-based
+store. These can be called a local image store *remote*, a **local remote**, or a
+networked image store remote, a **non-local remote**. Juju, however, only
+supports local remotes at this time. Images stored on them must also have the
+names (aliases) in the format `ubuntu-<codename>` such as 'ubuntu-trusty' and
+'ubuntu-xenial'.
+
+The first time an image is needed the store will supply it and any subsequent
+requests will be satisfied by the LXD cache. In this way, a store is sollicited
```
mbruzek: s/sollicited/solicited/

src/en/config-LXD.md
```diff
@@ -0,0 +1,204 @@
...
+
+LXD itself logs to `/var/log/lxd/lxd.log` and Juju machines created via the
+LXD local provider log to `/var/log/lxd/juju-{uuid}-machine-{#}`. However,
+the standard way to view logs is with the `juju debug-log` command. See
+[Viewing logs](./troubleshooting-logs.html) for more details.
+
+
+# Other useful commands
+
+There is a cornucopia of commands available. Some common ones not yet covered
```
mbruzek: While "cornucopia" is correctly used here it stuck out when I read it.  Did "plethora", "abundance", "multiple", or "many" not work in this context?

mbruzek:
@petermatulis thanks for letting me take a look at this.  Line 54 satisfies the problem I encountered.  I found a few other minor suggestions, but it looks great!

src/en/config-LXD.md
mbruzek: Perhaps the core developers could give us some feedback on which ports to enable on ufw which is installed on Ubuntu server by default IIRC.

src/en/config-LXD.md
```diff
@@ -0,0 +1,204 @@
...
+binary actually comes from the `lxd-client` package.
+
+
+# Configure and bootstrap
+
+If this is a new Juju install then you do not yet have a
+`~/.juju/environments.yaml` file. Create one with
+
+```bash
+juju generate-config
```
evilnick: this goes away in 2.0

evilnick:
looks good apart from the fact that some bits will not exists in 2.0. I recommend we land it with othe fixes and then update it when we know how the authentication model will actually work

src/en/config-LXD.md
petermatulis: To be clear, the disabling of the firewall was meant as a temporary troubleshooting technique only. That should be clarified.

Issue #747, opened 2015-11-26
Update needed: GCE provider


mitechie:
Copied from: https://github.com/CanonicalLtd/jujucharms.com/issues/185 per @loslosbaby

This page: https://jujucharms.com/docs/stable/config-gce

...is great, but there are issues:
- the screen captures are out of date
- the JSON for the OAuth has tags that have to be deleted before juju bootstrap will work

juju bootstrap fails with:
Bootstrap failed, cleaning up the environment.
ERROR there was an issue examining the environment: invalid config: key "auth_uri" not supported

The tags that have to be deleted are:
- auth_uri
- token_uri
- auth_provider_x509_cert_url
- client_x509_cert_url

Alternately, juju can be modified to ignore the tags.  I'll let you decide if a bug needs to be filed.

evilnick:
Hi @loslosbaby, thanks for the report

Yeah, sadly the GCE page changes quite a lot, but we will update the images. I am more concerned about the keys needing to be changed. Just to clarify, this is when you use the option of specifying the downloaded JSON with 

auth-file: /home/juju-user/.juju/Juju-GCE-7e04ea455a2d.json

or similar?

sheepeatingtaz:
I can confirm this. As well as the items listed above, I also had to remove "project_id" from the json when using the auth-file option.

chuckbutler:
Couple things here:

When i added the GCE provider and created a service account, the juju GCE provider forced me to remove ~ 6 keys from my config.json. This is problematic. 

I feel that the juju config scanner should only care that we have the 4 required keys, the rest should simply be ignored.

The error messaging around this could be more helpful too:

`ERROR there was an issue examining the environment: invalid config: key "auth_uri" not supported`
`ERROR there was an issue examining the environment: invalid config: key "project_id" not supported`

This was the output from 1.26-alpha3-trusty-amd64

evilnick:
I agree that it is a problem that should be fixed in juju rather than in docs.
However, I know a lot of this will be changing in 2.0, so I expect it is easier to document workarounds for now until the problem solves itself.

petermatulis:
Comment ( @tych0 ) from duplicate #811 

===
referring to: https://jujucharms.com/docs/1.25/config-gce

The screenshots for where to click and such are off, since google has (yet again, I'm sure :) changed the interface. You need to create a "Service Account" instead of just an Oauth 2.0 Client account as it says; only a service account gets the private key that juju understands.

[...]

PR #823, opened 2016-01-23
for issue #747; include warning until repaired

petermatulis:


evilnick:
You might want to backport that to stable also!

Issue #825, opened 2016-01-25
Docs need a 'new in 2.0' page


evilnick:
We need to have a guide for existing users, transitioning to 2.0

evilnick:
This has been started as per #826, but various parts still need to be confirmed

PR #826, opened 2016-01-25
Introduction to 2.0

evilnick:
part of #825 

Adds a new page for 1.x users moving to 2.0


evilnick:
N.B., this references a page which is currently missing! See #827

src/en/introducing-2.0.md
```diff
@@ -0,0 +1,75 @@
+Title: Introducing Juju 2.0
+TODO: Add extra changes as required
+      Test upgrade instructions
+
+# Introducing Juju 2.0
+
+If you are familiar with Juju, you are no doubt excited to get started with the
+wealth of exciting new features introduced in the new 2.0 release (and if you 
+haven't tried Juju before, see [this introduction](./about-juju.html) ). But
```
petermatulis: Let's not use "excite" twice.

src/en/introducing-2.0.md
```diff
@@ -0,0 +1,75 @@
...
+ - **Commands:** Probably the most immediate change you will come across is that
+practically every command has changed! Really! Significant work has been done to
+make the commands more consistent, logical and easy to understand. For example, 
+instead of ```juju authorised-keys add <xxxx>``` you will now use the command
+   ```juju add-ssh-key <xxxx>```    
+   It reads better, is more straightforward, and 
+works really well with tab completion! The documentation has been updated to
+include all these new commands, so you can check relevant pages for guides to
+the new commands, or browse through the [reference guide](./commands.html).
+There is also a handy [crib sheet](./command-changes.html) to show you the new
```
petermatulis: ... sheet at the end of this document that lists the new command names ... (?)

src/en/introducing-2.0.md
```diff
@@ -0,0 +1,75 @@
...
+is to do with the way Juju organises itself - now multiple models (environments)
+can be driven from one controller (state-server). This means that for a given
+cloud you only need one controller, but can drive several models. You should 
+also check through the (multi-user)(./juju-multiuser-environments.html) 
+documentation to understand how these models can be shared amongst multiple
+users.
+
+  - **Providers:** As well as updates to the Azure provider and the new 
+Rackspace provider, juju now makes extensive use of LXD. This makes 'local' 
+models leaner and faster and also brings benefits to placing containers within
```
petermatulis: Juju

src/en/introducing-2.0.md
```diff
@@ -0,0 +1,75 @@
...
+space on the controller, depending on the environment configuration.
+
+### If you are using a version < 1.25.2
+
+You must upgrade Juju to the latest 1.x series version of Juju first:
+
+```bash
+juju upgrade-juju --version=1.25.2
+```
+The proceed as below:
```
petermatulis: Then proceed: (?)

petermatulis:
Review complete.

src/en/introducing-2.0.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Upgrading from earlier versions
+
+!!! Note: This is pre-release documentation - upgrading does not currently work!
+
+Before upgrading from an earlier version, ensure that there is sufficient disk 
+space for an upgrade. Upgrading could consume an additional 500MB to 1GB of 
+space on the controller, depending on the environment configuration.
+
+### If you are using a version < 1.25.2
```
cherylj: @evilnick there's a lot of discussion right now around supporting upgrades from 1.25 to 2.0.  Given some of the direction in the last sprint, we cannot guarantee upgrades yet, so they're not going to be allowed until some of the details are worked out.

src/en/introducing-2.0.md
evilnick: yeah, i was going to put it there, but it is too big and ungainly. will fix

src/en/introducing-2.0.md
evilnick: Sure. This is more or less placeholder stuff until the details are clarified. We need to start writing something!

evilnick: Obviously, we are hoping there is a clear/simple upgrade process, but if not we will obviously change this

Issue #780, opened 2016-01-13
Docs needed: Azure changes


evilnick:
The azure provider is rewritten for 2.0, particularly important are the support for placement and resources.

petermatulis:
I need help from Juju team to get this provider to work.

petermatulis:
I'm re-writing [config-azure.md](https://github.com/juju/docs/blob/master/src/en/config-azure.md) but I don't believe that it's the place to put usage of the Azure provider's features such as 'placement'. My understanding of these config-* pages is that their objective is to supply the minimum information to get started with a provider.

A random examination of some other config-* pages do not show provider-specific features either. How should we deal with this in general? Perhaps sub-pages for each provider is the best way.

I'm not sure what is meant by 'resources' (description of this issue).

petermatulis:
Closing. MP #837 merged.

PR #837, opened 2016-01-27
for issue #780

petermatulis:


src/en/config-LXD.md
```diff
@@ -192,8 +193,10 @@ the standard way to view logs is with the `juju debug-log` command. See
 There are many commands available. Some common ones not yet covered
 are given below.
 
-client commands					| meaning
-------------------------------------------------|----------------------
+<style> table td{text-align:left;}</style>
```
evilnick: this looks familiar :)

src/en/config-azure.md
```diff
@@ -1,123 +1,223 @@
...
+TODO: Decide on what to do with provider-specific features (e.g. placement).
 
-This process requires you to have a Windows Azure account. If you have not
-signed up for one yet, it can obtained at http://azure.microsoft.com/en-us/.
 
-You should start by generating a generic configuration file for Juju, using the
-command:
+# Overview
+
+Juju now supports Microsoft Azure's Resource Manager API. The Azure provider
```
evilnick: I sort of feel that this whole paragraphe should be a note or something - this is information for existing Azure users

src/en/config-azure.md
```diff
@@ -1,123 +1,223 @@
...
+azure help
+azure config mode arm
+azure login
 ```
 
-This will generate a file, `environments.yaml`, which will live in your
-`~/.juju/` directory (and will create the directory if it doesn't already
-exist).
+You will be prompted to visit a website to enter the provided code. It will
+therefore be easier to perform this on a (graphical) Ubuntu desktop.
```
evilnick: try to be non-platform specific! Windows, Centos, MacOS...

src/en/config-azure.md
```diff
@@ -1,123 +1,223 @@
...
+The options `--name`, `--home-page`, and `--identifier-uris` are arbitrary but
+you should use values that make sense for your environment.
+
+Note the application ID, the **APP_ID**. It will look similar to:
+
+```bash
+APP_ID=f6ab7cbd-5029-43ef-85e3-5c4442a00ba8
+```
+
+Use the APP_ID to create a server principal:
```
evilnick: what is a 'server principal'? I mean, we can guess, but is this handily defined somewhere on the Azure site?

src/en/config-azure.md
```diff
@@ -1,123 +1,223 @@
...
+
+### `subscription-id`
+
+List your account and get the subscription ID, the **SUB_ID**:
 
-The generic configuration sections generated for Windows Azure will look
-something like this:
+```bash
+azure account list
+```
```
evilnick: could we include an example of the output?

src/en/config-azure.md
```diff
@@ -1,123 +1,223 @@
...
+
+Set its value in the terminal. It will look similar to:
```
evilnick: I would change this a bit to make sure people understand that they are fetching a value from above, and setting an environment variable to be equal to that (e.g. Mac  users may not understand what these things are).

Something like, "This value can now be stored in a local environment variable, by entering:

xxxxxxxxx

src/en/config-azure.md
```diff
@@ -1,123 +1,223 @@
...
-
-You can read more about
-[Azure Availability Sets](http://azure.microsoft.com/en-gb/documentation/articles/virtual-machines-manage-availability/).
+        storage-account-type: Standard_LRS
+        location: East US
+        subscription-id: f717c8c1-8e5e-4d38-be7f-ed1e1c879e1a
+        application-password: some_password
+        application-id: f6ab7cbd-5029-43ef-85e3-5c4442a00ba8
+        tenant-id: daff614b-725e-4b9a-bc57-7763017c1cfb
+```
```
evilnick: might be best to leave the actual yaml bits for azure in, eg:
   azure:
         type:azure
         ...
         ...

evilnick:
Wow, this is a real PITA, but good work figuring it all out

PR #839, opened 2016-01-28
Fixed some typos and replaced the tree image

bac:


bac:
The `charm build` tool expects a file called `layer.yaml` but the docs referred to it as `layers.yaml.`

The image of the output of `tree` in the vanilla example had the old `composer.yaml` instead of `layer.yaml`.

marcoceppi:
Lgtm :boat:

chuckbutler:
+1 LGTM

Thanks for catching these bac

PR #854, opened 2016-02-04
upgrade build tool to python3

evilnick:
final(?) changes for python3 mdbuild - faster, better, won't break in xenial!


petermatulis:
1. i installed a Xenial daily
1. copied over this branch
1. make sysdeps
1. make build

Built quickly and without error.

PR #861, opened 2016-02-07
fix broken links in developer-debug-dhx.md

ns950:
The document developer-debug-dhx.md has 2 broken links as well. Erroneous local (./) path had been provided. This addresses Issue 838 (which I neglected to mention in earlier clean-up).

### DIFF
--- a/src/en/developer-debug-dhx.md
+++ b/src/en/developer-debug-dhx.md
@@ -4,9 +4,9 @@
 scripts that extend the functionality of Juju -- which allows you to fully and
 automatically customize the machines created by Juju, making developing and
 debugging hooks as painless as possible. It is drop in replacement for
-[`juju debug-hooks`](./developer-debugging.html#the-'debug-hooks'-command),
+[`juju debug-hooks`](developer-debugging.html#the-'debug-hooks'-command),
 recommended for more advanced users who need to
-[debug hooks](./developer-debugging.html) repetitively.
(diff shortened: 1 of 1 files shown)


marcoceppi:
LGTM :boat:

evilnick:
Thanks for this. Just to clarify, relative links like this in the docs aren't actually broken, though it is inconsistent with the style

ns950:
Oh, interesting. It showed up as a 404 for me, which is how I noticed it. I've been reading the docs, and for any missing stuff or breakage, going to see if the bug already exists in the issue tracker or if I needed to file a new one for it (or just fix it). As the doc was there with no other obvious typo, I tried removing the relative path and it worked locally for me (the local server to test is awfully useful, btw!). I probably should have looked a little harder to see what was going on.

evilnick:
Thanks. We try to keep the local render as close as possible to the real site (sadly replicating the complex javascript isn't always possible, so the navigation doesn't work at the moment).
I think the reason this 404s in the local rendering is to do with the local server setup rather than the link. I will check into that. So, we tend NOT to use ./, but there is no reason it shouldn't work, and it does in the live docs

ns950:
Ah, that explains it. I'm reading the online docs for the most part, but probably wandered into a window which had the local server running a page which I had tested some earlier fix on. Too many windows open...Thanks for the explanation! I'll be more aware of that in future.

PR #862, opened 2016-02-07
fix broken links in author-charm-store.md

ns950:
Updating link issues in the document author-charm-store.md.  The links to the Charms and Bundles tutorials were pretty badly borked. The link text and address were reversed and were pointing to hardcoded stable external links. I'm assuming the desired/correct behavior here would be to point to the package source files being distributed with this file. It still better than the current 404 users get on the documentation...


marcoceppi:
Thank for you for updating these links, LGTM! :boat:

marcoceppi:
@evilnick potential candidate for backporting

Issue #776, opened 2016-01-10
Errors in all https://jujucharms.com/docs/devel/ pages.


malevasquez:
All pages within https://jujucharms.com/docs/devel/ have the following errors:

	Line 126 <a href="#canonlist" class="nav-toggle no-script">
		The anchor #canonlist does not exist in the page
		
	Line 496 <li><a href="developer-getting-started#prerequisites-and-tools">Prerequisites and Tools</a></li>
		The anchor #prerequisites-and-tools does not exist in the referred page, the nearest anchor in meaning is #install-libraries-and-tools. If this is the correct one, the line should be
		<li><a href="developer-getting-started#install-libraries-and-tools">Prerequisites and Tools</a></li>
			
	Line 517 <li><a href="developer-layers-interfaces#communication-scopes">Communication Scopes</a></li>
		The anchor #communication-scopes does not exist in the referred page, however the anchor #what-communication-scopes-are,-and-how-to-use-them exists, it should be changed to
		<li><a href="developer-layers-interfaces#what-communication-scopes-are,-and-how-to-use-them">Communication Scopes</a></li>
			
	Line 526 <li><a href="developer-leadership#leadership-howtos">Leadership How-Tos</a></li>
		The anchor #leadership-howtos does not exist in the referred page however the anchor #leadership-howto's does, it could be
		<li><a href="developer-leadership#leadership-howto's">Leadership How-Tos</a></li>
			
	Line 529 <li class="section"><a class="header" href="developer-action">Implementing Actions</a>
		Page https://jujucharms.com/docs/devel/developer-action does not exist the right name of the page has an s at the end, it should be
		<li class="section"><a class="header" href="developer-actions">Implementing Actions</a>
			
	Line 531 <li><a href="developer-actions#defining-actions">Defining Actions</a></li>
		The anchor #defining-actions does not exist in the referred page, the most similar one is #actions-for-the-charm-author.
			
	Line 550 <li><a href="developer-debugging-layers">Debugging Layers</li>
		Page https://jujucharms.com/docs/devel/developer-debugging-layers does not exist
			
	Line 579 <li><a href="authors-charm-store#submitting">Submit a charm</a></li>
		The anchor #submitting does not exist in the referred page, however the anchor #submitting-a-new-charm does, it could be
		<li><a href="authors-charm-store#submitting-a-new-charm">Submit a charm</a></li>

PR #863, opened 2016-02-07
Navigation box clean up. Broken links, incorrect anchors, some header consistency.

ns950:
There were some 404 links in the navigation box and some anchors that weren't referenced correctly. I fixed and tested all the links in the navigation box. This does not address all/any style or writing issues or consistency issues, but I did add a few missing headers  and changed a few for consistency reasons so that navigation becomes easier. 

There appears to be some issue with having a hyphen in the anchor name (howto vs how-to). I didn't try to resolve the weirdness I was seeing. The code that converts a header implicitly into an anchor drops any hyphens so it should have worked, so not sure what was going on. The current header had 'Leadership How-To's' which in any case is incorrect (the plural was intended, rather than the possessive case, I imagine, so if 'howto' is an acceptable noun, 'howtos' would be the correct plural). Python and Drupal and most other documentation use 'howto' rather than 'how-to'  or 'how to'.  Not sure if Juju docs has a position on this. 

Issue #776 (and possibly others, I haven't checked) is fixed by this. 

This also shouldn't have had the developer-debug-dhx.md fix in here (which is already in tree).

### DIFF
--- a/src/en/config-azure.md
+++ b/src/en/config-azure.md
@@ -2,7 +2,7 @@ Title: Juju Azure provider
 TODO: Decide on what to do with provider-specific features (e.g. placement).
 
 
-# Overview
+# Configuring for Azure 
 
 The new Azure provider is backwards compatible with the previous provider but
 supports several additional features, in particular, support for unit placement
--- a/src/en/config-gce.md
+++ b/src/en/config-gce.md
@@ -1,7 +1,7 @@
 Title: Juju GCE provider
 
 
-# Overview
+# Configuring for the Google Compute Engine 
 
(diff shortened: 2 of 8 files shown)


marcoceppi:
Thanks for the patches! Most of this LGTM, I'll leave it to @petermatulis or @evilnick to address the more subjective bits as they lead up a lot of the documentation work

petermatulis:
LGTM

@ns950 We actually do use './' quite a bit in the docs. Let's leave those aside in future PRs.

If you'd like to dig your teeth in a little more we track "easy" tasks here:

https://github.com/juju/docs/labels/easy

Thanks a bunch for your efforts in improving the Juju docs!  :candy:

PR #877, opened 2016-02-09
Misc

ns950:
A few trivial typo cleanups. Also made the text "command documentation" into a link to the command reference doc (assuming that's the one intended?) in getting-started.md. As the link "Juju commands" is way down below in the Reference section, it's not easily seen, and it's not obvious where the Getting Started developer guide is referring you to.


petermatulis:
LGTM

Thank you! :dolphin:

Issue #859, opened 2016-02-05
Need openstack provider storage example!


jamesbeedy:
I hear rumors that storage capability exists for the openstack provider, but the are no examples for how to use storage for the openstack provider. Can we have this knowledge portrayed in juju's storage documentation?

Thanks!

evilnick:
There are (AFAIK) no specific special settings for OpenStack/Cinder

Do the generic pool creation tools not work for you? (https://jujucharms.com/docs/1.25/storage#preparing-storage)

jamesbeedy:
Thats correct, they do not, possibly I am missing something (like example usage). Would you mind sharing the command that you use to provision storage using the openstack provider?

evilnick:
Hmm, no, it isn't very helpful on that front. I will do some digging and get some examples for you. I believe you need to create a pool, but there isn't a clear indication of how to do that.

axw:
@jamesbeedy the OpenStack provider's default storage is just called "cinder", and there should be a pool called that by default. You can either specify the pool name "cinder" in the `--storage` constraints, or you can just leave it off and it'll be used as the default.

As Nick says, there is configuration for the cinder storage provider in Juju. You should still be able to create a pool, I think, but you may as well just use the default one.

jamesbeedy:
@axw What your depicting is slightly concerning on multiple fronts..... Would you mind sharing such a command that works for you as an example? And for a huge +1 for an example addition to the documentation for storage for openstack provider.

jamesbeedy:
@axw per @evilnick and the docs --> "The OpenStack/Cinder provider does not currently have any configuration."

jamesbeedy:
@evilnick @axw Cinder provides block storage devices, not pools .....

evilnick:
Thanks @jamesbeedy. i think there are 2 issues here:

 - 1. Juju doesn't actually behave as expected with cinder/OS storage. E.g. the command:  
```juju storage pool list```
      does not actually return any pools for OpenStack at all, even though, as @axw suggests, there is a default 'pool' called cinder, so for example: ```juju deploy cs:~axwalk/postgresql --storage data=cinder,10G``` does work, allocating a 10G volume in cinder which does show up with ```juju storage volumes list``` and with ```cinder list```
 - 2. The current instructions do not show
       * This particular case for OpenStack
       * an example of ```juju storage add <unit>``` for OpenStack
[...]

PR #878, opened 2016-02-09
Storage additions

evilnick:
Some small changes to the storage doc to add a simple example for OpenStack/Cinder
also removed old storage.md file and updated out of date links in developer docs.

Also removed all the annoying double spaces after a full point in a paragraph. These are ignored by Markdown anyway, but they annoy me every time I see them.

Part of work for #859


petermatulis:
LGTM :space_invader:

PR #890, opened 2016-02-12
New commands reference

evilnick:
My diff is too large!


evilnick:
@pmatulis this is autogenerated stuff, so there may be a few things wrong with it that need patching up. Am not too bothered, as will have to do it again for the release, but it shows us where all the problem bits are and may be useful to experimenters

evilnick:
- belay that request, the code formatting bug has struck - need to fudge all the examples!

PR #894, opened 2016-02-15
added temporary release notes

evilnick:
added the release notes as a reference. These are temporary until the actual release notes for 2.0 are available.


evilnick:
This just patches in the already published alpha release notes, so I am going to use the authority of being the only one awake to merge this also

PR #903, opened 2016-02-17
updated set-config --to-default

evilnick:


evilnick:
I am only copying you on this so you are aware. In spite of what other documents may say, the way to reset config of a charm is

    juju set-config --to-default <xyz>

NOT
 
    juju set config --default <xyz>

pmatulis:
LGTM, merging.

PR #908, opened 2016-02-19
Fix header capitalization to be consistent; a few other capitalization/other cleanups.

ns950:
There were a lot of inconsistent capitalizations not only across files but within the same file and even within a header. This patchset fixes up all the headers to use standard sentence capitalization (as was done for the titles and was the case for the majority of the files). There were a couple of other capitalization/consistency/typo fixups also thrown in where spotted (ongoing effort).

Still remaining:
- clean up use of 'Charm Store' and 'charm store' where appropriate
- clean up use of 'Charm Tools', 'Charm Helpers' (the latter two are confusing pages due to variance in name of feature, name of library, name of plugin, etc.).

### DIFF
--- a/src/en/actions.md
+++ b/src/en/actions.md
@@ -1,6 +1,6 @@
 Title: Juju actions  
 
-# Juju Actions
+# Juju actions
 
 Juju Charms can describe Actions that users can take on deployed services.
 
--- a/src/en/authors-charm-actions.md
+++ b/src/en/authors-charm-actions.md
@@ -1,6 +1,6 @@
 Title: Implementing actions in Juju charms  
 
-# Actions for the Charm author
+# Actions for the charm author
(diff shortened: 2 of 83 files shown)


ns950:
Also remaining:
- fix up use of Action vs action

ns950:
Added commit to fix links in contributing.md (to file and view documentation bugs).

ns950:
Added a src files-wide edit to fix lower case 'juju' to 'Juju' where it refers to the product, not command syntax, for clarity.  Added a few command quotes where they were missing to ensure it was clear a command was being referred to. (e.g. juju set --> `juju set`). Apologies for the length of these patches.

evilnick:
i think most of this is fixed now anyway. Thanks for the help!

PR #909, opened 2016-02-20
Convert first h2 header to be h1 in about-juju.md

ns950:
All the files in the documentation have an h1 level header except this one. For consistency and clarity (especially as this is likely the first file/page that a new user reads), make the initial header a top level (h1) header. It helps to have the same look/feel across pages and doesn't lead you to confusion about whether you're at the top/start or not, and not missing content.

### DIFF
--- a/src/en/about-juju.md
+++ b/src/en/about-juju.md
@@ -95,7 +95,7 @@ If you want, you can also share your charm with the world and gain feedback and 
 
 What are the best workloads to try with Juju?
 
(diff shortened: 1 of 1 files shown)


ns950:
Added additional minor clean-ups in about-juju.md:
- link to Juju Charm Store
- fixed an occurrence of erroneous lower case juju where it should be Juju (the product, not the command)

evilnick:
thanks :cake:

PR #916, opened 2016-02-29
Renamed interface name from 'website' to 'http'.

### LINKED ISSUE #912: incorrect interface name
In https://jujucharms.com/docs/devel/developer-layers-interfaces#implementing-the-requires-side , the interface name 'website' is used. I think this should be 'http' to match all the previous examples, as the 'http' interface is what has been discussed and 'website' would be the relation name.

seman:
Resolve #912


marcoceppi:
LGTM :boat:

PR #918, opened 2016-03-03
Issue 781 config rackspace.md

pmatulis:
I used *define a model* as opposed to *create a model* because I thought *someone* has a fetish towards define (as seen in the proposed navigation menu).


src/en/controllers.md
```diff
@@ -0,0 +1,40 @@
...
+servers and underlying database.
+
+Once a controller model has been created you can define multiple models within
+that environment. See [Models](./models.html).
+
+When a controller is intended to be used by multiple people, it is recommended
+that Juju's [multi-user functionality](./juju-multiuser-environments.html) be
+leveraged.
+
+It is not possible to create a second controller model for the same cloud
```
evilnick: this is literally true but misleading. You can create a cloud (e.g. rax2) and then have a new controller in that

evilnick:
there are probably all sorts of objectionable things here, but I would rather now get these things kicked off so we can build on them so i am going to :+1:  I'm sure this isn't the last time we will be touching these files in the coming weeks!

PR #922, opened 2016-03-09
Update Amulet page to point to official package docs.

tvansteenburgh:
Now that we have official package docs for Amulet, including
auto-generated API docs, the jujucharms.com docs should simply
point to those, ensuring that these docs will stay up-to-date
with the actual code and not lag behind, as they have tended
to do in the past.


marcoceppi:
LGTM :boat:

Issue #913, opened 2016-02-24
Lack of layer.yaml, charm build examples


stub42:
~~There seems to be no documentation on how to declare that a layer requires an interface.~~ (Found one example)

There is information on writing layers, but seems to be no mention of how to run 'charm build' to generate a charm that can actually be used.

chuckbutler:
https://jujucharms.com/docs/devel/developer-getting-started#assemble-the-layers

Maybe this needs to be duplicated elsewhere?

chuckbutler:
@johnsca and i riffed on this a bit, This breaks down into a few things that need to be documented:

### Baselayer

> The baselayer has a [Readme](https://github.com/juju-solutions/layer-basic/blob/master/README.md), and we should leverage that for documenting its role in the pipeline.

- Wheelhouse support
- config states @when('config.changed')
- layer options for custom packages
- venv support 

### Layer.yaml

- Layer Options
- Yaml modifications (Deletes)
- Custom Tactics
- File Ignores

# Builder

- CLI Args
- Env Vars
- ?

PR #923, opened 2016-03-09
Adds a layer.yaml reference page.

chuckbutler:
This is a first pass, and consumes information from both the baselayer
readme and the charm-build docs in the source.

Prelim work for #913

### DIFF
--- a/src/en/reference-layer-yaml.md
+++ b/src/en/reference-layer-yaml.md
@@ -1,17 +1,19 @@
 Title: Charm Layer.yaml Reference  
 
-# Layer.yaml
+## Layer.yaml
 
 Each layer used to build a charm can have a `layer.yaml` file. The top layer
 (the one actually invoked from the command line) must. These tell the generator
-what do, ranging from which base layers to include, to which interfaces. They
+what do, ranging from which base layers, to which interfaces to include. They
 also allow for the inclusion of specialized directives for processing some
 types of files.
 
 
 ### Layer Options
 
-Any layer can define options in its `layer.yaml`. Those options can then be set by other layers to change the behavior of your layer. The options are defined using [jsonschema](http://json-schema.org/), which is the same way that action
+Any layer can define options in its `layer.yaml`. Those options can then be set
+by other layers to change the behavior of your layer. The options are defined
+using [jsonschema](http://json-schema.org/), which is the same way that action
 parameters are defined.
 
 For example, the foo layer could include the following option definitions:
@@ -37,7 +39,7 @@ options:
 
 ### Yaml Modifications
 
-config and metadata take optional lists of keys to remove from `config.yaml`
+Config and metadata take optional lists of keys to remove from `config.yaml`
 and `metadata.yaml` when generating their data. This allows for charms to,
 for example, narrow what they expose to clients.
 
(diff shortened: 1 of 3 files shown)


src/navigation.tpl
```diff
@@ -183,6 +183,7 @@
                   <li><a href="reference-charm-hooks.html">Juju Charm Hooks</a></li>
                   <li><a href="reference-environment-variables.html">Juju environment variables</a></li>
                   <li><a href="reference-hook-tools.html">Juju Hook Tools</a></li>
+                  <li><a href="reference-layer-yaml.html">Juju Hook Tools</a></li>
```
evilnick: I think you mean to call this link something else

src/en/reference-layer-yaml.md
```diff
@@ -0,0 +1,68 @@
...
+        - key names
+  metadata:
+    deletes:
+        - key names
+```
+
+
+### Custom Tactics
+
+Each file in each layer gets matched by a single Tactic. Tactics implement how
```
evilnick: WTF is a tactic?  Perhaps it would be better to lead with a brief summary of what 'tactics' are and why you should care about them - e.g. "When layers are combined in a final charm, the charm tools need to make decisions about what happens to each individual file in each layer. they do this using a set of rules known as 'tactics'...

Or whatever tactics actually are...

src/en/reference-layer-yaml.md
```diff
@@ -0,0 +1,68 @@
...
+# Layer.yaml
+
+Each layer used to build a charm can have a `layer.yaml` file. The top layer
+(the one actually invoked from the command line) must. These tell the generator
+what do, ranging from which base layers to include, to which interfaces. They
+also allow for the inclusion of specialized directives for processing some
+types of files.
+
+
+### Layer Options
```
evilnick: The headings *should* be sorted now, so h2s are good.

src/en/reference-layer-yaml.md
```diff
@@ -0,0 +1,68 @@
...
+
+Any layer can define options in its `layer.yaml`. Those options can then be set by other layers to change the behavior of your layer. The options are defined using [jsonschema](http://json-schema.org/), which is the same way that action
```
evilnick: I know I am being a git, but could you wordwrap :disappointed:

src/en/reference-layer-yaml.md
```diff
@@ -0,0 +1,68 @@
+Title: Charm Layer.yaml Reference  
+
+# Layer.yaml
+
+Each layer used to build a charm can have a `layer.yaml` file. The top layer
+(the one actually invoked from the command line) must. These tell the generator
+what do, ranging from which base layers to include, to which interfaces. They
```
evilnick: to which interfaces...

did you miss something out or is that it? which base layers and interfaces to include?

src/navigation.tpl
chuckbutler: Whoops! :+1:

chuckbutler:
Thanks for the rapid feedback Nick. I touched this up a bit after getting some feedback from you, Ben, and Cory.

chuckbutler:
The submission attempts to close out the ENV VARS of the builder and the layer.yaml guide. However the role of layer:basic is a bit bigger. @johnsca - Whats your take on splitting that out into #924, and dropping in a re-edit of layer:basic's readme?

johnsca:
+1 for splitting it out; `layer.yaml` and `layer:basic` are two different things and can be documented separately.

marcoceppi:
LGTM :boat:

PR #926, opened 2016-03-15
New clouds page

evilnick:
adds info on listing clouds, defining new clouds


src/en/clouds.md
```diff
@@ -0,0 +1,154 @@
...
+
+Juju ships with the ability to seamlessly use a number of public clouds
+(including Amazon Web Services, Azure, Google Compute Engine, Joyent and
+Rackspace) to deploy workloads. 
+
+The knowledge to run these clouds is 'baked-in', so for the majority of
+use-cases there is no additional configuration to be done - you can simply
+specify the cloud you wish to use, supply Juju with some
+[credentials][credentials] and start deploying services.
+
```
pmatulis: Mention the ability to also leverage private clouds like OpenStack. You mention it later (## Specifying additional clouds) but it would be good to say something at this time.

src/en/clouds.md
```diff
@@ -0,0 +1,154 @@
...
+!!! Note: juju can work with any OpenStack cloud, see the notes below for
+[specifying additional clouds](#specifying-additional-clouds)
+
+### Special clouds
+
+You will see, listed among the clouds Juju knows about, three special types of
+clouds: MAAS, LXD and Manual.
+
+  - **LXD:** This is the cloud you want to use if you are testing Juju or
```
pmatulis: I think this is an abuse of the word "cloud" but I won't insist on rewording.

src/en/clouds.md
```diff
@@ -0,0 +1,154 @@
...
+clouds: MAAS, LXD and Manual.
+
+  - **LXD:** This is the cloud you want to use if you are testing Juju or 
+  developing your own Juju charms - it is incredibly fast! 
+  [LXD is a container hypervisor][LXD-site] that runs on any Linux host, providing 
+  the ability to spin up containers on the host machine. For more details on
+  using LXD, please see the [LXD documentation][juju-lxd]
+  
+  - **MAAS:** An acronym of Metal As A Service, MAAS lets you treat physical
+  servers like virtual machines in the cloud. Rather than having to manage each
```
pmatulis: Ditto.

src/en/clouds.md
```diff
@@ -0,0 +1,154 @@
...
+The 'local:' prefix indicates that this is a cloud you have added yourself. 
+
+
+[credentials]: ./credentials.html "Juju documentation > Credentials"
+[LXD-site]: http://www.ubuntu.com/cloud/lxd "LXD"
+[juju-lxd]: ./cloud-LXD.html "Juju documentation > LXD"
+[maas-site]: http://maas.io "MAAS website"
+[juju-maas]: ./cloud-maas.html "Juju documentation > MAAS"
+[juju-manual]: ./cloud-manual.html "Juju documentation > Manual cloud"
+[yaml]: http://www.yaml.org/spec/1.2/spec.html
```
pmatulis: I'm not sure why we're using this type of link strategy. It's the first time I see it used and I find it inefficient for reading this source. I need to bounce somewhere else to know what the actual link is and then bounce back up to continue reading. I also favour a consistent way of doing things.

pmatulis:
Review complete. Fix and then merge. :candy:

src/en/clouds.md
evilnick: it is, but then, so is all of Juju!

src/en/clouds.md
evilnick: This is direct from the MAAS website. That is what we say it is

src/en/clouds.md
evilnick: Quite a few of the pages use this. It is easier to check them if you build the docs.

There are a few good reasons for doing it this way:
 - it makes the links easier to check and update (they are all at the bottom)
-  it prevents mistakes when you are using the same link more than once in a document
- it clears the actual text of the document - some of the links are, necessarily, very long, so it improves readability of the actual text

src/en/clouds.md
evilnick: ok

src/en/clouds.md
evilnick: I apologise if you find it harder to deal with, but I actually find it a lot easier to write and edit! Try it :)

pmatulis: It may be easier to write and makes the source neater but the logic is broken and that's what I don't like. I don't like bouncing around to understand stuff.

PR #932, opened 2016-03-22
changed maas.ubuntu.com to maas.io

matthewhelmke:
Note: there is one instance in /src/en/config-maas.md that was not changed as it is part of the URL to the docs, which are not migrated to maas.io. Quoting:

> For further information on using Juju with MAAS, see the
> [MAAS and Juju Quick Start guide](http://maas.ubuntu.com/docs/juju-quick-start.html).


pmatulis:
LGTM

PR #942, opened 2016-03-29
Another LXD refresh

pmatulis:


pmatulis:
I have suppressed the generation of the HTML table and did some minor rework at the bottom. The HTML source should list it as a comment. It took a long while to make the table and I am not ready to delete it at this time.

PR #951, opened 2016-04-01
Adding devel ppa for docs.

mbruzek:
A partner hit this problem today, the jujucharms.com/docs/devel/getting-started page advised them to add the "stable" ppa.  I want the devel docs to have the "devel" ppa, which is what was needed in this case.

Looking at the file structure I don't know how we have 2 getting-started pages or how to edit the other one.

Please review, and advise me on what else needs to be done.

### DIFF
--- a/src/en/getting-started.md
+++ b/src/en/getting-started.md
@@ -4,16 +4,16 @@ TODO: Testing section needs rewrite
       Headings need looking at!
       Links need to be reverified when supporting pages are written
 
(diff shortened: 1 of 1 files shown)


mbruzek:
I only changed 2 lines, but I see all sorts of line feed changes. Something with my editor (atom)? If someone could tell me how to change that I would be very grateful.

evilnick:
changes already made

PR #955, opened 2016-04-03
Update charms.md

evilnick:


evilnick:
actually @pmatulis, there is still stuff to add to this, but I need this page to exist, so I am going to add it now and we can finish it later

PR #957, opened 2016-04-05
Purge duplicate and unneeded pages

evilnick:


evilnick:
we don't need these anymore

pmatulis:
OK but I think KVM support will return. Otherwise LGTM.

evilnick:
oh yes, it probably will, but we don't need config pages like this now, so it will return in some different form

PR #960, opened 2016-04-06
Use stronger language for recommending layers.

### LINKED ISSUE #810: [2.0 docs] Why build from layers too much focused on existing users
> "When creating a charm, you always have the option of doing it the traditional way by creating each hook, implementing each side of the interface you need for each relation your charm requires or provides, manage the dependencies, such as charm-helpers, that your charm uses, et cetera."

This text is selling Charm layers to people who are already familiar with the "old" way of Charming. I don't think this should be the first sentence on that page. While the documentation should include statements like this, I think it might be confusing for new users.

castrojo:
Fixes #810

### DIFF
--- a/src/en/authors-charm-building.md
+++ b/src/en/authors-charm-building.md
@@ -3,25 +3,21 @@ Title: Building a Juju charm
 
 # Building is easier than writing
 
-When creating a charm, you always have the option of doing it by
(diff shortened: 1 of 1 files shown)


evilnick:
I will ignore your silly double-spaces after '.' (but only because Markdown does as well)

Thx :cake:

PR #963, opened 2016-04-06
Remove kvm

### LINKED ISSUE #946: The juju devel docs should remove all links to the local KVM provider
Against my numerous objections Juju 2.0 no longer supports the local KVM provider, and the docs in devel need to be removed as Juju users will no longer be able to bootstrap a KVM provider.

This is actually the provider I *was* using on a daily basis and it makes me extremely sad, so I probably can not work on resolving this issue without incredible bias and bitter remorse.

castrojo:
Fix #946 

See ya in juju 2.1!


evilnick:
ah thanks for that. We are going to have another general link check after the carnage has finished.

:cookie:

PR #965, opened 2016-04-06
Make `make serve` work on OS X

AdamIsrael:
Fix `make serve` on platforms where /usr/bin/python2 doesn't exist (OS X and potentially Xenial, where python2 isn't installed by default)


evilnick:
:cake:

Issue #967, opened 2016-04-07
tools-charm-tools needs some language tweaks


evilnick:
@castrojo has done some excellent work bringing this page up to date, but it could do with a once-over for language/spelling.

evilnick:
@matthewhelmke I thought you might be able to load up your English(British) dictionary and have a look over this.

matthewhelmke:
#968 has been merged. Closing.

evilnick:
@matthewhelmke I think if you put "fixes #967" in the PR it automatically closes

matthewhelmke:
I didn't know that! Thanks.

PR #968, opened 2016-04-07
edited to use EN_GB and to fix grammar and typos

matthewhelmke:
In response to #967


evilnick:
looks good, thanks!
:cookie:

PR #987, opened 2016-04-12
Model config

evilnick:
Fixes #987


src/en/models-config.md
```diff
@@ -120,37 +167,11 @@ enable-os-upgrade: false
...
-```
-
-It is not desirable to set this feature on hardware which is not NUMA
-compatible.
-
-
 ## Juju lifecycle and harvesting
 
-Juju keeps state on the running environment from which it builds a model.
+Juju keeps state on the running model from which it builds a model.
```
pmatulis: model model

src/en/models-config.md
```diff
@@ -1,72 +1,135 @@
...
+juju unset-model-config apt-mirror
 ```
-to restore the default behaviour in a running environment.
+to restore the default behaviour in a running model.
 
 
 ## Versions and Streams
 
-The ```agent-stream``` option selects the versions of Juju which an environment
+The ```agent-stream``` option selects the versions of Juju which an model
```
pmatulis: a model

src/en/models-config.md
```diff
@@ -1,72 +1,135 @@
...
 
 ```bash
-juju set-env noproxy=jujucharms.com
+juju set-model-config noproxy=jujucharms.com
+```
+It is possible to specify a list of space-delimited key-value pairs to set more
+than one configuration at a time:
+  
+```bash
+juju set-model-config test-mode=true enable-os-upgrade=false
```
pmatulis: I see this:

$ juju get-model-config | grep refresh
$ juju set-model-config enable-os-refresh-update=false
WARNING key "enable-os-refresh-update" is not defined in the current model configuration: possible misspelling
$ juju get-model-config | grep refresh
enable-os-refresh-update: false

Is this what we want?

src/en/models-config.md
```diff
@@ -1,72 +1,135 @@
...
 or running multiple instances of Juju, where it would be useful to control
-further aspects of the Juju environment. Many options can be changed by adding
-extra key:value pairs to the environments.yaml file which Juju uses to initiate
-environments. Options which are specific to certain cloud providers are
-detailed in the individual configuration pages, but there are additional
-settings which can be applied to any environment. Some of these may be useful
-for particular use cases as outlined in the sections below, or you can
-[skip to the table](#alphabetical-list-of-general-configuration-values) of all
-the values.
+further aspects of the models you create.
```
pmatulis: may want to state explicitly that any machines created in the configured model will inherit these settings

pmatulis:
Review complete

src/en/models-config.md
evilnick: Not sure what you mean. If you mean, do we want a scary error message every time we add some key that isn't there by default - see https://bugs.launchpad.net/juju-core/+bug/1413653

src/en/models-config.md
evilnick: the machines are created 'with these' settings, but they don't really inherit them. Anyhow, I shall reword

src/en/models-config.md
pmatulis: yeah, the bug covers it. warning looks like i did something wrong and the action did not take place

PR #990, opened 2016-04-13
remove experimental section

evilnick:
This removes 'Experimental' from the navigation (for now)

remove outdated user docs
move centos support under Juju
move spaces under Juju (for now - this will probably need its own subsection)


pmatulis:
LGTM :candy:

evilnick:
because I am thinking networks is going to require more than one page and will become a section

PR #997, opened 2016-04-15
Adjusted to correct Juju package name; removed excessive h1 headers

pmatulis:


evilnick:
h2s look rubbish, but we will give it a go for a day or 2

PR #1010, opened 2016-04-21
added maas page

evilnick:


evilnick:
@pmatulis you can argue about it later

Issue #1019, opened 2016-04-22
2.0 'create-model'=>'add-model'


evilnick:
The juju create-model command for 2.0 is now 'add-model' instead. All existing references and examples need to be updated, plus any contextual text which talks about 'creating' models.

PR #1025, opened 2016-04-25
Fixed missed 'create' reference and link.

degville:
Update to #1019 to patch missed link.


mbruzek:
I checked this out and there is a .../devel/models-adding page.

https://jujucharms.com/docs/devel/models-adding

+1 from me!

PR #1031, opened 2016-04-26
Release 2.0-beta6 is the current devel release.

sinzui:


warsaw:
It seems like beta4 is still the current version in Ubuntu though:

```
% rmadison juju-2.0
 juju-2.0 | 2.0~beta4-0ubuntu2 | xenial  | amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 juju-2.0 | 2.0~beta4-0ubuntu2 | yakkety | amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
```

evilnick:
I think it is probably likely to remain so until 2.0 is released. 

If you want to track the dev versions you should add the PPA
https://jujucharms.com/docs/devel/reference-releases#development

warsaw:
Got it, thanks.

PR #1033, opened 2016-04-26
google/joyent help

evilnick:
finishes cloud specific help, apart from plumbing

### DIFF
--- a/src/en/help-google.md
+++ b/src/en/help-google.md
@@ -1,0 +1,114 @@
+Title: Help with Google Compute Engine clouds
+
+# Using the Google Compute Engine public cloud
+
+Juju already has knowledge of the GCE cloud, so unlike previous versions there
+is no need to provide a specific configuration for it, it 'just works'. GCE
+will appear in the list of known clouds when you issue the command:
+  
+```bash
+juju list-clouds
+```
+And you can see more specific information (e.g. the supported regions) by 
+running:
+  
+```bash
+juju show-cloud google
+```
+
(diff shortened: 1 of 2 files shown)


src/en/help-google.md
```diff
@@ -0,0 +1,114 @@
...
+juju list-clouds
+```
+And you can see more specific information (e.g. the supported regions) by 
+running:
+  
+```bash
+juju show-cloud google
+```
+
+If at any point you believe Juju's information is out of date (e.g. Amazon just
```
pmatulis: we might want to change this blurb on all our sheets. how can a user determine if something is out of date? might be better to just say "run this to make sure it is up to date".

src/en/help-joyent.md
```diff
@@ -0,0 +1,59 @@
+Title: Help with Joyent clouds
```
pmatulis: did you actually test this cloud? i didn't think it worked well

pmatulis:
Review complete.

src/en/help-joyent.md
evilnick: it doesn't work well, but it does work

src/en/help-google.md
evilnick: I'm not that bothered about it. Really, the *only* thing that is ever likely to happen is that someone adds a new region. Juju isn't the mechanism people will use to know that anyhow.  So the only time anyone actually *needs* to update their cloud is when they _know_ a new region has been added and they want to use it.
I originally thought that it would be a good idea for people to run this regularly, but really, I don't think it is.

PR #1036, opened 2016-04-27
updated gui references 

degville:
partially fixes #1034 

updates references and inaccuracies to the older version of gui, also taking into account its auto-inclusion in every new controller. Removed the embedded video, as the gui is out of date.

needs new images.

Issue #1034, opened 2016-04-26
Juju 2.0 should have a GUI page


evilnick:
We will at some point want to include some GUI alternatives to existing pages.

Initially however we need to completely overhaul this page:
https://jujucharms.com/docs/devel/juju-gui-management

The gui is now deployed automatically in a controller for a start, and we need new instructions on logging in. See release notes

PR #1039, opened 2016-04-27
updated images for Juju GUI

degville:
update for #1034 and #1036 to include images.


evilnick:
Ok, looks good. 

:cake:

PR #1040, opened 2016-04-27
updated command name, removed info that no longer applies

matthewhelmke:


src/en/juju-ha.md
```diff
@@ -51,16 +51,10 @@ state servers nor remove the failed ones. However, as long as more than half of
 the original number of state servers remain available you can manually recover.
 The process is detailed below.
 
-1. Run `juju ensure-availability`.
-1. Verify that the output of `juju status` shows a value of `has-vote` for the
```
matthewhelmke: I really hope this comes back, but in my testing, there is nothing like this that shows in the output of juju status.

evilnick: the default format of juju status has changed since this was written.

try 

juju status --format=yaml

matthewhelmke: That works, but the attribute name has changed to "controller-member-status"

However, running 'juju enable-ha' a second time does not have the expected result. In fact, nothing changes. So, for now, it seems we can enable, which creates servers if needed, we can check to see that the "controller-member-status" is set, but we can not make Juju no longer consider machines as controller members.

Watch for an edit soon.

matthewhelmke: Urgh. I think that I can't remove because I only have three and a majority must remain to do this recovery. Testing again with more controller members.

matthewhelmke: Nope. I can't remove old controllers at all. I can add to my original pool of 3, up to 5 or 7 total. Perhaps controllers have to be in a 'failed' state before they can be removed. I can not produce this state.

matthewhelmke:
Do not accept this PR.

PR #1046, opened 2016-04-28
getting-started: Added a link to juju 2.0 docs

mattyw:
I think it's worth adding a note to the top of these docs to point folk at Juju 2.0 just in case we get users arriving from Xenial


evilnick:
will you remember to take it off again?

If it makes you happy, I will add it for now

PR #1048, opened 2016-04-28
WIP: Terms documentation

AdamIsrael:
This is the first draft of documentation of the new terms functionality in juju 2.0.


src/en/developer-terms.md
```diff
@@ -0,0 +1,58 @@
...
+be prompted to login when attempting to deploy a term-enabled charm.
+
+# How it works
+
+When a user tries to deploy a term-enabled charm from the charm store, Juju will
+check if the terms have been agreed to. If they have, the charm will deploy as
+usual. If not, the user will be prompted to run the `juju agree` command in
+order to deploy.
+
+You can view which charm terms you've accepted by installing
```
alesstimec: juju list-agreements show user's agreements

src/en/developer-terms.md
```diff
@@ -0,0 +1,58 @@
...
+[charm-tools](./tools-charm-tools.html) and running `charm terms`.
```
alesstimec: charm terms lists terms owned by the user.. or so the doc says..

alesstimec: "charm show $charmURL terms" will show terms required by a charm

alesstimec:
A few comments from me..

src/en/developer-terms.md
```diff
@@ -0,0 +1,58 @@
+Title: Writing charms using terms
+
+# Writing charms that use terms
+
+Many applications require the acceptance of Terms in order to be installed.
```
mattyw: s/Terms/terms

src/en/developer-terms.md
mattyw: Also you can list any terms you have already agreed to with juju list-agreements

src/en/developer-terms.md
```diff
@@ -0,0 +1,58 @@
...
+
+You can view which charm terms you've accepted by installing
+[charm-tools](./tools-charm-tools.html) and running `charm terms`.
+
+# Managing terms
+
+## Creating terms
+
+At the moment, terms can only be created by charm store administrators.
+Charm developers should send an email to juju@lists.ubuntu.com with:
```
mattyw: I'd recommend giving them a subject to use, that way we can add filters to catch these messages, also maybe we should ask them to fill out a form so they don't need to send their terms to the whole list

mattyw: I guess I'm basically saying - let's agree a process then document it

src/en/developer-terms.md
```diff
@@ -0,0 +1,58 @@
...
+In order to make a charm require terms, the `terms` key has been introduced to
+the charm's `metadata.yaml`. For example:
+
+```yaml
+name: terms1
+summary: "That's a dummy charm with terms."
+description: |
+    This is a longer description which
+    potentially contains multiple lines.
+terms: ["lorem-ipsum"]
```
mattyw: I think this needs a revision @cmars ?

src/en/developer-terms.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+The `terms` key can include multiple terms to be required. It can also require
+specific versions of a term, i.e., `lorem-ipsum/2` would reference the second
+version of the `lorem-ipsum` term. Omitting the version will require the latest
```
mattyw: @cmars is this right?

src/en/developer-terms.md
alesstimec: no.. without revision it will require agreement to the latest revision of that term

src/en/developer-terms.md
alesstimec: @mattyw yes, yes, it is :)

src/en/developer-terms.md
AdamIsrael: I completely agree. Since your team (I think?) will be the one responsible for creating terms, how do you want to go about it? A Google spreadsheet/form seems handy in this case, but I don't know if it can be setup to send notifications when new forms are submitted. 

I still favor the idea of using a Github repo, though, where the issues template can be customized for the information need. Notifications work, issues can be assigned and progress tracked.

cmars: Let's use a Google Form for this purpose: https://goo.gl/forms/7cGy0Oztmd

It collects all the form submissions to a spreadsheet, and I've enabled email notifications for it.

evilnick:
when you are happy with the peer-review on any PR, if you just make a comment @juju/docs we can scan it over and merge.

src/en/developer-terms.md
cmars: @AdamIsrael We've just released the fix for this to production. `cs:~cmars/trusty/canonical-terms-example` has a metadata.yaml requiring agreement to latest terms.

src/en/developer-terms.md
```diff
@@ -0,0 +1,94 @@
...
+        "term": "lorem-ipsum",
+        "revision": 1,
+        "created-on": "2016-04-20T21:01:24.84Z"
+    }
+]
+```
+
+## Upgrading terms
+
+Deployed charms will run under the terms they were installed with. If the terms
```
alesstimec: @cmars i think the wording here is a bit awkward.. charm is not running under terms.. term agreement record is kept in the terms service.. so should the updates charm require newer version of terms or additional term agreement then juju upgrade-charm <service> will ask for additional agreements.. but my english is not good enough to put this in proper sentences.

cmars: Let me see what I can come up with... how about:
```
Services run under an agreement to the terms applicable to the charm when it was deployed. When upgrading a service with `juju upgrade-charm`, the user may be prompted to agree to newer revisions of terms already agreed to, or to new terms that have been added to the upgrade charm revision.
```

cmars:
I would like to see some improvements from @AdamIsrael per comments above but I think this is a great start and from a technical perspective is good to land as-is.

LGTM.

evilnick:
ok, cool. words look fine. Thanks :cake:

AdamIsrael:
Thanks for the feedback, @cmars, @mattyw, and @alesstimec. I'll iterate on the docs on soon as time permits.

PR #1050, opened 2016-04-29
Clean up

pmatulis:
I could have sworn we blew juju-multiuser-environments.md out the air-lock already.

The charms-environments.md file gets moved under Models as 'Managing'. It obviously points into the void right now.


evilnick:
:hamster:

PR #1055, opened 2016-04-29
Fix wrong redirection in jujucharms.com as the scripts to transform to jinja2 replace some links.

fabricematrat:
…


evilnick:
:beer:

PR #1058, opened 2016-04-29
Commands b6

evilnick:
Gaze upon that which I have wrought...


mbruzek:
So much destruction!

evilnick:
I'm landing this. there are things which can be tweaked in the generation, and things which we will need to look at in juju itself, but this is not the final version anyhow - it is easier to see what needs to be fixed when we have the site rendering

PR #1064, opened 2016-05-03
`charm create` instead of `juju charm create`

julenl:
It seems that the command `juju` (1.25.5-wily-amd64) does not have the option `charm`. However, the same action can be achieved with the `charm create` or `charm-create` commands.


chuckbutler:
Juju should be passing this along to the charm command, but that may have changed recently and I'm unaware. Thanks for the contribution @julenl

PR #1065, opened 2016-05-03
Added .conf extension to the vanilla config file

julenl:
Since Apache version 2.4 configuration files should have the extension `.conf` by default.
Source: http://askubuntu.com/a/451223/287964


evilnick:
thanks :cake:

PR #1066, opened 2016-05-03
Issue 1051 models managing

### LINKED ISSUE #1051: Docs needed: Models > Managing
This issue spawned from #1008 .
This new file needs to be populated: models-managing.md .
Be sure to include `juju switch` usage.

pmatulis:
Fixes #1051

### DIFF
--- a/src/en/models.md
+++ b/src/en/models.md
@@ -1,30 +1,165 @@
-Title: Juju Models
-TODO: Review again soon (created: April 2016)
+Title: Juju models
+TODO: Previous warning (add-model): "For 'ec2' and 'openstack' cloud types, the
+      access and secret keys need to be provided." I tested ec2 and did not
+      need to do this. OpenStack?
+      Due to bug https://goo.gl/yEGzP2 examples cannot use syntax highlighting.
 
-# Models
 
-A Juju *model* is an environment associated with a cloud. When
-a *controller* (see [Controllers](./controllers.html)) is 
-created by bootstrapping a cloud, two models are provisioned along 
-with it. 
+# Juju models
 
-These initial models are named '**admin**' and '**default**'. The 'admin'
-model is not intended for general workloads but solely for monitoring
-and management purposes.
+A Juju *model* is an environment associated with a *controller* (see
+[Controllers](./controllers.html)). When a controller is created two models are
+provisioned along with it. These initial models are named '**admin**' and
+'**default**'. The 'admin' model is not intended for general workloads but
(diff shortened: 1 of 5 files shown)


src/navigation.tpl
```diff
@@ -19,10 +19,9 @@
         </li>
         <li class="section"><a class="header" href="models.html">Models</a>
             <ul class="sub">
-                <li><a href="models-adding.html">Adding</a></li>
-                <li><a href="models-managing.html">Managing</a></li>
-                <li><a href="models-configuring.html">Additional configuration</a></li>
+                <li><a href="models-config.html">Configuring</a></li>
                 <li><a href="models-upgrade.html">Upgrading</a></li>
+                <li><a href="models-ha.html">Juju HA</a></li>
```
evilnick: Isn't HA only applicable to the controller?

src/en/models.md
```diff
@@ -1,30 +1,165 @@
...
+   
+   Juju software is upgraded at the model level. This affects the Juju agents
+   running on every machine Juju creates. This upgrade process does not pertain
+   to the Juju software package installed on a client system.
+
+   This is a complex subject. See [Juju upgrades](./models-upgrade.html).
+
+
+
+^# Set up HA for a model
```
evilnick: as mentioned elsewhere, isn't HA applied at the controller level?

src/en/models.md
```diff
@@ -1,30 +1,165 @@
...
+   
+
+^# Grant and revoke user access
+   
+   This topic is treated in
+   [Users and models](./users-models.html#models-and-user-access).
+
+
+
+^# Configure a model
```
evilnick: I would move this further up the list

evilnick:
comments made inline.

Looks good, but beware of foldouts - because of that bug in the way they are processed, i'm not sure your examples will be rendered properly.

Otherwise :+1:, :dolphin:

src/navigation.tpl
pmatulis: Yeah, you're right. It is only HA for the model in an abstract, architectural sense.

src/en/models.md
pmatulis: Yeah.

src/en/models.md
pmatulis: OK

pmatulis:
Re rendering, let's see what happens. I'll follow up accordingly.

Since this organization is acceptable we can go on to do the same for section 'Controllers'.

PR #1073, opened 2016-05-04
First letter in upper case

julenl:
First letter in upper case


pmatulis:
Possible to add?:

s/url/URL
s/mysql/MySQL

julenl:
Of course...
Thanks

PR #1074, opened 2016-05-04
clarification on the directory

julenl:
Not sure whether the CWD is meant to be the one where the test scripts are located, or the one from where the test is executed.
In the second case, the sentence could be rewritten as "The tests are executed from the root directory of the charm containing the `tests` directory"


julenl:
Sure! Done.
Thanks

PR #1076, opened 2016-05-05
Charms bundles update

### LINKED ISSUE #1044: charms-bundles page needs a refresh
there are formatting errors, a poor graphic and some out of date concepts

degville:
Ready for review @juju/docs 
fixes #1044 
Updates charms-bundles; words, images, and accuracy.


evilnick:
definitely worth having a good look at the beginning of this. Bundles are now, to the user, just as simple as charms in terms of deployment. So the 'using' section should explain they can be deployed from the store exactly* the same way as charms.

*(nearly, when it works)

degville:
May make better sense to rename this Using and Creating Bundles to reflect the order of the text.

evilnick:
I agree, do so and merge it :+1:

PR #1087, opened 2016-05-09
added JUJU_DATA to env reference

### LINKED ISSUE #742: Undocumented Environment variables
Inside an executing action, this env var is set and contains the uuid of the currently executing action (e.g., d4044c7e-7fd1-4566-8aee-c574d3dedd5d). There is no mention of this var in the docs.

evilnick:
Fixes #742

Actually, there may be more new ones. Will add that to the list of info we need

### DIFF
--- a/src/en/reference-environment-variables.md
+++ b/src/en/reference-environment-variables.md
@@ -15,33 +15,18 @@ Use header tags so we can link to these variables individually.
 These variables are available on the juju client in order to change its default
 behavior.
 
-#### JUJU_ENV
 
-The current environment that juju will use by default. Juju has three ways it
-will  determine current environment, in descending order:
+#### JUJU_DATA
 
-1. Check to see if `JUJU_ENV` is set.
-```
-echo $JUJU_ENV
-local
+This sets the path where Juju will look for its configuration files. You do not need to set this - by default Juju follows XDG guidelines and on Linux systems it will use the path:
+
+```no-highlight
+~/.local/share/juju
 ```
 
-2. Read the `$JUJU_HOME/current-environment` file, which is set by the
-`juju switch` command.
-```
-juju switch amazon
-cat ~/.juju/current-environment
-amazon
-```
+#### JUJU_HOME (Deprecated)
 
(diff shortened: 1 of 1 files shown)


evilnick:
for review @juju/docs

src/en/reference-environment-variables.md
```diff
@@ -15,33 +15,22 @@ Use header tags so we can link to these variables individually.
...
-amazon
+```no-highlight
+~/.local/share/juju
 ```
 
-3. Read `$JUJU_HOME/environments.yaml` and use the default environment.
-```yaml
-default: local
-```
+#### JUJU_HOME (Deprecated)
```
marcoceppi: If this is deprecated and not available in 2.0, why even include it here?

evilnick: As there is no upgrade path for 2.0 and won't be for some time, the reality is that many users will be running both 1.x and 2.x for some time. There are other docs detailing how to use both side by side, but since it is likely that this may be set to support a 1.x install, it doesn't hurt to have it documented here imo

marcoceppi: Given that we have a 1.25 and (soon to be) 2.0 version doc, if users what to collocate both they should use either version of the doc. Bleeding the two just creates a confused experience and sets an experience that I don't think we want to set.

evilnick: THIS documentation, i.e. the 2.0 docs will be referencing this very variable, in this version of the docs.

pmatulis: I too appreciate a clean division of doc versions but we should be able to make exceptions. Due to the non-ready software we need to foresee what sort of information a user may want to look up and the JUJU_HOME variable is such a beast. The minor inconvenience in including 3 lines in the documentation far outweighs the frustration a user will likely feel by not including them. Since this is the only variable in this category, also clearly marked 'deprecated', it is indeed an exception.

I favour the inclusion of this variable in the 2.0 docs until the upgrade mechanism is built.

PR #1092, opened 2016-05-11
typo: added a missing 'and'.

degville:
@juju/docs

For these very small fixes, is it more practical to just merge them ourselves?


evilnick:
yes. if it is an obvious mistake, that is cool.

PR #1100, opened 2016-05-13
Update Juju2 metadata docs.

wwwtyro:
Adds documentation for `resources`, `payloads`, `extra-bindings`, `series`, `terms`, and `min-juju-version`. #1079

Issue #1079, opened 2016-05-05
Charm Metadata Docs Out of Date


ericsnowcurrently:
The page on charm metadata [1] is missing mention of some sections, including resources (2.0+) and payloads (1.25+).  There may be other out-of-date info.  For the authoritative source:

1.25:  https://github.com/juju/charm/blob/v5/meta.go#L669
master (2.0):  https://github.com/juju/charm/blob/v6-unstable/meta.go#L803

----
[1] 
https://jujucharms.com/docs/1.25/authors-charm-metadata
https://jujucharms.com/docs/devel/authors-charm-metadata

presumably, respectively:
https://github.com/juju/docs/blob/1.25/src/en/authors-charm-metadata.md
https://github.com/juju/docs/blob/master/src/en/authors-charm-metadata.md

chuckbutler:
@evilnick - the attached branches/PR's LGTM. once merged this should be good to close :cake:

pmatulis:
Fixed via #1100

PR #1102, opened 2016-05-16
update juju 1.25 metadata docs

Cynerva:
Add docs for `storage`, `payload`, `series`.

Related to #1100 which is for Juju2 docs - I pulled these definitions straight out of there.

Issue: #1079

### DIFF
--- a/src/en/authors-charm-metadata.md
+++ b/src/en/authors-charm-metadata.md
@@ -76,7 +76,21 @@ keep the Charm Store organised.
 
 ![Juju Charm Store metadata Listing](./media/authors-metadata-display.png)
 
-Finally, a metadata file defines the charm's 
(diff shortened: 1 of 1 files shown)


pmatulis:
If this is the same addition as in #1100 then we should request a backport in the associated bug (#1079).

pmatulis:
@Cynerva You pulled those things from 2.0. Are we sure those items apply to 1.25?

Cynerva:
@pmatulis Looks correct to me based on the schema here, which I believe is 1.25: https://github.com/juju/charm/blob/v5/meta.go#L669

chuckbutler:
Yep, storage, payload, and series are all valid for 1.25.  Do note, that not all providers support storage, so its tricky to validate. At time of writing, I believe only OpenStack, and Amazon currently have storage support in 1.25.

PR #1114, opened 2016-05-22
Fixed "juju add-crdential" typo in GCE help

mjs:
Also tidied up the text describing the handling of the credentials file.


mjs:
Does this also need to be backported into the 2.0 docs branch?

pmatulis:
Thanks. And, no, the devel release (current) *is* the 2.0 release.

PR #1125, opened 2016-06-02
Move definition of authentication types outside regions

theosotr:
Juju ignores authentication types when there are defined inside regions.


evilnick:
Thanks! :Cake:

PR #1128, opened 2016-06-02
Getting started again

### LINKED ISSUE #1109: https://jujucharms.com/docs/devel/getting-started has some extra uneeded steps
Moved from https://github.com/CanonicalLtd/jujucharms.com/issues/259

evilnick:
Fixes #1109 

@juju/docs Check that the foldout works for you, and the instructions are sane. I have run through this 4 times on clean xenial systems and had no problem with groups etc.


matthewhelmke:
Instructions are clear and work for me in my tests.
:cake:

src/en/getting-started.md
```diff
@@ -24,65 +21,116 @@ Both the above are provided with Ubuntu 16.04LTS.
...
-sudo zpool create lxd /var/lib/zfs/lxd.img
+groups
 ```
+This should indicate the user is a member of the lxd group, amongst others:
 
-As this is sparse storage, it won't actually take up disk space until it is 
-actually being used. You can check the file has been added to a ZFS pool with 
-the following command:
+```bash 
+lxd adm cdrom sudo dip plugdev lpadmin sambashare ubuntu
```
pmatulis: Listing groups can be a cause of confusion/frustration if the reader's user looks different.

s/bash/no-highlight if we keep this kind of output.

src/en/getting-started.md
```diff
@@ -24,65 +21,116 @@ Both the above are provided with Ubuntu 16.04LTS.
 
 Run the following commands to install the required software:
 
-```bash
-sudo add-apt-repository ppa:juju/devel
+```no-highlight
```
pmatulis: s/no-highlight/bash

src/en/getting-started.md
```diff
@@ -24,65 +21,116 @@ Both the above are provided with Ubuntu 16.04LTS.
...
 sudo apt update
-sudo apt install juju zfsutils-linux lxd
-newgrp -
+sudo apt install juju-2.0 zfsutils-linux
```
pmatulis: Although 'juju-2.0' works, I'm pretty sure we should be just using 'juju'.

evilnick: i will change it again when we are actually done. All the different package names are getting confusing

src/en/getting-started.md
evilnick: it was highlighting random things again so i turned it off

src/en/getting-started.md
evilnick: I will clarify in the text

PR #1137, opened 2016-06-06
rework example to use new mediawiki bundle

evilnick:
also service->application


evilnick:
@juju/docs some renaming and using the newer mediawiki bundle as an example

PR #1149, opened 2016-06-07
Removed reference to deprecated local url scheme.

petevg:
Juju 2.0 no longer supports invocations like "juju deploy
local:trusty/somecharm". Changed the instructions to instead reference
"juju deploy $JUJU_REPOSITORY/trusty/somecharm --series trusty".


evilnick:
:cake:

Issue #1168, opened 2016-06-10
Add default region info to clouds.md


matthewhelmke:
To see what clouds are available, use:

    juju list-clouds

      CLOUD              TYPE        REGIONS
      aws                ec2         us-east-1, us-west-1, us-west-2, ...
      aws-china          ec2         cn-north-1
      aws-gov            ec2         us-gov-west-1
      azure              azure       japanwest, centralindia, eastus2, ...
      azure-china        azure       chinaeast, chinanorth
      cloudsigma         cloudsigma  mia, sjc, wdc, zrh, hnl
      google             gce         us-east1, us-central1, ...
      joyent             joyent      us-east-1, us-east-2, us-east-3, …
      lxd                lxd
      maas               maas
      manual             manual
      rackspace          rackspace   lon, syd, hkg, dfw, ord, iad

The default region is shown first in the list-clouds output so an
aws bootstrap that doesn't specify a region will
create a controller called "mycontroller" on us-east-1. To specify
a different region, use:

    juju bootstrap mycontroller aws/us-west-2

To set the default region for a cloud, use:

    juju set-default-region aws us-west-1

matthewhelmke:
Fixed via #1178

PR #1178, opened 2016-06-13
added info about default region and finding cloud details

### LINKED ISSUE #1167: Add show-clouds to clouds.md
After you use `juju list-clouds`, if you want more detail about a particular cloud, use:

    juju show-cloud azure

Seems like useful info to include in the main cloud page.

matthewhelmke:
@juju/docs
Fixes #1167 and #1168

### DIFF
--- a/src/en/clouds.md
+++ b/src/en/clouds.md
@@ -41,10 +41,7 @@ This lists the cloud name (which you will use to specify the cloud you want to
 use), its type (the API used to control it) and the known regions for the cloud.
 The default region for a cloud is shown first in the list-clouds output, so
 in the above, `us-east-1` is the default region for an aws cloud. To specify
-a different region during controller creation, use:
-```bash
-juju bootstrap mycontroller aws/us-west-2
-```
+a different region, see [Creating a controller](./controllers-creating.html).
 
 Set the default region for a cloud with:
 ```bash
--- a/src/en/clouds.md
+++ b/src/en/clouds.md
@@ -39,6 +39,22 @@ rackspace    rackspace   DFW, ORD, IAD, LON, SYD, HKG
 
 This lists the cloud name (which you will use to specify the cloud you want to 
 use), its type (the API used to control it) and the known regions for the cloud.
+The default region for a cloud is shown first in the list-clouds output, so
+in the above, `us-east-1` is the default region for an aws cloud. To specify
+a different region during controller creation, use:
+```bash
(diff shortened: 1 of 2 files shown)


src/en/clouds.md
```diff
@@ -39,6 +39,22 @@ rackspace    rackspace   DFW, ORD, IAD, LON, SYD, HKG
 
 This lists the cloud name (which you will use to specify the cloud you want to 
 use), its type (the API used to control it) and the known regions for the cloud.
+The default region for a cloud is shown first in the list-clouds output, so
+in the above, `us-east-1` is the default region for an aws cloud. To specify
+a different region during controller creation, use:
+```bash
+juju bootstrap mycontroller aws/us-west-2
```
pmatulis: We can mention region defaults (I didn't know this) but I think the actual example command should be in [controllers-creating](https://jujucharms.com/docs/devel/controllers-creating) and possibly provide a link to that page.

evilnick:
Yes, the region stuff should go with the bootstrap command. Just put it on that page instead and link it from here "(To specify a different region, see xxxx)"

matthewhelmke:
@pmatulis @evilnick 

That makes sense. I have moved the info.

src/en/clouds.md
```diff
@@ -39,6 +39,19 @@ rackspace    rackspace   DFW, ORD, IAD, LON, SYD, HKG
 
 This lists the cloud name (which you will use to specify the cloud you want to 
 use), its type (the API used to control it) and the known regions for the cloud.
+The default region for a cloud is shown first in the list-clouds output, so
+in the above, `us-east-1` is the default region for an aws cloud. To specify
+a different region, see [Creating a controller](./controllers-creating.html).
```
pmatulis: Now it looks confusing. I would put this link under the global `set-default-region` command below and say like: "To set a non-default region during the creation of a controller see *link*..."

matthewhelmke: I hear what you are saying. I don't feel strongly either way, although my original thought in placing it where I did is that it creates a nice segue from "here is how to find your default region" to "set the default region to something else". Putting "use a region other than the default" either between those two or after makes sense to me. Hmm. @evilnick , do you have a preference?

matthewhelmke: Ahh, I see he is out today. In that case, I'll merge as-is and we can revisit the location for the info again later.

PR #1192, opened 2016-06-18
Changed 'controller' model into 'admin' model

mever:


pmatulis:
Rejected. The current text is correct. Every controller comes with two models, named 'controller' and 'default'.

pmatulis:
Recall that the master branch follows Juju 2.0.

mever:
On various other places in the documentation it says differently:
https://github.com/juju/docs/blob/master/src/en/models.md#juju-models (third sentence)

Also, when executing "juju list-models" on my machine (juju 2.0-beta7) it says:

> NAME      OWNER        STATUS     LAST CONNECTION
> admin     admin@local  available  never connected
> default*  admin@local  available  44 minutes ago

marcoceppi:
The latest beta, beta9 renames this to controller from admin.

Marco

mever:
Ah, my bad. Thanks for clearing that up for me.

PR #1197, opened 2016-06-21
housecleaning updates due to networking changes

matthewhelmke:
models-config.md and refrence-release-notes.md - drop references to prefer-ipv6 as it is no longer present in Juju 2.0

models-config.md - proxy-ssh is now false by default

@juju/docs for review


evilnick:
thanks

PR #1203, opened 2016-06-23
More renaming of service->application

evilnick:
@matthewhelmke there are still some in reference sections and of course, most of the 'developer' docs


matthewhelmke:
LGTM
:cake:

PR #1204, opened 2016-06-23
the address-allocation feature flag was removed...

matthewhelmke:
...so the references to it were also removed. That's the only change here.

@juju/docs for confirmation before merge

### DIFF
--- a/src/en/reference-release-notes.md
+++ b/src/en/reference-release-notes.md
@@ -1703,6 +1703,18 @@ The versions covered here are:
   leader-settings-changed as soon as possible, delaying only doing so only
   to run the install hook; complete any queued or in-flight operation; or
   resolve a hook or upgrade error.
+
+ ### Experimental: Addressable LXC Containers and KVM Instances on AWS and MAAS
+ 
+ The Juju AWS and MAAS providers now support starting LXC containers. The
+ MAAS providers also supports networking on KVM. Containers and Virtual
+ Machines will be given statically allocated private IP addresses from
+ the same subnet as their host machine. For example on MAAS you can now:
+ 
+```bash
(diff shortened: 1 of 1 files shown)


src/en/reference-release-notes.md
```diff
@@ -1713,25 +1704,17 @@ The versions covered here are:
...
-  MAAS providers also supports networking on KVM. Containers and Virtual
-  Machines will be given statically allocated private IP addresses from
-  the same subnet as their host machine. For example on MAAS you can now:
-
-  ```bash
-  juju deploy wordpress --to lxc:0
-  juju add-unit mysql --to kvm:1
-  ```
-
-  This experimental feature can be enabled when the environment is
```
matthewhelmke: I only intended to delete this part, not the above.

src/en/reference-release-notes.md
```diff
@@ -1713,25 +1704,17 @@ The versions covered here are:
...
-  bootstrapped like so:
-
-  ```no-highlight
-  JUJU_DEV_FEATURE_FLAGS="address-allocation" juju bootstrap
-  ```
+ ### Experimental: Addressable LXC Containers and KVM Instances on AWS and MAAS
+ 
+ The Juju AWS and MAAS providers now support starting LXC containers. The
+ MAAS providers also supports networking on KVM. Containers and Virtual
```
pmatulis: provider (singular)

dimitern: This whole section should be gone as well in 2.0 rel notes

matthewhelmke: Thanks, @dimitern , I'll take care of it.

PR #1208, opened 2016-06-27
Clarify MAAS credential configuration

dannf:
The current documentation describes adding MAAS credentials, but it isn't
clear that this is actually the MAAS API key, nor where it can be found.
Clarify this by providing an example 'juju add-credential' session with
some text about where the API key can be found.

Prefix shell commands with "$" to distinguish them from the command output.


pmatulis:
Hi dannf. Our documentation style guide excludes the usage of prompts. Also, 'juju' should be 'Juju' when  it refers to the software (i.e. not the command `juju`). Note that the API key can also be obtained via the CLI:

`sudo maas-region apikey --username=<username>`

dannf:
On Mon, Jun 27, 2016 at 01:29:42PM -0700, Peter Matulis wrote:
> Hi dannf. Our documentation style guide excludes the usage of
> prompts.

Hi Peter.

Can you point me to the style guide so I can find out how to represent
command output?

> Also, 'juju' should be 'Juju' when  it refers to the software
> (i.e. not the command `juju`).

OK, will fix.

> Note that the API key can also be obtained via the CLI:
> 
> `sudo maas-region apikey --username=<username>`

OK. Are you implying that you'd like for me to mention this in
addition to/instead of the web interface method, or is that just an
fyi?

  -dann

pmatulis:
Hi dann,

> Can you point me to the style guide so I can find out how to represent
command output?

At least on the devel part of the site the bottom of the menu (left pane) has a link [Help improve these docs](https://jujucharms.com/docs/devel/contributing).

> OK.Are you implying that you'd like for me to mention this in addition to/instead of the web interface method, or is that just an fyi?

Both would be useful.

Thanks for helping.

Issue #808, opened 2016-01-21
[2.0 docs] More explanation of `export JUJU_REPOSITORY` statements


galgalesh:
First of all: The getting started page looks good. It's very focused, and provides a lot of links to other examples. The creating your charm repo instructions are a great addition Creating the local repo was my first big hurdle when I first started using Juju about a year ago. I'ts nice to see this problem tackled. 

However, I think the environment variables need a little bit more explanation. I think you overestimate the skills of a lot of newcomers. Stuff that I think should be explained here:

 - JUJU_REPOSITORY will only be available in the current console. When you open a new console, the environment variables will 'disappear'. Other existing consoles will not have the var.
 - Where to put the environment variables so that every new console will have them.

Problems like this are hard to figure out when you start using Linux as a dev environment.

mbruzek:
For what it is worth we did have the section about [preparing your workspace](https://jujucharms.com/docs/devel/developer-layer-example#prepare-your-workspace) in the developer documentation. 

But I have added notes to the other sections that use the environment variables.

PR #1214, opened 2016-06-28
Env variables

mbruzek:
Addresses #808

### DIFF
--- a/src/en/authors-charm-building.md
+++ b/src/en/authors-charm-building.md
@@ -32,6 +32,11 @@ mkdir -p $JUJU_REPOSITORY/{layers,trusty}
 cd $JUJU_REPOSITORY/layers
 ```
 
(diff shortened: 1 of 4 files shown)


evilnick:
I know it may seem a bit nitpicky, but can we use 

      !!! Note: xxx xx xx x xxx

notation for notes? We are trying to standardise on the more common syntax to integration with other docs.

Apart from that, :cake: :cocktail: :tada: :candy: :fireworks:

mbruzek:
I could not remember if that still worked and since my editor did not render !!! I didn't want to introduce something that I could not verify. It is done as you requested.

evilnick:
you can always `make && make serve` to check locally if it will build. Soon we hope to have a better staging area, just to reassure you!

:candy:

Issue #1200, opened 2016-06-23
network-get hook environment tool undocumented


stub42:
Juju 2.0 seems to have just grown a network-get hook environment tool, and it is not yet listed on https://jujucharms.com/docs/devel/reference-hook-tools

In particular, I'd like to know if this is blessed over unit-get for things like retrieving the local IP addresses, and if we actually get an IP address or maybe something else like a DNS name. IPv4 and IPv6 may be mentionable too.

dimitern:
It's currently being documented, with examples, etc. for the 2.0 release. Until done, `juju help-tool network-get` is a way to get help.

PR #1216, opened 2016-06-28
Adds/updates the payload commands to the reference guide

### LINKED ISSUE #927: No mention of storage or payloads on metadata.yaml page
Storage and payloads are both available in 1.25, but neither are mentioned here: https://jujucharms.com/docs/1.25/authors-charm-metadata

There is a page devoted to storage here: https://jujucharms.com/docs/1.25/storage. Perhaps we could link to that.

There is some info about payloads in the 1.25.0 release notes here: https://jujucharms.com/docs/1.25/reference-release-notes

chuckbutler:
fixes #927

### DIFF
--- a/src/en/reference-hook-tools.md
+++ b/src/en/reference-hook-tools.md
@@ -296,6 +296,28 @@ open-port 80/tcp
 open-port 1234/udp
 ```
 
+## payload-status-set
+
+`payload-status-set` is used to update the current status of a registered payload.
+The `class` and `id` provided must match a payload that
+has been previously registered with juju using [payload-register](#payload-register). The status must be one of the following:
+
+- starting
+- started
+- stopping
+- stopped
+
+python:
+```python
+from charmhelpers.core.hookenv import payload_status_set
+
+payload_status_set('monitor', '0fcgaba', 'stopping')
+```
+
+bash:
+```shell
+payload-status-set monitor abcd13asa32c starting
+```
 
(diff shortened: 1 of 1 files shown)


src/en/reference-hook-tools.md
```diff
@@ -296,6 +296,28 @@ open-port 80/tcp
 open-port 1234/udp
 ```
 
+## payload-status-set
+
+`payload-status-set` is used to update the current status of a registered payload.
+The `class` and `id` provided must match a payload that
+has been previously registered with juju using [payload-register](#payload-register). The <status> must be one of the following:
```
mbruzek: You can not use less-than and greater-than symbols in markdown like that. When rendered I do not see "&lt;status&gt;"

evilnick: I would be tempted to leave them out - we aren't referring to an example or anything that we have identified as '&lt;status>'

evilnick: otherwise you need to use `&lt;`

mbruzek: `&lt;` and `&gt;` is how I got the status to render in the message.

chuckbutler:
Pushed a fix on that last commit. LMK if there's anything else in here that needs doing.

evilnick:
:cake:

evilnick:
P.S., do you have examples for network-get too? #1200

PR #1217, opened 2016-06-28
Making note of the automatic retry in the docs.

### LINKED ISSUE #831: [2.0 docs] Hook errors advice needs to be updated
Since we now retry hooks on error automatically [in Juju 2.0], this has implications for the advice we give charm developers on [hook errors](https://jujucharms.com/docs/devel/authors-hook-errors).

While nothing really changes in terms of the importance of avoiding hook errors due to charm defects (badly-written charms won't heal themselves!), I think it does have potential impact on how hooks should deal with unexpected externalities such as a temporary network connectivity issue during package installs, etc.

The current docs advise the charm author to "wait a moment and retry a couple of times" in the event of a transient error. I think this advice now adds needless complexity -- it would be better for the hook to fail and retry in such cases where basic assumptions on the state of the system become invalid.

Thoughts? Where do we draw the line on this?

(Thanks @jrwren for bringing this to my attention!)

mbruzek:
Fixes #831 

These are the best words I can find for this awkward topic. Juju automatically retries after version 2.0 but you still have to manually retry in earlier versions. If someone has better words I am interested in the assistance.

Also it should be noted the `authors-*` docs are older and replaced with the `developer-*` docs, but this fix includes changes to both.

### DIFF
--- a/src/en/authors-hook-debug.md
+++ b/src/en/authors-hook-debug.md
@@ -1,4 +1,4 @@
-Title: Debugging Juju charm hooks   
+Title: Debugging Juju charm hooks
 
 # Debugging hooks
 
@@ -27,7 +27,7 @@ check the `db-relation-joined` and `db-relation-broken` hooks:
 juju debug-hooks mysql/0 db-relation-joined db-relation-broken
 ```
 
-**Note:** It is possible and often desirable to run debug-hooks on more than
+!!! Note: It is possible and often desirable to run debug-hooks on more than
 one unit at a time. You should open a new terminal window for each.
 
 
@@ -71,9 +71,9 @@ queue until you exit your current window. See the  special considerations below.
 The queue for pending hooks will restart once you exit the window with an `exit`
 command.
 
-**Note:** To allow Juju to continue processing events normally, you **must**
-exit the hook execution window with the `exit` command, otherwise all further
-events on that unit will be paused indefinitely.
+!!! Note: To allow Juju to continue processing events normally, you **must**
+exit the hook execution with a zero return code (using the `exit` command),
+otherwise all further events on that unit may be paused indefinitely.
 
 The queue can be halted by exiting with an `exit 1` command, which will flag the
 hook as failed. Juju will revert to its normal behaviour of suspending
@@ -92,8 +92,10 @@ after the unit comes up, making it difficult to start a debug-hooks session in
 time to intercept them. If you're having difficulties, you can temporarily
 return an error code from your `install` hook (e.g. add an `exit 1` at the end
 of it), and start your session only when the unit reports an [error status
-](./authors-hook-errors.html). You should then run `juju resolved --retry` for
-the affected unit, and go back to the debug-hooks session to interact.
+](./authors-hook-errors.html). Juju version 2.0 and up will automatically retry
+the hooks in error. Earlier versions of Juju can be manually retried by issuing
+the command `juju resolved --retry unit-name\#` for the affected unit, and go
+back to the debug-hooks session to interact with the Juju environment. 
 
 
 ## Special considerations
--- a/src/en/authors-hook-debug.md
+++ b/src/en/authors-hook-debug.md
@@ -7,9 +7,6 @@ don't happen quite as you expected. That's why Juju includes two tools to help
 you debug charm hooks: The `juju debug-hooks` command and the `juju debug-log`
 command.
 
-The [dhx debugging plugin](./authors-hook-debug-dhx.html) is also available.
-
-
 ##  The 'debug-hooks' command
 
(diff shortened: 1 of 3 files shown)


mbruzek:
@cmars please take a look and tell me if you have better words.

src/en/authors-hook-debug.md
```diff
@@ -92,8 +92,10 @@ after the unit comes up, making it difficult to start a debug-hooks session in
 time to intercept them. If you're having difficulties, you can temporarily
 return an error code from your `install` hook (e.g. add an `exit 1` at the end
 of it), and start your session only when the unit reports an [error status
-](./authors-hook-errors.html). You should then run `juju resolved --retry` for
-the affected unit, and go back to the debug-hooks session to interact.
+](./authors-hook-errors.html). Juju version 2.0 and up will automatically retry
```
cmars: When I develop charms with Juju 2.0, I usually don't wait around for the hook to retry though. How about something like:

> Hooks in error must be retried manually in Juju releases prior to 2.0, by issuing the command `juju resolved --retry unit-name\#` for the affected unit. Juju version 2.0 and up will automatically retry hooks in error periodically. However, `juju resolved --retry` may still be used to retry the hook immediately. After retrying, go back to the debug-hooks session to interact with the Juju environment.

I'm probably not fit to review `juju debug-hooks` in particular. I never use it; I just use print statements to debug things and use `charm build ...; juju upgrade-charm ...; juju resolved --retry ...` to try changes out iteratively.

src/en/authors-hook-debug.md
```diff
@@ -71,9 +71,9 @@ queue until you exit your current window. See the  special considerations below.
 The queue for pending hooks will restart once you exit the window with an `exit`
 command.
 
-**Note:** To allow Juju to continue processing events normally, you **must**
-exit the hook execution window with the `exit` command, otherwise all further
-events on that unit will be paused indefinitely.
+!!! Note: To allow Juju to continue processing events normally, you **must**
+exit the hook execution with a zero return code (using the `exit` command),
+otherwise all further events on that unit may be paused indefinitely.
```
cmars: s/paused/blocked/ ?

cmars:
Some suggestions.

src/en/authors-hook-debug.md
evilnick: yeah, I think blocked is better

src/en/authors-hook-debug.md
evilnick: I recall there was discussion around some sort of upper limit to automatic retries. Did that ever happen?

src/en/developer-debugging.md
```diff
@@ -74,14 +74,18 @@ check the `db-relation-joined` and `db-relation-broken` hooks:
 juju debug-hooks mysql/0 db-relation-joined db-relation-broken
 ```
 
-Additionally, if your hook is already in an error state, you can tell Juju to
-retry the hook after running `juju debug-hooks`:
+Additionally, if your hook is already in an error state, you can enter in an
```
evilnick: enter into

src/en/troubleshooting-upgrade.md
```diff
@@ -53,5 +53,8 @@ sometimes lead to hook failures. Connect to that unit, see what is wrong, and
 retry with:
 
 ```bash
-juju resolved --retry
+juju resolved --retry unit-name/#
 ```
+
+Where unit-name is replaced with the name of the unit, and # is replaced with
+the unit number you wish to restart.
```
evilnick: could possible do with an example, e.g.

   juju resolved --retry mariadb/1

src/en/authors-hook-debug.md
cmars: Not sure. I'll ask the mailing list.

cmars: @evilnick To answer your question.. there is no upper limit, but there is a backoff delay. See https://lists.ubuntu.com/archives/juju-dev/2016-June/005716.html

cmars: Also note that automatic retry can be disabled: https://lists.ubuntu.com/archives/juju-dev/2016-June/005718.html

evilnick: thanks for that, I will add another issue to make sure this is captured in other relevant places

src/en/authors-hook-debug.md
mbruzek: I am happy to fix that, but want to point out that is the only part of the section that I did not change.

evilnick:
@mbruzek I'm not sure what you did in your last commit but there now seem to be conflicts :(

evilnick:
:fireworks: :tada: :balloon: :candy: :icecream:

PR #1220, opened 2016-06-29
Added a note about not using -broken

### LINKED ISSUE #1032: Don't use -broken in reactive
```
<cory_fu> magicaltrout: Don't use -broken in reactive
<cory_fu> Also, your hook line should be @hook('{requires:dcos}-relation-{joined,changed}')
<cory_fu> Other than that, what's the problem?
<magicaltrout> yeah i saw it ;(
<cory_fu> I was in a meeting, and not really able to pay attention before.  :(
<magicaltrout> if you shouldn't use broken you should probably get evilnick to touch up the docs: https://jujucharms.com/docs/devel/developer-layers-interfaces
<cory_fu> Yeah, we should update that.  It will only actually cause problems if you use set_state in there, but it doesn't work like you'd expect so it's better to avoid it entirely
```

chuckbutler:
Interface layers tend to yield odd behavior when you decorate for the -broken event. This note adds a callout for that behavior.

Fixes #1032

### DIFF
--- a/src/en/developer-layers-interfaces.md
+++ b/src/en/developer-layers-interfaces.md
@@ -155,8 +155,8 @@ class HttpProvides(RelationBase):
         # Signify that the relationship is now available to our principal layer(s)
         self.set_state('{relation_name}.available')
 
-    @hook('{provides:http}-relation-{broken,departed}')
-    def broken(self):
+    @hook('{provides:http}-relation-{departed}')
+    def departed(self):
         # Remove the state that our relationship is now available to our principal layer(s)
         self.remove_state('{relation_name}.available')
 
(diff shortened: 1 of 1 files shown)


src/en/developer-layers-interfaces.md
```diff
@@ -171,6 +171,10 @@ class HttpProvides(RelationBase):
         self.set_remote(**relation_info)
 ```
 
+!!!Note: You may notice we do not decorate for the -broken hook event. This
```
evilnick: Sorry, this won't work - it needs to be "!!! Note: "

evilnick:
otherwise LGTM

chuckbutler:
Ah good catch. Fixed with last push

evilnick:
:candy:

Issue #802, opened 2016-01-19
Outdated language in Developer guide


evilnick:
The charm developer guide uses 1.x terms and should be updated to reflect new terms in Juju 2.0, e.g. 'controller' and 'model'

mbruzek:
@evilnick Can you be more specific on the terms that need to be replaced or in which files you see this problem? I did a search and find "controller" in many of the documents.

I read introducing-2.md and understand "state server" is replaced with "controller" and there are several models per controller. I don't see "state sever" in the developer documentation either.

evilnick:
I think some instances have been updated due to other changes . The things to watch out for are:

environment - in a lot of cases this should be updated to reference a model
state-server - we prefer to use 'controller' now. I think it may be refernced a few times in the old 'authors' docs.
the big one is 'services', which are now 'applications'

i changed many of these myself, but there are loads of references to services, understandably.

mbruzek:
I will need @johnsca 's help deciphering developer-layers-interfaces.md that document uses the word "service" 39 times!

Issue #879, opened 2016-02-09
'About Juju' page needs a review


pmatulis:
The [About Juju](https://jujucharms.com/docs/stable/about-juju) page is badly in need of a review. I guess this should be written with Juju 2.0 in mind.

pmatulis:
not an orchestrator
an application modelling tool

pmatulis:
![](https://github.trello.services/images/mini-trello-icon.png) [Rewrite About-Juju page](https://trello.com/c/XtoKtYZQ/433-rewrite-about-juju-page)

pmatulis:
Put this also into Discourse.

pmatulis:
In the meantime, the about-juju page has been removed from the menu.

PR #1221, opened 2016-06-30
About juju 2.0

mbruzek:
Issue #802 indicates we have to update the language to use Juju 2.0 terms. I searched for "service" and "services" in the docs and about-juju.md was the first page in the results that needed many updates.

This is my first crack at the terminology update and will likely need some word-smithing.


pmatulis:
Thank you. :candy: 

This page still needs a fundamental refresh and is tracked with #879 .

PR #1226, opened 2016-07-01
Changed service to application and environment to model.

mbruzek:
Also fixed up some paragraphs to support the terminology change.

Please forgive my editor removing trailing whitespace. I realize that is annoying to review in a diff. I only
noticed my editor was removing whitespace about half way through this adventure.

### DIFF
--- a/src/en/actions.md
+++ b/src/en/actions.md
@@ -2,7 +2,7 @@ Title: Juju actions
 
 # Juju Actions
 
-Juju charms can describe *actions* that users can take on deployed services.
+Juju charms can describe *actions* that users can take on deployed applications.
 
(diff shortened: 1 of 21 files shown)


mbruzek:
This is a small step toward issue #802 but I hesitate to use the word "fixes" because there are many other documents that need the updated terminology.

evilnick:
Yes, it is annoying! 
But thanks for your efforts here. I think there is more language to be rooted out.
Also, I know it can be a pain but generally best to stick to low number of files per PR - reduces the risk of conflicts and makes them easier to fix if they do happen.

evilnick:
:cake:

PR #1231, opened 2016-07-05
fixes misstated support for network spaces

### LINKED ISSUE #1210: Support statement for spaces is incorrect
In this page:  https://jujucharms.com/docs/devel/network-spaces

It is stated that spaces are only supported for AWS.  It is currently only supported for MAAS.

matthewhelmke:
Fixes #1210 
@juju/docs ready for review, should be quick


pmatulis:
Spaces on MAAS only. If you say so. :smile: 
:candy:

PR #1232, opened 2016-07-05
more cleanup was needed

### LINKED ISSUE #1210: Support statement for spaces is incorrect

matthewhelmke:
After #1231 was merged, I noticed further cleanup was needed. I feel like this whole page needs some editing, once we better understand the feature and can get it up and running somewhere where the docs team can test it. Until then, at least I have removed the remaining mentions of AWS in favor of only MAAS.

@juju/docs for a second set of edits to fix #1210

### DIFF
--- a/src/en/network-spaces.md
+++ b/src/en/network-spaces.md
@@ -93,7 +93,7 @@ For more information regarding constraints in general, see "juju help constraint
 
 ## Example
 
-Let's model the following deployment in Juju on AWS:
+Let's model the following deployment in Juju:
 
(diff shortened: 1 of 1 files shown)


pmatulis:
:candy: Thanks for the cleanup. What do we need in order to test this stuff? :candy:

matthewhelmke:
I think we just need the networking set up on our "new" servers and we can then do a MAAS install and start testing the instructions ourselves to find any gaps, etc.

PR #1237, opened 2016-07-08
Added credential info for add-model, remove outdated bug reports

### LINKED ISSUE #1174: Expand info about creating new models
This would be useful in models.md.

Controller admin users can create new models without needing to specify
any additional configuration:

    juju add-model mynewmodel

In such cases, the new model will inherit the credentials and authorized
ssh keys of the admin model.

Where a cloud requires credentials, non-admin users, and admin users if
they wish, are required to specify a named credential (so that resources
created by the new model are allocated to the cloud account of the model
creator), as well as authorized keys:

     juju add-model mynewmodel --credential mysecrets
        --config authorized-keys="ssh-rsa ...."

Additional model config just for the new model may also be specified:

    juju add-model --config image-stream=daily

evilnick:
Fixes #1174 

There is certainly potential to expand this further, but at the moment this at least gives examples of --credential usage, and using the non-current controller

### DIFF
--- a/src/en/models-adding.md
+++ b/src/en/models-adding.md
@@ -1,42 +1,44 @@
-Title: Adding a Juju Model
-TODO: Review again soon (created: March 2016)
-      Remove Note re bug 1552469 when fixed; remove Note and --config option from Rackspace example
-
+Title: Adding a model
 
 # Adding a model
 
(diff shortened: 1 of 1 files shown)


evilnick:
@matthewhelmke

matthewhelmke:
Very nice. LGTM
:gift:

Issue #1185, opened 2016-06-16
Review and update juju status output


pmatulis:
Review where output to `juju status` is displayed and update it. So far I only found commands.md but we can't change that. There may be more.

https://jujucharms.com/docs/devel/commands#status

PR #1239, opened 2016-07-08
deleted extra spaces and added = in a couple places where the help…

matthewhelmke:
…output shows it should be even though the command switch works with or without the = sign

I checked all files and all instances of 'juju status' examples in the docs (on 31 different pages).

These few were the only ones needing adjustment, and even then it seems optional. I tested '--format <option>' and '--format=<option>' and both work. Since the help text and the list on commands.md show and equals sign, I went ahead and put it in for consistency.

### DIFF
--- a/src/en/authors-charm-writing.md
+++ b/src/en/authors-charm-writing.md
@@ -1,4 +1,4 @@
-Title: Writing your first Juju charm  
+Title: Writing your first Juju charm
 
 
 # Your first charm starts here!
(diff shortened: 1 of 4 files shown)


matthewhelmke:
@evilnick 
This will be a quick and easy one

matthewhelmke:
Closes #1185

evilnick:
OK, good :cake:

PR #1243, opened 2016-07-11
rephrased in places for enhanced clarity

matthewhelmke:
Our steps were correct. I just rephrased and reorganized a little to (hopefully) make the details clearer to readers. I eliminated the creation of variables in our steps and added two notes to make it clear what should happen when we reference variable names.

Also, the install location of the Azure CLI tool has changed, so I updated that info.

@pmatulis for review

### DIFF
--- a/src/en/help-azure.md
+++ b/src/en/help-azure.md
@@ -59,8 +59,8 @@ therefore be easier to perform this on a graphical desktop.
 
 ### Registering azure services
 
-Juju requires certain services to be active for your account. Use these 
-to register using the Azure CLI tool:
+Juju requires certain services to be active for your account. Enter these 
+commands to register using the Azure CLI tool:
 
 ```
 azure provider register Microsoft.Compute
--- a/src/en/help-azure.md
+++ b/src/en/help-azure.md
@@ -81,13 +81,13 @@ the actual ID that corresponds to the variable.
 
 ### `subscription-id`
 
-List your account and get the subscription ID, the **SUB_ID**:
+List your account. Note the subscription ID, the **SUB_ID**.
 
 ```bash
 azure account list
 ```
 
-Sample output:
+**SUB_ID** will appear on a line like this:
 
 ```no-highlight
 info:    Executing command account list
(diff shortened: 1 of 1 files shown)


src/en/help-azure.md
```diff
@@ -56,8 +59,8 @@ therefore be easier to perform this on a graphical desktop.
 
 ### Registering azure services
 
-Juju requires certain services to be active for your account. This can 
-be done with the Azure CLI tool:
+Juju requires certain services to be active for your account. Use these 
+to register using the Azure CLI tool:
```
evilnick: use these commands?

src/en/help-azure.md
```diff
@@ -90,24 +97,21 @@ data:    Free Trial  f717c8c1-8e5e-4d38-be7f-ed1e1c879e18  true     Enabled
 info:    account list command OK
 ```
 
-The subscription ID can now be stored in a local environment variable, by
-entering:
+In this sample,
 
 ```bash
 SUB_ID=f717c8c1-8e5e-4d38-be7f-ed1e1c879e18
 ```
```
pmatulis: We do not show output to the other commands. Why show it for just this command? Also, to be consistent with the below stuff, we shoulds say "Note the application ID, the **SUB_ID**. It will look similar to:...".

src/en/help-azure.md
```diff
@@ -179,12 +187,13 @@ You can now run the interactive command:
 juju add-credential azure
 ```
 
-Which will ask for a credential name, and then the values discovered above.
+Which will ask for a credential name, which you determine for yourself, and
+then the values discovered above.
```
pmatulis: The phrase "which  you determine for yourself" is vague. It can be construed as yet another variable which you need to find out, but on your own. What about "an arbitrary name" or similar?

We should also list the entities it will ask for:

APP_ID
SUB_ID
TENANT_ID
APP_PASSWORD

pmatulis:
Review complete.

src/en/help-azure.md
matthewhelmke: I included more output because Azure does not label the item in the output with "subscription ID" or "SUB_ID" or anything else easily identifiable. Adding the full output, since it was only 5 lines, seemed to me to provide a road sign to more easily find what we are referring to. I'll try to further clarify what is going on here.

matthewhelmke:
Thanks for the helpful comments!
:fish_cake:

PR #1244, opened 2016-07-12
Added cloud help pages to main navigation

evilnick:
People were getting lost so I added these cloud specific pages back to the main navigation


evilnick:
:+1:  :cookie:

PR #1248, opened 2016-07-14
removed ellipses and added spaces to properly format code quote

### LINKED ISSUE #1247: Missing <pre> in temp-release-notes
https://jujucharms.com/docs/devel/temp-release-notes

is missing some sort of <pre> stanza around these lines:
<p>Example bundle.yaml excerpt:
    ...
    mysql:
        charm: cs:trusty/mysql-53
        num_units: 1
        constraints: mem=4G
        bindings:
            server: database
            cluster: internal
    ...</p>

matthewhelmke:
Fixes #1247 
@juju/docs Ready for review

### DIFF
--- a/src/en/temp-release-notes.md
+++ b/src/en/temp-release-notes.md
@@ -585,7 +585,7 @@ when deploying individual charms. The bundle YAML can include a section
 called "bindings", defining the map of endpoint names to space names.
 
 Example bundle.yaml excerpt:
-    ...
+
     mysql:
         charm: cs:trusty/mysql-53
         num_units: 1
@@ -593,7 +593,6 @@ Example bundle.yaml excerpt:
         bindings:
             server: database
             cluster: internal
-    ...
 
 Deploying a bundle including a section like in the example above, is
 equivalent to running:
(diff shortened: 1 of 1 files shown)


src/en/temp-release-notes.md
```diff
@@ -585,15 +585,14 @@ when deploying individual charms. The bundle YAML can include a section
 called "bindings", defining the map of endpoint names to space names.
 
 Example bundle.yaml excerpt:
-    ...
+
```
evilnick: for the purposes of docs, it is best to code fence these so the highlighting (may) work:

```yaml
   
```

evilnick: but not indented - that was just to make it show up!

matthewhelmke: I'm happy to do that. I just followed most of the (currently rendered properly) other code boxes on the page. Shall I go through the whole thing?

evilnick: no, it's fine. they are only temporary after all - I just thought as you were fixing it, you might as well fix it nicely :)

matthewhelmke:
They are better with the backticks and code type indicator...

evilnick:
:cake: 
We can fancify it better when there is a proper release

PR #1250, opened 2016-07-14
Release 1.25.6 is the current proposed release.

bz2:
Updating current proposed juju as part of the release process for upcoming stable 1.25.6 version.


evilnick:
thanks :beer:

PR #1254, opened 2016-07-15
Add information about turning off auto-retry of hooks.

### LINKED ISSUE #1222: Model configuration has an option to disable automatic retry of hooks
As per conversation in #1217

This should be added to model/controller config options and referenced from the charm developer docs

evilnick:
Fixes #1222 

@matthewhelmke


matthewhelmke:
LGTM
:beer:

PR #1260, opened 2016-07-19
suggest use of discrete resources

jrwren:


chuckbutler:
@jrwren - can you give us some more accurate file size limits in there?

I know that we don't want to put hard limits, but if we have a range of file sizes we are testing with, it may be good to list those sizes inline in that doc note.

"We have tested with increments of 5mb, 33 MB, 125mb, and 150mb"    that gives the operator some clue that their "dependency" of another vendors package which sums up to 600MB may not make the cut.

Nebulous wording around this feature is only going to serve to further punish users who try it with large files :(

jrwren:
This suggestion is not about the file size limits. It is about helping charms execute quickly and allowing juju to use optimizations which are in place.

If ever there is a choice of bundling all resources into one versus keeping them separate, prefer keeping them separate for the ability to update them independently. We want to document the antipattern that may be discovered when someone needs 3 dependency tarballs and decides to  group them together.

To answer your question directly: There are no hard limits on the file sizes. We want to support large files. There is currently no hard limit on file size. The errors which have observed are related to load of the underlying system storing the files. At any given time that load can be different and the file size attempted to upload may fail, even at a small size if the load is heavy. We are working on improving this situation.

chuckbutler:
That makes sense, I was simply trying to play advocate for the people who are not tracking this conversation and have no idea. Its likely they will make an incorrect assumption based on lack of information.

But I concede to your point, we want to empower them not limit them.

chuckbutler:
+1 LGTM

evilnick:
Cool. there is a small typo but I will merge and fix.

:candy:

Issue #1265, opened 2016-07-21
Manual Provisioning broken link


nskaggs:
We're missing documentation on configuring manual provider for 2.0.

We're redirecting https://jujucharms.com/docs/stable/ to 2.0 docs, thus causing this link in particular is now a 404 https://jujucharms.com/docs/stable/config-manual.

evilnick:
Where is the link directing people to https://jujucharms.com/docs/stable/config-manual ?

There are docs for setting up a manual cloud https://jujucharms.com/docs/stable/clouds-manual
These are including in the main navigation and referenced from other pages within the 2.0 docs

If this is a link from an external source, we should either change it if possible, or set up a redirect

nskaggs:
https://jujucharms.com/docs/1.25/config-manual, https://jujucharms.com/docs/1.24/config-manual has a banner at the top as well.

Weird I couldn't find it. I would simply redirect https://jujucharms.com/docs/stable/config-manual  to https://jujucharms.com/docs/stable/clouds-manual then.

evilnick:
Okay, I will take a scrub through the old docs as well, they shouldn't reference stable/devel for precisely this reason!

evilnick:
The archive docs banner thing is a jujucharms.com issue. I filed https://github.com/CanonicalLtd/blues_browser/issues/1320

evilnick:
Aha! Sorry, I have just twigged. You are saying that when looking at the page in the 1.25 docs, you click on 'stable' above the navigation menu and you get a 404?

That's completely the JS on the jujucharms site - it should, if there is no equivalent page, be crossed out with no link. We can make a redirect for this case and maybe some of the other config pages, but in general the behaviour is wrong. I will raise another issue for that

On the plus side, I *did* find some wrongly specific links in the 1.25 docs, so thanks!

evilnick:
I don't think the Jujucharms code has been corrected, but we have redirects for all the old pages so from a docs perspective this is closed.

PR #1266, opened 2016-07-21
Fixed erroneous links -> /stable/

### LINKED ISSUE #1259: Digital Ocean provider is no longer maintained
The DO provider references old work by Kapil that has not been updated to work with the newer DO API, thus no longer functions in its current form

[doc page](https://jujucharms.com/docs/stable/config-digitalocean)

[Repo](https://github.com/kapilt/juju-digitalocean)

I don't think these pages were forward ported to the 2.0 documentation. It's probably worthwhile to deprecate and remove these pages from the 1.25 docs otherwise users will find themselves in broken provider territory, and it's not due to juju-core keeping up with the provider spec.

evilnick:
Fixes some links incorrectly made to /stable/ docs rather than relative links. (#1265)
Also removes DigitalOcean config as this no longer supported

Fixes #1259


evilnick:
@matthewhelmke note this is for the 1.25 branch

matthewhelmke:
@evilnick good to know
LGTM
:cake:

PR #1276, opened 2016-08-03
updated steps and graphics following changes at GCE

matthewhelmke:
@evilnick, please check closely to ensure I have named the changed files correctly. The order of some steps changed and I followed the existing order of filenames in the actual file while retaking screenshots...for better or for worse. Let me know if that is acceptable or whether you prefer something different.


evilnick:
Yup, this all looks good. Thanks!

i will backport it to 2.0, but we should also backport to 1.25, which may take some extra work

evilnick:
:custard:

PR #1291, opened 2016-08-11
Add GUI to navigation.

evilnick:
Minor updates to controller

@matthewhelmke

### DIFF
--- a/src/navigation.tpl
+++ b/src/navigation.tpl
@@ -23,7 +23,14 @@
                 <li><a href="help-rackspace.html">Rackspace</a></li>
             </ul>
         </li>
-        <li><a href="controllers.html">Controllers</a></li>
+        <li class="section><a href="controllers.html">Controllers</a>
+            <ul class="sub">
+                <li><a href="controllers-creating.html">Creating/Bootstrap</a></li>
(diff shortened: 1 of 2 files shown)


src/navigation.tpl
```diff
@@ -23,7 +23,14 @@
                 <li><a href="help-rackspace.html">Rackspace</a></li>
             </ul>
         </li>
-        <li><a href="controllers.html">Controllers</a></li>
+        <li class="section><a href="controllers.html">Controllers</a>
+            <ul class="sub">
+                <li><a href="controllers-creating.html">Creating/Bootstrap</a></li>
```
matthewhelmke: We have a gerund/imperative combo here with Creating/Bootstrap. Should we pick one form and use it for both? Either "Create/Bootstrap" or "Creating/Bootstrapping"?

matthewhelmke:
Only one comment. Otherwise, LGTM.
:v:

src/navigation.tpl
evilnick: yeah, I hate 'bootstrapping'. fair enough, I will change it to Create

evilnick:
fixed. thx!

PR #1294, opened 2016-08-12
the tables were hard to read when formatted as code so I reformatted …

matthewhelmke:
…into actual tables

@evilnick any reason I shouldn't do this?


evilnick:
No, that's good work. I am concerned with it getting overwritten at some point, but...
:cookie:

PR #1296, opened 2016-08-12
moved the asterick notes under the first fixed table

matthewhelmke:
because they are currently mis-styled as a single bullet point. I put them all back as a code block, which doesn't feel ideal, but should be clearer than the jumbled mess they are atm. I'm happy to do something different if there is a better solution.

@evilnick 
Any idea why the top left cell in each table is the only cell following the markdown style of "left align" while all the other cells have been centered, even though that isn't in the markdown?

### DIFF
--- a/src/en/temp-release-notes.md
+++ b/src/en/temp-release-notes.md
@@ -116,10 +116,10 @@ Juju commands have moved to a flat command structure instead of nested command s
 | juju subnet add                       | juju add-subnet *** |
 | juju ensure-availability              | juju enable-ha *** |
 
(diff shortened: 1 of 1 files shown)


evilnick:
yes, because the dodgy css doesn't work properly...

I don't think the code style works well either. Maybe just bold them up. It may be something to do with them being actual asterisks - markdown is quite forgiving, but they should probably be escaped:
\*a literal asterisk\*
```\*a literal asterisk\*``
not *asterisks*

matthewhelmke:
Doh! That seems obvious in retrospect. Thanks!

Issue #1299, opened 2016-08-17
Service level status not reflected in docs for status-{get,set}


sparkiegeek:
- https://jujucharms.com/docs/devel/authors-hook-environment#status-set
- https://jujucharms.com/docs/devel/authors-hook-environment#status-get

Both `status-set` and `status-get` hook commands have a `--service` flag which allows the leader of the service to set/get the overall status of the service. This should be documented beyond just the `--help` of the hook tools (hint: `juju run --unit ubuntu/0 'status-get --help'`).

The option has been there since the introduction of the commands, so docs for 1.24 onwards should be updated appopriately.

matthewhelmke:
To help, I did a fresh install/bootstrap/deploy and got the help text referred to above.

`ubuntu@20160812-beta15:~$ juju deploy ubuntu
Added charm "cs:ubuntu-0" to the model.
Deploying charm "cs:ubuntu-0" with the user specified series "xenial".
ubuntu@20160812-beta15:~$ juju status
MODEL    CONTROLLER  CLOUD/REGION         VERSION
default  lxd-test    localhost/localhost  2.0-beta15

APP     VERSION  STATUS   EXPOSED  ORIGIN      CHARM   REV  OS
ubuntu           unknown  false    jujucharms  ubuntu  0    ubuntu

UNIT      WORKLOAD  AGENT  MACHINE  PUBLIC-ADDRESS  PORTS  MESSAGE
ubuntu/0  unknown   idle   0        10.27.236.78           

MACHINE  STATE    DNS           INS-ID         SERIES  AZ
0        started  10.27.236.78  juju-028a1a-0  xenial  



ubuntu@20160812-beta15:~$ juju run --unit ubuntu/0 'status-get --help'
Usage: status-get [options] [--include-data] [--application]

Summary:
print status information

Options:
--application  (= false)
    print status for all units of this application if this unit is the leader
--format  (= smart)
    Specify output format (json|smart|yaml)
--include-data  (= false)
    print all status data
-o, --output (= "")
    Specify an output file

Details:
By default, only the status value is printed.
If the --include-data flag is passed, the associated data are printed also.


ubuntu@20160812-beta15:~$ juju run --unit ubuntu/0 'status-set --help'
Usage: status-set [options] <maintenance | blocked | waiting | active> [message]

Summary:
set status information

Options:
--service, --application  (= false)
    set this status for the application to which the unit belongs if the unit is the leader

Details:
Sets the workload status of the charm. Message is optional.
[...]

PR #1300, opened 2016-08-17
added info on --service flag

matthewhelmke:
@mbruzek I hope I'm not stepping on ecoteam toes by adding this in for you. I had time and since I had already commented on #1299 I thought I would do more than just give info and try to help. I would appreciate it if you would read my additions and confirm if they are correct or give edits before this is merged. Thanks in advance!

I also did some minor cleanup in the file removing extraneous spaces.

### DIFF
--- a/src/en/reference-hook-tools.md
+++ b/src/en/reference-hook-tools.md
@@ -540,10 +540,10 @@ status-get
 status-get --include-data
 ```
 
-Use `--service` to get the overall status for the service to which the unit
-belongs if the unit is the leader, for example:
+Use `--application` to get the overall status for the application to which
+the unit belongs if the unit is the leader, for example:
 ```bash
-juju run --unit ubuntu/0 'status-get --service
+juju run --unit ubuntu/0 'status-get --application
 ```
 
 ## status-set
--- a/src/en/reference-hook-tools.md
+++ b/src/en/reference-hook-tools.md
@@ -1,4 +1,4 @@
-Title: Hook tools  
+Title: Hook tools
 
 # Hook tools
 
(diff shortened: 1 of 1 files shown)


src/en/reference-hook-tools.md
```diff
@@ -527,19 +527,24 @@ the current workload status. Without arguments, it just prints the workload
...
 ```
-bash:  
+bash:
 ```bash
 status-get
 
 status-get --include-data
 ```
 
+Use `--service` to get the overall status for the service to which the unit
```
marcoceppi: Are we sure it's not --application now?

matthewhelmke: I am not. Both --service and --application are in the help text for status-set, but only --application is in the help text for status-get.

Hmm.

matthewhelmke:
Added this comment in #1299 and meant to add it here. I'll copy/paste and link from here.

The help text I included in #1299  is from beta15. I haven't yet been able to check for changes in beta16 due to https://bugs.launchpad.net/juju/+bug/1618636

When I am able to check a newer release, I'll see whether the --service flag has disappeared. Keeping this on hold meanwhile.

matthewhelmke:
beta18 still includes the help text with both --application and --service.

matthewhelmke:
rc1 still includes both flags.

src/en/reference-hook-tools.md
mitechie: If the help text shows --service then it's a bug as we've completely renamed service to application. We support --service so that we don't break existing charms that have not had a chance to update, but our visible face (including documentation) needs to be the one true way forward with is --application. 

This should be updated to --application and anywhere you see service in the user facing juju output needs to have a bug filed against it so that it gets corrected.

matthewhelmke:
I think this whole page needs to be gone over with a fine toothed comb, but I'll file a new issue for that.

Issue #1161, opened 2016-06-09
Add resource-get to Charm documentation


matthewhelmke:
This is a new charmer concept that should be documented somewhere appropriate in the Charms section.

Use 'resource-get' while a hook is running to get the local path
to the file for the identified resource. This file is an fs-local
copy, unique to the unit for which the hook is running. It is
downloaded from the controller, if necessary.

If 'resource-get' for a resource has not been run before (for the unit)
then the resource is downloaded from the controller at the revision
associated with the unit's service. That file is stored in the unit's
local cache. If 'resource-get' **has** been run before then each
subsequent run syncs the resource with the controller. This ensures that
the revision of the unit-local copy of the resource matches the revision
of the resource associated with the unit's service.

Either way, the path provided by 'resource-get' references the up-to-
date file for the resource. Note that the resource may get updated on
the controller for the service at any time, meaning the cached copy
**may** be out of date at any time after you call 'resource-get'.
Consequently, the command should be run at every point where it is
critical that the resource be up to date.

The 'upgrade-charm' hook will be fired whenever a new resource has
become available. Thus, in the hook the charm may call 'resource-get',
forcing an update if the resource has changed.

Note that 'resource-get' only provides an FS path to the resource file.
It does not provide any information about the resource (e.g. revision).

matthewhelmke:
Closed via #1302

PR #1302, opened 2016-08-18
added info about resource-get

### LINKED ISSUE #1301: get-resource missing from hook-tools reference
I think that the new get-resource command documented here:
     https://jujucharms.com/docs/stable/developer-resources
Should also be in the list of hook-tools, documented here:
     https://jujucharms.com/docs/stable/reference-hook-tools

matthewhelmke:
@mbruzek another one that I would like you to double check, please.

Fixes #1301

### DIFF
--- a/src/en/reference-hook-tools.md
+++ b/src/en/reference-hook-tools.md
@@ -520,6 +520,31 @@ relation-set -r server:3 username=jim password=12345
 ```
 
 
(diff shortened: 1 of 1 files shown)


matthewhelmke:
Also fixes #1161

chuckbutler:
👍

PR #1303, opened 2016-08-18
added info about network-get

### LINKED ISSUE #1200: network-get hook environment tool undocumented

matthewhelmke:
@mbruzek one more for your review. This one would benefit from the addition of a Python example in the next round of edits.

Fixes #1200

### DIFF
--- a/src/en/reference-hook-tools.md
+++ b/src/en/reference-hook-tools.md
@@ -248,6 +248,17 @@ leader-set cluster-leader-address=10.0.0.123
 ```
 
 
+## network-get
+`network-get` returns the network config for a given binding name. The only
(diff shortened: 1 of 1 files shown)


chuckbutler:
This is still really minimal but +1 LGTM to merge. It at least gives it an entry in the reference doc until we get better documentation around modeling networks with Juju.

PR #1312, opened 2016-08-25
Release devel 2.0 beta16

nskaggs:
Also add note about getting beta from snap store

### DIFF
--- a/src/en/temp-release-notes.md
+++ b/src/en/temp-release-notes.md
@@ -71,70 +71,72 @@ The "state-server" from Juju 1.x becomes a "controller" in 2.0.
 
 Juju commands have moved to a flat command structure instead of nested command structure:
 
-    1.25 command                          2.0-beta command
+| 1.25 command                          | 2.0-beta command |
+| --- | --- |
+| juju environment destroy              | juju destroy-model * *** |
+| juju environment get                  | juju get-model-config *** |
+| juju environment get-constraints      | juju get-model-constraints ** |
+| juju environment retry-provisioning   | juju retry-provisioning |
+| juju environment set                  | juju set-model-config *** |
+| juju environment set-constraints      | juju set-model-constraints ** |
+| juju environment share                | juju share-model *** |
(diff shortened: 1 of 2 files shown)


src/en/temp-release-notes.md
```diff
@@ -69,72 +71,70 @@ The "state-server" from Juju 1.x becomes a "controller" in 2.0.
...
-| juju space create                     | juju add-space *** |
-| juju space list                       | juju list-spaces *** |
-| juju subnet add                       | juju add-subnet *** |
-| juju ensure-availability              | juju enable-ha *** |
-
-\* the behaviour of destroy-environment/destroy-model has changed, see
-https://jujucharms.com/docs/devel/controllers
-\*\* these commands existed but are now the recommended approach
-\*\*\* this is an alias, but will be the primary command going forward
+    1.25 command                          2.0-beta command
```
matthewhelmke: Curious why we are reformatting this from an easy-to-read table to a code block that is harder to read...

src/en/temp-release-notes.md
```diff
@@ -819,8 +819,8 @@ relevant sub-commands:
 
 For more information on the new support for channels in the Charm Store
 and how they work, please see our
-[documentation](https://jujucharms.com/docs/devel/authors-charm-store#entities-explained)
-on the subject.
+[documentation](https://jujucharms.com/docs/devel/authors-charm-store
+#entities-explained) on the subject.
```
matthewhelmke: The entire link needs to be on one line or it breaks the link. This line should not have been changed.

matthewhelmke:
I left a couple of comments. Other than those, this looks good.

src/en/temp-release-notes.md
nskaggs: I'm not sure; I can simply revert the change and we can ask about it in the doc itself.

nskaggs:
Got rid of unwanted stuff, ty.

matthewhelmke:
ty
:beer:

Issue #1313, opened 2016-08-29
Broken link


bac:
At the bottom of 

https://jujucharms.com/docs/stable/developer-layers

```The Getting started guide contains steps on how to build a layer into a charm.```

has a link to https://jujucharms.com/docs/stable/devel/developer-getting-started#assemble-the-layers , which does not exist.

bac:
The links in that file are not relative as they should be, which leads to the 'stable/devel' situation above.

marcoceppi:
Good point, I'll open a new PR

PR #1314, opened 2016-08-29
Tighten messaging around states and hooks

marcoceppi:
Needs to be backported to 2.0


bac:
Well, that was fast.
I was hoping the fix for https://github.com/juju/docs/issues/1313 could be included in this branch.

pmatulis:
@marcoceppi I found a broken link in this PR.

> Consider the example illustrated in the `[Getting Started]()` guide. `apache.available` is set
from the apache-php layer.

What should this link be? FWIW I do not see an occurrence of `apache.available` in https://jujucharms.com/docs/devel/developer-getting-started

PR #1330, opened 2016-09-02
Added note about Azure core limit

### LINKED ISSUE #1329: Azure core limits
Azure apparently limits standard accounts to 10 cores. We should note this in the setup page to forewarn users who otherwise may not know why deploys fail.

evilnick:
Fixes #1329


evilnick:
:+1:

PR #1333, opened 2016-09-02
added info about multi-series charms

### LINKED ISSUE #1157: Add info about multi-series charms
Charms now have the capability to declare that they support more than
one series. Previously a separate copy of the charm was required for
each series.

This should be noted somewhere in the Charms section of our documentation.

An important constraint here is that for a given charm,
all of the listed series must be for the same distro/OS; it is not
allowed to offer a  single charm for Ubuntu and CentOS for example.
Supported series are added  to charm metadata as follows:
```
    name: mycharm
    summary: "Great software"
    description: It works
    maintainer: Some One <some.one@example.com>
    categories:
       - databases
    series:
       - precise
       - trusty
       - wily
    provides:
       db:
         interface: pgsql
    requires:
       syslog:
         interface: syslog
```
The default series is the first in the list:
```
    juju deploy mycharm
```
will deploy a mycharm service running on precise.

A different, non-default series may be specified:
```
    juju deploy mycharm --series trusty
```
It is possible to force the charm to deploy using an unsupported series
(so long as the underlying OS is compatible):
```
    juju deploy mycharm --series xenial --force
```
or
```
    juju add-machine --series xenial
    Machine 1 added.
    juju deploy mycharm --to 1 --force
```
'--force' is required in the above deploy command because the target
machine  is running xenial which is not supported by the charm.

The 'force' option may also be required when upgrading charms. Consider
the  case where a service is initially deployed with a charm supporting
precise  and trusty. A new version of the charm is published which only
supports  trusty and xenial. For services deployed on precise, upgrading
to the newer  charm revision is allowed, but only using force (note the
use of  '--force-series' since upgrade-charm also supports '--force-
units'):
```
    juju upgrade-charm mycharm --force-series
```

matthewhelmke:
Fixes #1157 
@juju/docs
@evilnick 

This is more detail than you initially suspected we might need to add, but after reading through the existing info more closely, I think it is warranted.

### DIFF
--- a/src/en/charms-deploying.md
+++ b/src/en/charms-deploying.md
@@ -100,6 +100,73 @@ juju charm-upgrade mysql --channel edge
 ```
 
 
+## Deploying a multi-series charm
+
+Charms can be created that support more than one release of a given operating
+system distro, such as the multiple Ubuntu releases shown below. It is not
+possible to create a charm to support multiple distros, such as one charm for
+both Ubuntu and CentOS. Supported series are added to the charm metadata like
+this:
+
+```
+name: mycharm
+summary: "Great software"
+description: It works
+maintainer: Some One <some.one@example.com>
+categories:
+   - databases
+series:
+   - trusty
+   - xenial
+provides:
+   db:
+     interface: pgsql
+requires:
+   syslog:
+     interface: syslog
+```
+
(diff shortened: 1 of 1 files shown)


evilnick:
No, that's good, I like it
:+1:  :+1:  :+1:

Issue #1249, opened 2016-07-14
Constraints document needs updating


nskaggs:
The current constraints docs were last updated in 1.20. It's missing the list of valid options for constraints, as well as examples of machine type constraints. For example, constraints around containers, virt-type, machine properties, etc.

nskaggs:
@anastasiamac perhaps you can help here?

anastasiamac:
@nskaggs Sure thing \o/ I'll update it as part of the work I am currently doing for fixing virt type in image constraints :D Thank you for heads up!

evilnick:
any progress on this @anastasiamac ? Would be useful to have a list if you have one, otherwise we will have to go on a hunt through the code again...

anastasiamac:
@evilnick
I can give you all information I have. It's a little bit jumbled and I would really appreciate if you could polish and make it understandable \o/ It's applicable to current Juju 2.0-beta16.

These are all possible constraints that can be specified:
    * "arch";
    * "container";
    * "cpu-cores";
    * "cpu-power"
    * "mem";
    * "root-disk";
    * "tags";
    * "instance-type";
    * "spaces";
    * "virt-type".
Different providers will support different constraints and, sometimes, different values for these. Different providers also dictate different conflicting constraints, i.e constraints attributes that can not be used in combination.

###Azure Provider:
    * Unsupported: [cpu-power, tags, virt-type]
    * Valid values: arch=[amd64]; instance-type=[defined on the cloud]
    * Conflicting constraints: [instance-type] vs [mem, cpu-cores, arch]

###Cloudsigma (currently behind development flag):
    * Unsupported: [container, instance-type, tags, virt-type]

###EC2 Provider:
    * Unsupported: [tags, virt-type]
    * Valid values: instance-type=[defined on the cloud]
    * Conflicting constraints: [instance-type] vs [mem, cpu-cores, cpu-power]

###GCE Provider:
    * Unsupported: [tags, virt-type]
    * Valid values: instance-type=[defined on the cloud]; container=kvm
    * Conflicting constraints: [instance-type] vs [arch, cpu-cores, cpu-power, mem, container]

###Joyent Provider:
    * Unsupported: [cpu-power, tags, virt-type]
    * Valid values: instance-type=[defined on the cloud]

###LXD Provider:
    * Unsupported: [cpu-cores, cpu-power, instance-type, tags, virt-type]
    * Valid values: arch=[host arch]

###MAAS Provider:
[...]

PR #1336, opened 2016-09-06
updated constraints and removed pre 1.0 legacy section

matthewhelmke:
Do not merge yet--I'm waiting for an answer as to whether to keep the existing `networks` constraint info or whether that constraint is no longer valid. Once answered, I will either update or we can merge as-is.

### DIFF
--- a/src/en/reference-constraints.md
+++ b/src/en/reference-constraints.md
@@ -1,16 +1,16 @@
-Title: Juju constraints  
+Title: Juju constraints
 
 # Constraints
 
-Constraints constrain the possible instances that may be started by juju
+Constraints set limits on the possible instances that may be started by juju
 commands. They are usually passed as a flag to commands that provision a
 new machine (such as bootstrap, deploy, and add-machine). See [using
 constraints](charms-constraints.html) for how to specify these in a
 deployment.
 
 Each constraint defines a minimum acceptable value for a characteristic of a
-machine.  Juju will provision the least expensive machine that fulfils all the
-constraints specified.  Note that these values are the minimum, and the actual
+machine. Juju will provision the least expensive machine that fulfils all the
+constraints specified. Note that these values are the minimum, and the actual
 machine used may exceed these specifications if one that exactly matches does
 not exist.
 
@@ -19,14 +19,14 @@ environment, no machine will be provisioned, and an error will be printed in the
 machine's entry in juju status.
 
 Constraint defaults can be set on an environment or on specific services by
-using the set-constraints command (see juju help set-constraints).  Constraints
+using the set-constraints command (see juju help set-constraints). Constraints
 set on the environment or on a service can be viewed by using the get-
-constraints command.  In addition, you can specify constraints when executing a
+constraints command. In addition, you can specify constraints when executing a
 command by using the --constraints flag (for commands that support it).
 
 Constraints specified on the environment and service will be combined to
 determine the full list of constraints on the machine(s) to be provisioned by
-the command.  Service-specific constraints will override environment-specific
+the command. Application-specific constraints will override environment-specific
 constraints, which override the juju default constraints.
 
 Constraints are specified as key value pairs separated by an equals sign, with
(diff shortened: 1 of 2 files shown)


matthewhelmke:
Will eventually fix #1249

src/en/reference-constraints.md
```diff
@@ -1,16 +1,16 @@
-Title: Juju constraints  
+Title: Juju constraints
 
 # Constraints
 
-Constraints constrain the possible instances that may be started by juju
+Constraints set limits on the possible instances that may be started by juju
```
evilnick: Juju

src/en/reference-constraints.md
```diff
@@ -19,14 +19,14 @@ environment, no machine will be provisioned, and an error will be printed in the
 machine's entry in juju status.
 
 Constraint defaults can be set on an environment or on specific services by
-using the set-constraints command (see juju help set-constraints).  Constraints
+using the set-constraints command (see juju help set-constraints). Constraints
```
evilnick: backtick the command

src/en/reference-constraints.md
```diff
@@ -19,14 +19,14 @@ environment, no machine will be provisioned, and an error will be printed in the
...
 set on the environment or on a service can be viewed by using the get-
-constraints command.  In addition, you can specify constraints when executing a
+constraints command. In addition, you can specify constraints when executing a
 command by using the --constraints flag (for commands that support it).
```
evilnick: I would probably backtick '--constraints' too

src/en/reference-constraints.md
```diff
@@ -19,14 +19,14 @@ environment, no machine will be provisioned, and an error will be printed in the
...
 
 Constraints specified on the environment and service will be combined to
 determine the full list of constraints on the machine(s) to be provisioned by
-the command.  Service-specific constraints will override environment-specific
+the command. Application-specific constraints will override environment-specific
```
evilnick: :cake:

src/en/reference-constraints.md
```diff
@@ -109,33 +109,56 @@ multiple constraints delimited by a space.
...
 
-In pre-1.0 juju some additional or differently named constraints were
-also supported, these need to be migrated when upgrading.
+    Specifies the type of virtualization to be used, such as `kvm`.
 
-- cpu
 
-    Number of CPU cores for most providers, but equivalent to an Amazon
-    ECU on AWS. Use `cpu-cores` instead.
+## Provider differences
```
evilnick: can we talk about clouds rather than providers.

src/en/reference-constraints.md
```diff
@@ -109,33 +109,56 @@ multiple constraints delimited by a space.
...
-- ec2-zone
+Different providers support different constraints and sometimes different
+values for these constraints. Sometimes, different providers also dictate
+constraints that would conflict with other providers and cannot be used
+in combination. Use this list to help you understand the differing needs.
 
-    EC2 availability zone that a service unit must be deployed into. No
-    equivalent implemented as of juju 1.12, follow [bug
-    1183831](https://bugs.launchpad.net/juju-core/+bug/1183831).
+###Azure Provider:
```
evilnick: This is better than not having the information, but it still feels rather clunky to use and understand. Maybe for now it is sufficient to add a TODO to the doc and file a new issue (medium priority)

evilnick:
I think that's the end of my input here - good work on spotting the MAAS and service bits :)

matthewhelmke:
Thanks! I'll update what we have so far in the PR.

matthewhelmke:
Fixes #1249

nskaggs:
This doesn't appear to be live yet -- when will it be?

evilnick:
It is live in the /dev docs - https://jujucharms.com/docs/devel/reference-constraints

I was waiting to see if there was an idea when the cpu-core -> core change will land, but it doesn't look like it will be soon. In which case we will put a note on it and push it out to 2.0

PR #1338, opened 2016-09-06
Whoami inclusion

### LINKED ISSUE #1334: Document new whoami command

pmatulis:
Fixes #1334

### DIFF
--- a/src/en/users-creating.md
+++ b/src/en/users-creating.md
@@ -89,5 +89,5 @@ to determine the current user (as well as the current controller and model):
 juju whoami
 ```
 
-The command `juju show-user` can also be used to get information on the current
-user.
+The command `juju show-user` can also be used to get the current user, in
+addition to other information on the user.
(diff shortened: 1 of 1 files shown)


src/en/users-creating.md
```diff
@@ -83,8 +83,11 @@ juju login
...
 Once a user logs in they become the current user. The following is a quick way
-to determine the current user:
+to determine the current user (as well as the current controller and model):
 
 ```bash
-juju show-user
+juju whoami
 ```
+
+The command `juju show-user` can also be used to get information on the current
```
matthewhelmke: Would it be useful to remove the word "also" since `show-user` appears to do more than `whoami`?

pmatulis: Command `show-user` *does display* what the current user is, along with some other information on the user. So it **also** gives the current user *in addition* to other information. So maybe:

"The command `juju show-user` can also be used to get the current user, in addition to other information on the user."

matthewhelmke: Works for me.

Issue #1175, opened 2016-06-10
Add info about binding when deploying applications


matthewhelmke:
I think this belongs in charms-deploying.md. Need to study it more closely.
## Binding Service Endpoints to Spaces

Binding means the "bound" endpoints will have IP addresses from
subnets that are part of the space the endpoint is bound to.

Use the optional '--bind' argument when deploying a service to
specify to which space individual charm endpoints should be bound. The
syntax for the '--bind' argument is a whitespace-separated list of
endpoint and space names, separated by "=".

When '--bind' is not specified, all endpoints will use the same
address, which is the host machine's preferred private address, as
returned by "unit-get private-address". This is backwards-compatible behaviour.

Additionally, a service-default space can be specified by omitting
the "<endpoint>=" prefix before the space name. This space will
be used for binding all endpoints that are not explicitly specified.

Examples:

```
juju deploy mysql --bind "db=database server=internal"
```

Bind "db" endpoint to an address part of the "database" space (i.e. the
address is coming from one of the "database" space's subnets in MAAS).

```
juju deploy wordpress --bind internal-apps
```

Bind _all_ endpoints of wordpress to the "internal-apps" space.

```
juju deploy haproxy --bind "url=public internal"
```

Bind "url" to "public", and all other endpoints to "internal".
## Binding Endpoints of Services Within a Bundle to Spaces

Bindings can be specified for services within a bundle the same way as
when deploying individual charms. The bundle YAML can include a section
called "bindings", defining the map of endpoint names to space names.

Example bundle.yaml excerpt:
    ...
    mysql:
        charm: cs:trusty/mysql-53
        num_units: 1
        constraints: mem=4G
        bindings:
            server: database
            cluster: internal
    ...

Deploying a bundle including a section like in the example above, is
equivalent to running:

```
juju deploy mysql --bind "server=database cluster=internal"
```

There is currently no way to declare a service-default space for all
endpoints in a bundle's bindings section. A workaround is to list all
endpoints explicitly.

matthewhelmke:
Related.

## New Hook Command: network-get

When deploying a service with endpoint bindings specified, charm authors
can use the new "network-get" hook command to determine which address to
advertise for a given endpoint. This approach will eventually replace
"unit-get private-address" as well as various other ways to get the
address to use for a given unit.

There is currently a mandatory '--primary-address' argument to 'network-
get', which guarantees a single IP address to be returned.

Example (within a charm hook): 

    relation-ids cluster
    url:2

[...]

PR #1341, opened 2016-09-08
Added info about binding to multiple locations

### LINKED ISSUE #1163: Document new extra bindings support for charms metadata
Need to determine where this belongs.

Many charms use a simple model where a relationship with another charm
also indicates there is a network connection between units of those
services. We have utilized this fact to add a network model that allows
system operators to control where those TCP connections are made by
binding the service relation endpoints onto a network space.

However, some charms specifically use relation endpoints only as a way
to pass configuration around, and the relations don't map directly to
services that are running in that charm and/or networking configuration.
These charms want to be able to express that they have more networking
configuration that an operator wants to control without having yet-
another interface that will never be related to another service.

Juju solves the aforementioned issues by introducing an optional new
section in the charm metadata,yaml. The new section is called "extra-
bindings". Similarly to the peers/provides/requires sections, extra-
bindings contains a list of names, which can be used with 'juju deploy
--bind' like relation names. Unlike relations, you don't have to define
hooks or anything more complex to allow the users of the charm to bind
those names to Juju spaces.

Example: ceph charm's metadata.yaml would look like:

    name: ceph
    summary: Highly scalable distributed storage
    maintainer: OpenStack Charmers <openstack-charmers@lists.ubuntu.com>
    description: |
     Ceph is a distributed storage and network file system designed to provide
     excellent performance, reliability, and scalability.
    tags:
      - openstack
      - storage
      - file-servers
      - misc
    peers:
      mon:
        interface: ceph
    extra-bindings:
      cluster:
      public:
    provides:
      nrpe-external-master:
        interface: nrpe-external-master
        scope: container
    …
    storage:
      osd-devices:
        type: block
        multiple:
          range: 0-
      osd-journal:
        type: block
        multiple:
          range: 0-1

As a user then you can deploy this charm and tell Juju to provide
distinct addresses for it on one or more spaces:

    juju deploy ~/path/to/charm/ceph --bind 'cluster=admin-api public=public-api internal-api'

Units of the ceph service will be deployed on machines which have access
to the "admin-api", "public-api", and "internal-api" spaces in MAAS, a
different network interface and address for each binding.

Then, e.g. in a hook of the same charm, running 'network-get cluster
--primary-address' will only return the correct address - the one coming
from the "admin-api" space.

matthewhelmke:
Fixes #1163 and #1175 
@juju/docs

This is a complex topic with information (I believe appropriately) spread across multiple pages. I would appreciate a careful reading and review for both accuracy and clarity.

### DIFF
--- a/src/en/authors-charm-metadata.md
+++ b/src/en/authors-charm-metadata.md
@@ -119,7 +119,7 @@ resources:
     description: example resource
 ```
 
-`payloads` allows you to register payloads such as LXC, KVM, and docker with
+`payloads` allows you to register payloads such as LXD, KVM, and docker with
 Juju. This lets the operator better understand the purpose and function of these
 payloads on a given machine.
 
@@ -132,7 +132,7 @@ payloads:
 ```
 
 `extra-bindings` represents an extra bindable endpoint that is not a relation.
-These are useful whenyou want to have Juju provide distinct addresses for an
+These are useful when you want to have Juju provide distinct addresses for an
 application on one or more spaces. For example, adding this section to a YAML
 file for an application called "foo":
 
@@ -149,7 +149,7 @@ different network interface and address for each binding, using this:
 juju deploy ~/path/to/charm/foo --bind "cluster=admin-api public=public-api internal-api"
 ```
 
-And running `juju network-get cluster --primary-address` will return only the
+And running `network-get cluster --primary-address` will return only the
 address coming from the "admin-api" space.
 
 
--- a/src/en/authors-charm-metadata.md
+++ b/src/en/authors-charm-metadata.md
@@ -1,4 +1,4 @@
-Title: Charm metadata  
+Title: Charm metadata
 
 # Charm metadata
 
@@ -26,7 +26,7 @@ Here's a valid metadata file:
     description: |
       MongoDB is a high-performance, open source, schema-free document- oriented
       data store that's easy to deploy, manage and use. It's network accessible,
-      written in C++ and offers the following features:  
+      written in C++ and offers the following features:
       - Collection oriented storage
       - easy storage of object-style data
       - Full index support, including on inner objects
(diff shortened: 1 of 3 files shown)


src/en/authors-charm-metadata.md
```diff
@@ -132,12 +132,26 @@ payloads:
 ```
 
 `extra-bindings` represents an extra bindable endpoint that is not a relation.
+These are useful whenyou want to have Juju provide distinct addresses for an
```
evilnick: missing space

src/en/charms-bundles.md
```diff
@@ -223,6 +223,35 @@ other than "0", which is used to represent the bootstrap node.  Leaving the
...
+## Binding endpoints of applications within a bundle
+
+You can configure more complex networks using [spaces](./network-spaces.html)
+and deploy charms with binding, as described in [Deploying applications](./charms-deploying.html).
+Bindings can also be specified for applications within a bundle. To do so,
+add a section to the bundle's YAML file called `bindings`. For example:
+
+```
+...
+mysql:
```
evilnick: if this is a YAML fragment, I don't think it is formatted properly

src/en/charms-deploying.md
```diff
@@ -281,6 +281,46 @@ are supplied, then the remaining units will be assigned as normal to a new,
...
+
+You can configure more complex networks using [spaces](./network-spaces.html). This
+also permits the use of `--bind` to specify which network space individual
+charm endpoints should use. When you bind, the endpoint(s) of the charm will
+have IP addresses from subnets that are part of the specified space.
+
+When `--bind` is not used, all endpoints will use the host machine's preferred
+private address, which you can see using `unit-get private-address`.
+
+Create your space before deploying a charm that attempts to use it.
```
evilnick: could we link to where that is described?

evilnick:
It is probably worth asking Dimiter to have a look over this as well

matthewhelmke:
@dimitern can you please take a look at this? Thank you!

src/en/charms-deploying.md
matthewhelmke: The link to the page describing configuring/creating spaces is in the very first sentence. Did you mean something else, or would you like it linked twice?

src/en/authors-charm-metadata.md
```diff
@@ -132,12 +132,26 @@ payloads:
...
 ```
+Will permit you to deploy the charm using `--bind` to deploy on units that have
+access to the "admin-api", "public-api", and "internal-api" spaces with a'
+different network interface and address for each binding, using this:
+
+```bash
+juju deploy ~/path/to/charm/foo --bind "cluster=admin-api public=public-api internal-api"
+```
+
+And running `juju network-get cluster --primary-address` will return only the
```
dimitern: s/juju network-get/network-get/ - it's not a juju cli command, but a hook tool.

src/en/charms-bundles.md
dimitern: +1

dimitern:
@matthewhelmke  Looks good - only a couple of minor issues. I've also reviewed the full text of the 3 changed files and noticed "lxc" used in a few places. Those should instead use "lxd" - e.g. "lxd:0".

matthewhelmke:
@dimitern Thank you. I'll fix those now along with the issues @evilnick pointed out.

src/en/charms-deploying.md
matthewhelmke: I'm going to move the sentence up to the opening paragraph. It still fits there, but makes the link between the two topics more obvious.

evilnick:
Ok :+1:

PR #1343, opened 2016-09-08
review of getting-started.md

pmatulis:


src/en/getting-started.md
```diff
@@ -4,80 +4,90 @@ TODO: Mediawiki status needs new screenshot when status has been updated in Juju
...
 ```
 
-You will be prompted for various options. As an example, to configure LXD to 
-create a new 32GB ZFS pool to use, called 'lxd-pool', and set up a bridge 
-network (required for Juju), your session would look like this:
+You will be asked several questions. In the example below, LXD will i) create a
+32GB ZFS pool, ii) refrain from putting the pool on a separate block device,
+iii) refrain from listening over the network, and iv) trigger the setup of a
+bridge network (required for Juju).
+
```
matthewhelmke: I haven't seen us use the lower-case Roman numeral style before. I don't think it is bad, but since it appears to be unique I thought it should be highlighted and considered before merging.

src/en/getting-started.md
```diff
@@ -217,33 +229,29 @@ juju status
...
-the [Applications] section, which show that Mediawiki and MySQL are installed, 
-and
-the [Units] section, which crucially shows the IP addresses allocated to them.
+![juju status](./media/juju-status-wiki-simple.png)
 
-By default, Juju is secure - you won't be able to connect to any applications 
-unless they are specifically exposed. This adjusts the relevant firewall 
-controls (on any cloud, not just LXD) to allow external access. To make
-our Mediawiki visible, we run the command:
+There is lots of juicy information there! The important parts for now are
```
matthewhelmke: "There is lots of juicy information...."
* I can't recall the proper British rule; is "information" a collective noun or a plural noun? I'm asking because to my ear, the verb sounds alternately acceptable or like it should be "are". That and the word "juicy" made me stop while reading and took me out of the narrative. Maybe we want to rephrase the sentence to prevent distracting the reader from the content?

matthewhelmke:
I think this is well done! Thanks!
:cookie:

src/en/getting-started.md
pmatulis: Looks fine to me. I avoided bullets because of the real estate it requires in our theme (or whatever it's called).

src/en/getting-started.md
pmatulis: Juicy. I threw that in to see if anyone noticed. Bravo. I'll reword.

Collective nouns such as *information* can be used as singular or plural in British English, there is no preference.

evilnick:
Please do not land this

PR #1344, opened 2016-09-09
Added info about model names and multiple users

### LINKED ISSUE #1289: Model names and multiple users
We should check our multiple users section of the docs and get this added. From the Juju mailing list, from Andrew Wilkins <andrew.wilkins@canonical.com>:

Hi folks,

We've just landed a change that will be part of beta15 which changes how you refer to models owned by other users.

Model names may be reused by different users, e.g. alex and billie can each have a model called "foo". Until beta15, there is no way for either alex or billie to refer to each other's foo after being granted access.

As of beta15, you can now refer to others' models with the syntax "owner/model". Just saying "foo" is short-hand for foo within the logged-in user's namespace. e.g. If I'm alex, I can get the status of billie's model foo with "juju status -m billie/foo".

Cheers,
Andrew

--
Juju mailing list
Juju@lists.ubuntu.com

matthewhelmke:
Fixes #1289 
@juju/docs Ready for review

### DIFF
--- a/src/en/users-models.md
+++ b/src/en/users-models.md
@@ -80,3 +80,10 @@ juju add-user --models=mymodel --acl=read ben
 ```
 
 Naturally, the model in the above example needs to already exist.
+
+!!! Note: Each user has control over naming the models they own. This means
+it is possible for two users, `jane` and `claire`, to each have a model with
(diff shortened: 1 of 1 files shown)


evilnick:
:+1:

Issue #857, opened 2016-02-04
Review/convert all pages to new Juju 2.0 commands


evilnick:
The pages below need to be updated with the [new juju 2.0 commands](https://github.com/juju/docs/blob/master/src/en/command-changes.md).

Master bug: #901

  - [x] about-juju.md
  - [x] actions.md
  - [x] charms-bundles.md
  - [x] charms-config.md
  - [x] charms-constraints.md
  - [x] charms-deploying.md
  - [x] charms-destroy.md
  - [x] charms-exposing.md
  - [x] charms-ha.md
  - [x] charms.md
  - [x] charms-relations.md
  - [x] charms-scaling.md
  - [x] charms-service-groups.md
  - [x] charms-storage.md
  - [x] charms-working-with-units.md
  - [x] command-changes.md
  - [x] commands.md
  - [x] config-tagging.md
  - [x] contributing.md
  - [x] getting-started-keygen-win.md
  - [x] getting-started.md
  - [x] glossary.md
  - [x] howto-charm-with-docker.md
  - [x] howto-drupal-iis.md
  - [x] howto-harvesting.md
  - [x] howto-highavailability.md
  - [x] howto.md
  - [x] howto-node.md
  - [x] howto-privatecloud.md
  - [x] howto-rails.md
  - [x] howto-vagrant-workflow.md
  - [x] introducing-2.0.md
  - [x] juju-backups.md
  - [x] juju-block.md
  - [x] juju-gui-management.md
  - [x] juju-ha.md
  - [x] juju-managing.md
  - [x] juju-misc.md
  - [x] juju-offline-charms.md
  - [x] juju-upgrade.md
  - [x] reference-constraints.md
  - [x] reference-environment-variables.md
  - [x] reference-hook-tools.md
  - [x] reference-numbering.md
  - [x] reference-release-notes.md
  - [x] reference-releases.md
  - [x] reference-status.md
  - [x] troubleshooting-logs.md
  - [x] troubleshooting.md
  - [x] troubleshooting-upgrade.md

pmatulis:
@evilnick 
@degville 
@matthewhelmke 
This issue needs to be updated and maintained.

Note: [glossary.md](https://jujucharms.com/docs/devel/glossary) shows as being done but it is obviously out of date.

evilnick:
I think it is actually accurate (i just removed a load of pages I deleted). 
I am losing faith in these long lists.

matthewhelmke:
[...]

PR #1360, opened 2016-09-13
fix formatting of list in actions page

matthewhelmke:
Also checked page to help with #857, no changes were needed for that.


matthewhelmke:
Simple change, merging.

PR #1370, opened 2016-09-14
juju_log -> log

mistotebe:
As per [charmhelpers docs](https://pythonhosted.org/charmhelpers/api/charmhelpers.core.hookenv.html#charmhelpers.core.hookenv.log).


evilnick:
this seems 👍 to me, but I'd like to tag in @chuckbutler

evilnick:
🍰

PR #1383, opened 2016-09-19
updated page title and added links to remove models and controllers

matthewhelmke:
Several small changes:
- Removed "environments" from the page as we no longer destroy-environments.
- Added Oxford commas.
- Fixed broken link to models documentation.
- Added link to controllers documentation.
- Moved info about removing relations out of separate section because the section was far away from where the info was useful; placed the info in the Note under Removing an application which mentions the possible need to do this.


matthewhelmke:
I hadn't seen that before. I'll do all the links on the page now and resubmit soon.

matthewhelmke:
@evilnick please review. I want to make sure the link with the anchor is acceptable as formatted.

PR #1390, opened 2016-09-20
instructions were correct but screenshot was old

### LINKED ISSUE #1388: GCE credentials setup needs updating on docs
https://jujucharms.com/docs/2.0/help-google has changed a bit since the original publication. The key information can still be gleaned, but It should be updated.

matthewhelmke:
Fixes #1388 
@evilnick, just a quick check needed. The instructions are correct as written, but one screenshot showed a dropdown menu box with an old name. This replaces that image.


evilnick:
Cool, thanks 
:cookie:

PR #1395, opened 2016-09-21
Update command ref for RC1

evilnick:
@matthewhelmke here isn't much point looking at the diff, but it would be useful if you could build locally and confirm the page looks okay


matthewhelmke:
LGTM
:tada:

PR #1403, opened 2016-09-22
added username to juju login example to fit current usage

### LINKED ISSUE #1352: juju login changes should be documented
`juju login` now supports external users. If you have identity-url  configured, you must now explicitly specify a user name on the command line to log in as a local user.

When a login expires for local users, you will now be automatically  prompted, rather than getting an error back telling you to run `juju login`.

Macaroons for local users are now stored in the cookie jar, as with  external users. There is a known issue with logout (#1621375) which will be addressed in beta19.

matthewhelmke:
Fixes #1352 
In addition to the attached change, all pages referencing logging in to Juju have been checked.

### DIFF
--- a/src/en/users-creating.md
+++ b/src/en/users-creating.md
@@ -76,10 +76,11 @@ the current Juju user session:
 juju logout
 ```
 
-And to log back in, either on the same client system or not:
+And to log back in, either on the same client system or not, using the same
+user we added earlier:
 
 ```bash
-juju login
+juju login jon
 ```
 
 Once a user logs in they become the current user. The following is a quick way


evilnick:
was that all it needed? wow. easier than I thought :+1:

matthewhelmke:
Sometimes we are privileged to discover that the details are abstracted away enough to require only a simple change in the docs. It is the research and understanding that make up the submerged bit of the iceberg in those instances. :)

evilnick:
You make it sound like we wrote the docs in such a sensible way that only a minimal change was required...

matthewhelmke:
LOL, of COURSE we did. :grinning:

PR #1414, opened 2016-09-29
Release devel 2.0 rc2

nskaggs:
Note, the next update should be for 2.0 ;-)

### DIFF
--- a/src/en/reference-releases.md
+++ b/src/en/reference-releases.md
@@ -76,8 +76,8 @@ agent-stream: proposed
 
 ## Development
 
-2.0-rc1 is currently in development for testing.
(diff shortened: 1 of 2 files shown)


matthewhelmke:
Thank you!
:cake:

matthewhelmke:
@nskaggs I have merged, but also noticed this line and wonder whether "atomically" should be "automatically" or whether the original is what is intended. I kind of thought "atomically" is intended, but want to check. I'll edit, if needed.

"The upgrade-charm command now has --config and --storage flags, so you can atomically update..."

nskaggs:
@AlexisBruemmer, @axw can you answer matthew's question? ^^

PR #1415, opened 2016-09-29
Add documentation on metrics.

cmars:

PR #1419, opened 2016-10-03
backports

matthewhelmke:


matthewhelmke:
Backporting today to take care of https://github.com/juju/docs/pull/1415#issuecomment-250841584 as well as a couple small fixes here and there.

PR #1429, opened 2016-10-07
Adding private cloud info for OpenStack

evilnick:
@matthewhelmke, some light reading for you. At the moment it is untested


matthewhelmke:
Understand we don't have the ability to test this ourselves. I feel good about our info source and I like how this info is presented.

LGTM
:cake:

PR #1430, opened 2016-10-07
Expand coverage of status-set and status-get

sparkiegeek:
As a follow up to changes made for #1299, I've tried to explain why charm authors should want to use `--application`.

Drive-by fix for s/state server/controller/.

Please help word-smith as you see fit.

### DIFF
--- a/src/en/reference-hook-tools.md
+++ b/src/en/reference-hook-tools.md
@@ -616,10 +616,9 @@ status-get
 status-get --include-data
 ```
 
-Use `--application` to get the overall status for the application to which
-the unit belongs if the unit is the leader, for example:
+Use `--application` to get the overall status for the application, as set by the leader:
 ```bash
-juju run --unit ubuntu/0 'status-get --application
+status-get --application
 ```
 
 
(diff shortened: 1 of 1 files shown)


sparkiegeek:
PS The spec also covers the following aggregation rules and behaviour, didn't know if we wanted to include these somewhere (and where that place should be).

> When there is no leader, Juju will set service status based on the following aggregation rules for the unit statuses:
> - If the service is “dying” internally, then the service state is “waiting”.
> - If any unit is in error, the service is in error.
> - If any unit is blocked, the service is blocked.
> - If any unit is in maintenance, the service is in maintenance.
> - If any unit is waiting, the service is waiting.
> - If any unit is active, the service is active.
> - Otherwise, the service status is unknown.
>
> When a service is destroyed, Juju will set its status to “terminated”.
>
> If a unit charm hook crashes and there is a lead unit, Juju will call the leader’s update-status hook to allow the leader to determine whether the service is up. If there is no leader, or if the leader’s update-status hook crashes, then Juju will set the service state to “error”.

evilnick:
BTW the guidelines are in this page:
https://jujucharms.com/docs/devel/reference-status
We can probably amend this there to make it clear

PR #1442, opened 2016-10-13
Add pointer to new tutorial

evilnick:
+ change install source
+ remove older notes and warnings


evilnick:
@matthewhelmke one for you. There will be another branch shortly with more changes to the first tutorial and some nave changes

matthewhelmke:
LGTM so far, @evilnick

evilnick:
good enough for me

PR #1444, opened 2016-10-13
Tutorial for sharing clouds with new users and permissions

degville:
- ported over words from the original document
- added images


matthewhelmke:
@degville this is clear and easy to follow. I like it. I'll give @evilnick the final okey-dokey, though.

One thought on the graphics and the TODO: some of the web developer plugins for Firefox and Chrome let you change text in the UI. You might find that useful for removing your username and replacing it with "newuser". I haven't done it lately, but used to use one often for one project I wrote documentation for.

chuckbutler:
Nice bundle choice for the visuals :) 👍

PR #1448, opened 2016-10-13
bundle machine spec and placement updates

kwmonroe:
Juju 2 supports placing applications as defined in the machines specification
of the bundle. Update the docs to match this behavior.


matthewhelmke:
Very useful, thanks!
:cake:

Issue #1166, opened 2016-06-10
need bootstrap version of models-config


mitechie:
This is a great set of info for the models side, but it leaves out the fact that it matches the bootstrap experience and that the same keys can be used/etc. 

https://jujucharms.com/docs/devel/models-config

matthewhelmke:
Closed via #1450

PR #1450, opened 2016-10-13
Updates to model keys and bootstrap options

degville:
in reference to #1166 

- updated docs to account for model keys passed via bootstrap no longer being inherited by new models and 'model-defaults' now being used to serve this purpose. 
- clarified and updated missing bootstrap keys.


matthewhelmke:
LGTM
Thanks, @degville 
:beers:

PR #1466, opened 2016-10-14
Improve charm bundle bindings example

szeestraten:
Fixes the bindings code example which had broken syntax highlighting.
Also simplified the bundle example to focus on the bindings part and stay in line with earlier examples in the doc.
Finally, the `juju deploy` example was made more explicit to reflect the proper equivalency to the bundle example.


evilnick:
Thanks very much for the contribution!

PR #1480, opened 2016-10-19
Fixes 1449 - added example showing setting model config per region

### LINKED ISSUE #1449: Add examples showing setting model config per region
Set the default configuration value for all models in the controller region
for key to value and key2 to value2.

     juju model-defaults us-east-1 key=value key2=value2

matthewhelmke:
Fixes #1449 

The original issues asks for 'examples' and I only added one because it fit so smoothly in the existing set of examples for 'model-config'

As I am writing this, I realized that the issue uses 'model-defaults' and I used 'model-config' in the PR, so this requires one more change from me before review. Coming shortly.


matthewhelmke:
@juju/docs
Ready for review.

Issue #1456, opened 2016-10-14
all status output examples need to be updated


evilnick:
to reflect recent changes in the output

bac:
Yes, some fundamental commands like bootstrap have changed parameter order.

matthewhelmke:
That is more of an input issue and has already been updated in #1440. This issue is about capturing the output of commands and updating the places in the documentation where we show output. That involves checking each instance, comparing the new and old output, and updating those which do not match.

bac:
Sorry, I thought this bug referenced the generated file served at:

https://jujucharms.com/docs/2.0/commands

That page is incorrect.

matthewhelmke:
Ahh, I see. Thanks. We will look into that. :)

evilnick:
that was fixed in #1473

bac:
Thanks @evilnick

PR #1482, opened 2016-10-20
getting-started updated output

### LINKED ISSUE #1479: getting-started page has outdated list-controllers example
The list-controllers example is outdated and needs to be updated to reflect the current reality.
This will apply to all other doc pages with example tabular output.

degville:
fixes #1479 

- updated list-controllers output
- updated images

part of a wider fix needed for #1456


matthewhelmke:
Excellent! Thank you.
:birthday:

PR #1485, opened 2016-10-20
Removed step numbers from GCE tutorial, updated command output.

degville:
- removed step numbers for consistency with other tutorial-style pages
- updated command output
- added missing link
- added direction for readers from the Getting Started page

@juju/docs


matthewhelmke:
LGTM
Thanks!
:pizza:

Issue #1487, opened 2016-10-21
How to get Juju 1


anastasiamac:
As per https://bugs.launchpad.net/juju/+bug/1634405, it would be very helpful to have a doc page that outlines how to get Juju 1.

In interim, I am putting up a PR (I'll link it in the bug) that will describe the basics, something along the lines of:

    Note: 
    juju-1 command comes from juju-1-default package.
    On trusty, use 'update-alternatives'.
    On xenial/yakkety, use juju-1-default package install.
    After the installation, you should have /usr/bin/juju-1 directory to use 'juju-1' command.

PR #1491, opened 2016-10-21
Add page on installing both Juju series

evilnick:
Partially addresses #1487, but more work is needed on other OSes

### DIFF
--- a/src/en/introducing-2.md
+++ b/src/en/introducing-2.md
@@ -1,6 +1,6 @@
 Title: Introducing Juju 2.0
 TODO: Add extra changes as required
-      Test upgrade instructions
+
 
 # Introducing Juju 2.0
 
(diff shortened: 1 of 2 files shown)


evilnick:
@juju/docs @anastasiamac @nskaggs

matthewhelmke:
I think this is a very good start. The info for Ubuntu is clear and easy to follow. Once we are able to get info on other distros and OSes, this page will be even more useful.

evilnick:
I have tacit IRC approval also, so I think we should at least push this to devel

matthewhelmke:
Agreed. We can do new PRs for the other stuff.

evilnick:
though I did just notice some unbalanced parentheses...
i will fix that first :smile:

matthewhelmke:
You would think that I would remember to count parens after that Lisp class in the 1990s...but, no.

PR #1493, opened 2016-10-21
Added co-exist info for CentOS etc

evilnick:


evilnick:
@juju/docs

matthewhelmke:
LGTM
:dancer: (since you already had lunch)

PR #1500, opened 2016-10-26
Added an example of charm configurations options to bundles docs

### LINKED ISSUE #1346: Bundles page lacks information about charm configuration
The bundles page should document how to set configuration options on the deployed charms. An example of this in use is the [Kubernetes](https://jujucharms.com/kubernetes-core/) bundle.

szeestraten:
Added an example of charm configuration options to the bundles docs.
I think this should fix #1346 

**NOTE:** I could not find a standard of how linking between documents is done so please double check that to confirm and/or fix it.

### DIFF
--- a/src/en/charms-bundles.md
+++ b/src/en/charms-bundles.md
@@ -174,6 +174,29 @@ mysql:
       "gui-y": "168"
 ```
 
+## Setting charm configurations options in a bundle
+
+When deploying an application, the charm you use will often support or even
+require specific configuration options to be set. These options can be set in
+a bundle as a simple key addition to the application definition, using the
+configuration key/value pair.
+[See the documentation on application configuration](charms-config.html#discovering-application-configuration-options)
+on how to discover which options are available for the different charms.
+
+For example, to set the flavor of the MySQL charm to Percona in a bundle, the
+bundle file would have an additional `options` field with specific value:
+
+```yaml
+mysql:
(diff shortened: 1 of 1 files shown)


src/en/charms-bundles.md
```diff
@@ -174,6 +174,29 @@ mysql:
       "gui-y": "168"
 ```
 
+## Setting charm configurations options in a bundle
+
+When deploying an application, the charm you use will often support or even
+require specific configuration options to be set. These options can be set in
+a bundle as a simple key addition to the application definition, using the
+configuration key/value pair.
+[See the documentation on application configuration](charms-config.html#discovering-application-configuration-options)
```
matthewhelmke: If you can change the link line to include the ./ that would be great. Like this:
[See the documentation on application configuration](./charms-config.html#discovering-application-configuration-options)

src/en/charms-bundles.md
```diff
@@ -174,6 +174,29 @@ mysql:
...
+on how to discover which options are available for the different charms.
```
matthewhelmke: Remove "on how".

szeestraten:
Link and wording have been fixed.

matthewhelmke:
Thank you very much!
LGTM
:cake:

szeestraten:
Glad to help. Will you follow up on #1346?

PR #1501, opened 2016-10-26
Updated output and bootstrap syntax

degville:
More for #1456 

- updated output and bootstrap syntax
- first use of embedded HTML comments for output

@juju/docs


matthewhelmke:
Nice! Thank you.
:cookie:

PR #1502, opened 2016-10-26
Fix indentation on annotations examples in bundles docs

szeestraten:
Fixes the indentation level of the annotations in the bundle examples which are not consistent.


matthewhelmke:
Thank you! I appreciate your attention to detail here.
:cookie:

PR #1503, opened 2016-10-26
Fixes inconsistent linking in PR #1500

szeestraten:
This fixes the linking from #1500 to be consistent with the rest of the document.
Apologies for the mix up.


matthewhelmke:
Fantastic! Thanks.
:beer:

PR #1510, opened 2016-10-28
Further output updates 2

degville:
more updates for #1456

- Updated various output sections in actions, charms and clouds.
- Fixed some typos and broken syntax.

@juju/docs


matthewhelmke:
LGTM
:cake:

PR #1514, opened 2016-11-01
added metadata-source parameter

### LINKED ISSUE #1199: Private clouds require both metadata-source parameter for bootstrap and the image-metadata-url in environments.yaml
This is certainly true for 1.25

matthewhelmke:
Fixes #1199
@juju/docs
@anastasiamac please check also to make sure I have the details correct.


evilnick:
This looks good to me following my understanding of the issue, but I will defer to @anastasiamac

matthewhelmke:
Thanks, @anastasiamac!

PR #1522, opened 2016-11-03
More output updates

degville:
- fixed output formatting
- updated Azure doc to remove deprecated userpass credential

@juju/docs


matthewhelmke:
Good stuff! Thanks!!
:beers:

PR #1523, opened 2016-11-04
Change 'user enable' to 'enable-user'

degville:
Tiny change from 'user enable' to 'enable-user'.

@juju/docs


matthewhelmke:
LGTM, thanks!
:cookie:

PR #1524, opened 2016-11-06
Corrected juju config command syntax

deanmaniatis:
`juju get-config` command doesn't seem to be supported anymore on 2.0.1


pmatulis:
Thanks. :candy:

PR #1527, opened 2016-11-09
Update Ubuntu charm-tools instructions to use the snap.

cmars:
charm-tools deb packages are no longer supported.

@marcoceppi PTAL


matthewhelmke:
@cmars this LGTM. When we hear from @marcoceppi and get sign off, we are prepared to merge.

PR #1528, opened 2016-11-10
Updated block document for enable and disable commands

### LINKED ISSUE #1521: All block(s) commands have changed
I can't find any reference to this, but the group of commands used for restricting changes to the Juju environment have changed. These used to be 'block' and 'unblock', along with 'list-blocks/list-all-blocks' and 'remove-blocks/emove-all-blocks' but they now seem encapsulated by 'enable-command' and 'disable-command'.

Mostly within https://github.com/juju/docs/blob/master/src/en/juju-block.md

degville:
fixes #1521 

- reworked the old block document for the new enable and disable commands
- placed the command groups into a table

@juju/docs

### DIFF
--- a/src/en/juju-block.md
+++ b/src/en/juju-block.md
@@ -37,6 +37,8 @@ ERROR cannot destroy model: Check with SA before destruction.
 
 destroy-model operation has been disabled for the current model.
 To enable the command run
+
+    juju enable-command destroy-model
 ```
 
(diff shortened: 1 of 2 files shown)


matthewhelmke:
This is fantastic, the result is not only updated but MUCH clearer. Well done!
:tada:

evilnick:
thanks!

PR #1532, opened 2016-11-16
User access with grant/revoke update

### LINKED ISSUE #1526: grant and revoke syntax needs updating
in users-models.md and perhaps other documents, an access level needs to be specified (rather than defaulting to 'read'), and is now defined without '--acl='.

eg:
'juju grant bob mymodel' is now 'juju grant bob write mymodel'

degville:
- updated grant and revoke syntax
- added section on controller access
- added 'users' output
- clarified details around access control

fixes #1526 
@juju/docs

### DIFF
--- a/src/en/users-models.md
+++ b/src/en/users-models.md
@@ -122,11 +122,11 @@ use the following:
 juju revoke jim add-model
 ```
 
-After a `revoke` command has been issued, a user's access will revert to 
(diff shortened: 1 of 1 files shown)


matthewhelmke:
LGTM - along with your command updates the explanation changes make this much clearer. Well done!
:cake:

PR #1535, opened 2016-11-17
replace list-shares command with list-users and model argument.

degville:
(found the note I'd left to myself)

@juju/docs


matthewhelmke:
LGTM
:pizza:

PR #1541, opened 2016-11-21
Getting-started update for clarity after feedback

degville:
- small clarifications to fix potential ambiguity
- updated output

@juju/docs
(@evilnick may also want to look over these changes)

### DIFF
--- a/src/en/getting-started.md
+++ b/src/en/getting-started.md
@@ -4,7 +4,7 @@ Title: Getting started with Juju
 
 These instructions will get you up and running and deliver the best-possible
 experience with Juju. At the moment, that means using the latest release of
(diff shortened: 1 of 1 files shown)


src/en/getting-started.md
```diff
@@ -146,14 +148,14 @@ Once the process has completed you can check that the controller has been
...
-juju list-controllers 
+juju controllers 
 ```
 
 This will return a list of the controllers known to Juju, which at the moment is
 the one we just created:
   
 ```no-highlight
-Use --refresh to see the latest information.
+Use --refresh flag with this command to see the latest information.
```
matthewhelmke: Should that be "Use the --refresh flag..."?

matthewhelmke: Fair enough. I'll merge it.

PR #1546, opened 2016-11-28
the command is charm proof not juju charm proof

### LINKED ISSUE #1545: charm proof under "Your first charm starts here!"
The command for the charm proof is not "juju charm proof" it is just "charm proof" in
https://jujucharms.com/docs/stable/authors-charm-writing

Thanks

matthewhelmke:
Fixes #1545 
@juju/docs

I would especially like confirmation that it is only this charm command that loses the initial 'juju', as there are other instances of commands in the format 'juju charm _something_' that seem to be correct. I plead ignorance as to how to know the difference and would love to learn.

### DIFF
--- a/src/en/authors-charm-writing.md
+++ b/src/en/authors-charm-writing.md
@@ -404,7 +404,7 @@ Another part of the Charm Tools plugin is a useful lint-like tool which will
 check for errors in the files of your charm. Run it like this:
 
 ```bash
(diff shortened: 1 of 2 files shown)


evilnick:
the 'charm' command can be run as a plugin (see the plugins page), but not all of the commands are mirrored through the plugin for somewhat obscure reasons (this last part was news to me also).

Issue #1533, opened 2016-11-16
https://jujucharms.com/docs/2.0/tools-charm-tools references publish vs release


mitechie:
The commands are out of date with the tool.

matthewhelmke:
Fixed in #1557

Issue #1348, opened 2016-09-10
There is no "charm publish" command anymore


anthonydillon:
The page at https://jujucharms.com/docs/stable/authors-charm-store talks about the `charm publish` command, but that command was replaced by `charm release` in 2.2.0-0ubuntu1~ubuntu16.04.1~ppa2 which is currently in the stable juju ppa.

Moving: https://github.com/CanonicalLtd/jujucharms.com/issues/331

matthewhelmke:
This page also needs charm command updates: https://jujucharms.com/docs/stable/tools-charm-tools

matthewhelmke:
Fixed in #1557

Issue #1345, opened 2016-09-09
channels docs are out of date


mitechie:
https://jujucharms.com/docs/stable/authors-charm-store

The channels are different per:
https://blog.jujugui.org/2016/08/30/jujucharms-com-updated-with-new-channel-support/

evilnick:
The user docs for channels are already updated. This section of developer docs need to be updated too, but I'm not sure that the charm tools actually support this yet

https://github.com/juju/charmstore-client/issues/81

matthewhelmke:
Fixed in #1557

PR #1557, opened 2016-12-05
fixed command to be charm release instead of publish

### LINKED ISSUE #1555: /authors-charm-store#publishing-to-channels: publish not a valid command
http://pastebin.ubuntu.com/23570197/

I believe publish is now release.

matthewhelmke:
Will eventually fix #1555, but IMO this PR isn't ready yet.

I have fixed the command to be `juju release` instead of `juju publish`, but the context in these pages as well as other places in our documentation need to be massaged further. I would like to know whether we are completely abandoning the idea of "publishing to the Charm Store" and need to change our terminology throughout our docs to instead say "releasing to the Charm Store" or whether changing just the command is adequate (I strongly suspect and recommend a complete change).

I would appreciate comments from ecoteam members as well as from @juju/docs before pushing on with an additional draft.

### DIFF
--- a/src/en/authors-charm-store.md
+++ b/src/en/authors-charm-store.md
@@ -203,8 +203,8 @@ The author could publish foo-9 to either the stable or development channel
 as follows, showing the commands for stable and development respectively:
 
 ```
-charm publish cs:~kirk/foo-9
-charm publish cs:~kirk/foo-9 --channel development
+charm release cs:~kirk/foo-9
+charm release cs:~kirk/foo-9 --channel development
 ```
 
 After running both commands, revision 9 exists in both the stable channel
--- a/src/en/authors-charm-store.md
+++ b/src/en/authors-charm-store.md
@@ -54,9 +54,9 @@ pushed 4 times, the revision history would look like this:
 entity: 0---1---2---3
 ```
 
-Channels group entity revisions into named streams. There are currently two
-released channels: stable and development. Each channel also tracks
-history of revisions. When a revision is released to a channel, that
+Channels group entity revisions into named streams. There are currently four
+released channels: edge, beta, candidate, and stable. Each channel also
+tracks history of revisions. When a revision is released to a channel, that
 channel pointer is changed and the history for the channel is updated.
 
 Building on the previous example, when a user released revision 2 to the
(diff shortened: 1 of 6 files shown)


matthewhelmke:
In researching when this change posted to our code, it seems to have happened a while ago, but we had no way of knowing. I greatly appreciate the docs issue being filed and would love to know which releases of juju are impacted so we can make a wise decision about the need to backport docs changes to earlier juju releases.

marcoceppi:
The charm command is independent of juju versions

matthewhelmke:
So, does this also affect people writing charms for juju 1.xl?

marcoceppi:
Yes

matthewhelmke:
Thanks, @marcoceppi

matthewhelmke:
With the second commit here, I have gone ahead and changed the language surrounding making a charm available in the store from 'publishing a charm' to 'releasing a charm' to match the command name change. I think this will make it easier for users to remember.

I left paranthetical mentions of 'published'  and the 'unpublished channel' in authors-charm-store.md because I wanted to make sure that anyone doing a doc search on 'publish' could find information about 'release'. Also, since this is the historic language, I felt it gives a bridge to existing users who may find the terminology change confusing.

@juju/docs please review that and let me know if anyone has ideas for improvement

src/en/authors-charm-store.md
```diff
@@ -108,13 +108,13 @@ entity: 0---1---2---3---4---5---6
...
 scenarios where the stable channel may be pointed to a higher revision
 even though development revision is actually newer.
 
 In the following example revision 8 is development, a bug is found in the
 latest stable revision (5) so a hot fix is applied and pushed as 9.
-That revision is then published to the stable channel, like this:
+That revision is then released to the stable channel, like this:
 
 ```
                                               stable (9, 5, 3, 2)
```
evilnick: The channels mentioned here are also out of date, but I think there is an issuee already covering that

matthewhelmke: Let me see if I can find that issue and incorporate that change in here, too.

matthewhelmke: Found it, #1345. Working on that now, too.

matthewhelmke:
Fixes #1557 and #1348 and #1345.
@evilnick for one more pass, just to check the channels-related changes.

matthewhelmke:
Also fixes #1533

PR #1571, opened 2016-12-13
added info about external user access

### LINKED ISSUE #1567: Missing docs: `juju grant` with external (USSO) users
Lots of info on the topic here:
https://lists.ubuntu.com/archives/juju/2016-September/007843.html

matthewhelmke:
The process feels awkward, but I believe I have all the steps and information presented clearly and in the proper order for it to work as currently implemented.

Will fix #1567 
@juju/docs

### DIFF
--- a/src/en/users-models.md
+++ b/src/en/users-models.md
@@ -112,6 +112,50 @@ bob                   login      1 hour ago    58 minutes ago
 jim                   add-model  2016-11-14    58 minutes ago
 ```
 
+### Controller access for external users
+
+It is possible to give a user access to a controller without creating a local
+account for them. To do this, these criteria must first be met:
+
+- The user must already have an account on an external identity manager,
+  such as the Ubuntu SSO
+- The controller must have been created (bootstrapped) using the identity
+  configuration option, like here where we use the URL for the Ubuntu SSO
+  and Juju: `--config identity-url=https://api.jujucharms.com/identity`
+- The user must first log in to http://jujucharms.com at least once before
+  attempting to log in to Juju as an external user
+- The user must install Juju on the machine from which they will access the
+  controller
+
+On the controller, you grant Frances access to add models using:
+
+```bash
+juju grant frances@external addmodel
+```
+
+You can allow anyone with an Ubuntu SSO account to create models on this
+controller like this:
+
+```bash
+juju grant everyone@external addmodel
+```
+
+The external user will log in from their machine with `juju login`. They will
+be directed to the URL for the external identity provider so that they may
+log in there and then will be granted access to the controller.
+
+For the external user to create models from the controller, they must have
+credentials for that provider, for example, GCE or AWS. Any models created
+by this user will use these credentials.
+
+```bash
+juju add-model test --credential gce
+```
+
+Sharing controller information must be done directly between the controller
+owner and the external user, such as via email, and manually adding the
+controller information to the local `$HOME/.local/share/juju/controllers.yaml`.
+
 ## Revoke access rights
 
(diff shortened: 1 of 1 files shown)


tvansteenburgh:
I think after you do `juju grant tim@external <controller ip address>`, then I would do:

```
juju login
juju register <controller ip address>
```

Would be good to get @urosj to review this.

src/en/users-models.md
```diff
@@ -112,6 +112,50 @@ bob                   login      1 hour ago    58 minutes ago
...
+
+It is possible to give a user access to a controller without creating a local
+account for them. To do this, these criteria must first be met:
+
+- The user must already have an account on an external identity manager,
+  such as the Ubuntu SSO
+- The controller must have been created (bootstrapped) using the identity
+  configuration option, like here where we use the URL for the Ubuntu SSO
+  and Juju: `--config identity-url=https://api.jujucharms.com/identity`
+- The user must first log in to http://jujucharms.com at least once before
```
urosj: This is true atm, however, we'll deprecate this requirement in 2 weeks so it should be revisited later. Added a todo item on our side to ping you when we remove this requirement.

urosj:
@tvansteenburgh Yes, "juju register" is the easier way to distribute the information about the controller, but it also only works with controllers that have proper certs provided at the bootstrap time and that have valid DNS names set.

src/en/users-models.md
```diff
@@ -112,6 +112,50 @@ bob                   login      1 hour ago    58 minutes ago
...
+  attempting to log in to Juju as an external user
+- The user must install Juju on the machine from which they will access the
+  controller
+
+On the controller, you grant Frances access to add models using:
+
+```bash
+juju grant frances@external addmodel
+```
```
degville: I think @external (and @local?) might be worth a small explanation, as this reader may think this is an arbitrary example rather than a requirement.

src/en/users-models.md
```diff
@@ -112,6 +112,50 @@ bob                   login      1 hour ago    58 minutes ago
...
+You can allow anyone with an Ubuntu SSO account to create models on this
+controller like this:
+
+```bash
+juju grant everyone@external addmodel
+```
+
+Sharing controller information must be done directly between the controller
+owner and the external user, such as via email, and manually adding the
+controller information to the local `$HOME/.local/share/juju/controllers.yaml`.
```
degville: file (maybe add that this location is for Ubuntu/Linux)

src/en/users-models.md
```diff
@@ -112,6 +112,50 @@ bob                   login      1 hour ago    58 minutes ago
...
+log in there and then will be granted access to the controller.
+
+For the external user to create models from the controller, they must have
+credentials for that provider, for example, GCE or AWS. Any models created
+by this user will use these credentials.
+
+```bash
+juju add-model test --credential gce
+```
+
```
degville: Link to our Cloud credentials page?

src/en/users-models.md
matthewhelmke: Thank you for committing to let us know. That is always helpful and appreciated!

I'll take your other comment into advisement and I think the suggestion is a good one.

src/en/users-models.md
matthewhelmke: Added

src/en/users-models.md
matthewhelmke: Added.

src/en/users-models.md
matthewhelmke: Good idea. Added.

PR #1577, opened 2016-12-15
Juju 2.1 migrate updates

degville:
- added brief overview on using the new migrate command
- added migrate to model reference

@juju/docs / @evilnick

### DIFF
--- a/src/en/models-migrate.md
+++ b/src/en/models-migrate.md
@@ -1,5 +1,6 @@
 Title: Migrating models
-TODO: How are subnet and end point bindings handled?
+TODO: Needs adding to navigation
+      How are subnet and end point bindings handled?
       Check and add migration resources for different controllers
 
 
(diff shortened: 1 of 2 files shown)


src/en/models-migrate.md
```diff
@@ -0,0 +1,104 @@
...
+Migration is brilliant when updating Juju because you can first migrate a model
+to a different controller, update the original controller, then migrate and
+update the original model back without risking your deployment. 
+
+But migration is equally useful for load balancing. If a controller that's
+hosting multiple models reaches capacity, for example, you can now move the
+most intensive models to a new controller, reducing load without affecting your
+applications.
+
+For migration to work, the destination controller needs to be running on the
```
evilnick: Might be good to do this as a bulleted list of things

src/en/models-migrate.md
```diff
@@ -0,0 +1,104 @@
...
+
+```bash
+watch --color -n 1 juju status --color
+```
+
+In the status output, a 'Notes' column is appended to the model overview line
+at the top of the output. This new column will step through the following
+'migrating' states during the process:
+
+1. starting
```
evilnick: Just as a style thing, MD numbered lists should all be numbered '1.'. Makes it easier if it needs to be edited at a later date

PR #1580, opened 2016-12-19
Use 'juju thing' instead of the aliases 'juju list-thing'.

bac:


bac:
IMO, the docs should prefer the actual command not the aliases. There may be more out there but the list commands were an easy target to fix.

matthewhelmke:
Thank you, @bac. Is this change effective immediately, or as of Juju 2.1?

bac:
Thanks for landing @matthewhelmke . The command changes were effective with 2.0.

evilnick:
Thanks @bac !

@matthewhelmke, the defaults changed in the final stages of 2.0 away from the list-* version. I believe there may be an issue about this.

Also, changing commands.md doesn't do anything - if there is an error there either the generation of the page is out of date (it is 2.0 currently, not 2.0.1) or the jury help text is out of date. Either way, any change made there is lost when the page is regenerated.

PR #1581, opened 2016-12-20
Explicitly document positive and negative tags are supported

freyes:
This is a feature supported since juju 2.0

### DIFF
--- a/src/en/reference-constraints.md
+++ b/src/en/reference-constraints.md
@@ -88,8 +88,10 @@ multiple constraints delimited by a space.
 
 - tags
 
-    Comma-delimited tags assigned to the machine. Currently only
-    supported by MAAS.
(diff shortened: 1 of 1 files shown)


matthewhelmke:
I can accept this, but before I do, I wonder if it might be helpful to the reader to give a bit more information about what positive and negative tags are. What does this mean?

I'm happy to help massage the manner of expressing the information if I am able to better understand the feature.

pmatulis:
I think: A positive tag identifies a node that has such a tag. A negative tag identifies a node that does *not* have such a tag.

freyes:
@matthewhelmke what peter said, I just borrowed the text from the 'spaces' section.

matthewhelmke:
Ahh, I see that now. I'll approve and merge this and create a new issue for later to expand the definition.

Thank you, @freyes 
:cookie:

PR #1583, opened 2016-12-21
added brew install for OS X.

### LINKED ISSUE #1576: getting juju 2 from brew should be preferred
https://jujucharms.com/docs/stable/reference-releases

Juju 2.0 is now in brew and able to be installed with brew install juju-2.0 and possibly even juju? 

The instructions for OSX users should lean that way as an easier way than a link to the tarball.

degville:
fixes #1576

- added information on installing OS X Juju with brew
- re-orders OSes for better formatting of stable release, and for consistency with the other downloads

@juju/docs


matthewhelmke:
This looks good, @degville !
🎅

Issue #1570, opened 2016-12-12
close-port/ open-port no xrange in python3


mattyw:
https://jujucharms.com/docs/2.0/developer-hook-tools#close-port

https://jujucharms.com/docs/2.0/developer-hook-tools#open-port

Both sections mention looping using xrange. In python3 (which layers uses by default) there is no xrange

chuckbutler:
Looks like xrange was replaced with range for syntax purposes. In fact it was python2's range() function that was deprecated.

This should be a couple character cleanup.

matthewhelmke:
Closed by #1584

PR #1584, opened 2017-01-03
Cleanup of #1570

chuckbutler:
According to the python3 docs, it was the python2 range() function that
was deprecated and the python3 range() is equivalent to the python2
xrange function. I've removed the extra characters to reflect this
change.

I feel that python3 being the default for our recommended way to write
charms, this becomes a correct representation of an implementation.


matthewhelmke:
Looks good. Thank you.
Fixes #1570 
🍰

PR #1590, opened 2017-01-10
Added description of why external user authentication may be useful

matthewhelmke:
Based on this comment in #1571 

> I think this is ok for a bare minimum of information that is required to use this feature. What I'd suggest though is that in next PR better explanation is given to why one would want to link the controller to the external identity provider, highlight the benefits and provide alternative comparison if admin of the controller manages the users without identity provider.

I think this description can be enhanced further before we merge. @urosj do you have any suggestions? Others?

### DIFF
--- a/src/en/users-models.md
+++ b/src/en/users-models.md
@@ -115,7 +115,12 @@ jim                   add-model  2016-11-14    58 minutes ago
 ### Controller access for external users
 
 It is possible to give a user access to a controller without creating a local
-account for them. To do this, these criteria must first be met:
+account for them. Linking the controller to the external identity manager in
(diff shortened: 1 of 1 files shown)


src/en/users-models.md
```diff
@@ -115,7 +115,12 @@ jim                   add-model  2016-11-14    58 minutes ago
 ### Controller access for external users
 
 It is possible to give a user access to a controller without creating a local
-account for them. To do this, these criteria must first be met:
+account for them. Linking the controller to the external identity manager in
```
matthewhelmke: Thanks. Done.

PR #1593, opened 2017-01-12
Updates to models-migrate doc

degville:
Updated with further details and clarifications after some great feedback from @mjs.

@juju/docs

### DIFF
--- a/src/en/models-migrate.md
+++ b/src/en/models-migrate.md
@@ -42,7 +42,7 @@ configuration cache. See the '[clouds][clouds]' documentation for details on
 how to do this.
 
 While the migration process itself is robust, thanks to extensive checks before
-and during the process, we'd still recommend creating a backup of your source
+and during the process, we still recommend creating a backup of your source
 controller before performing a migration. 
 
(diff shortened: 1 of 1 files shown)


src/en/models-migrate.md
```diff
@@ -37,13 +41,9 @@ To start a migration, the target controller must be in the Juju client's local
...
 
-Although migration will pause a model's state and queue events until the model
-become reactivated, it's worth checking the model isn't deploying new
-applications or resources before migrating. You can check with the `juju status
-<model>` command. 
-
-While the migration process itself is robust, we'd also highly recommend
-creating a backup of your source controller before performing a migration. 
+While the migration process itself is robust, thanks to extensive checks before
+and during the process, we'd still recommend creating a backup of your source
```
matthewhelmke: I would prefer to be direct with "...we still recommend..." rather than the contraction and conditional "would".

degville:
thanks @matthewhelmke. I agree - I've just updated the doc.

matthewhelmke:
LGTM
🍰

PR #1597, opened 2017-01-17
added note regarding the need to use backports on trusty

### LINKED ISSUE #1595: Missing note to add backports when installing on trusty
According to https://bugs.launchpad.net/juju/+bug/1656027, the instructions to install Juju on https://jujucharms.com/docs/devel/getting-started-general need to have a special note for trusty users.
Namely, 'apt install -t trusty-backports lxd' before 'apt install juju'

matthewhelmke:
Fixes #1595 
We should wait on the answer to a follow-up question in that thread before merging, but my gut feeling is that this is correct as it is.

### DIFF
--- a/src/en/getting-started-general.md
+++ b/src/en/getting-started-general.md
@@ -25,7 +25,7 @@ Juju is currently available for Ubuntu, CentOS, MacOSX and Windows.
 
 ### Ubuntu
 
-To install Juju on Ubuntu 16.04 LTS (xenial), you simply need to grab the
+To install Juju on Ubuntu 16.04 LTS (Xenial), you simply need to grab the
 'juju' package from the PPA:
 
 ```bash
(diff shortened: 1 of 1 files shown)


matthewhelmke:
Based on the replies in #1595, this is ready for final review.
@juju/docs

src/en/getting-started-general.md
```diff
@@ -25,8 +25,8 @@ Juju is currently available for Ubuntu, CentOS, MacOSX and Windows.
 
 ### Ubuntu
 
-To install Juju, you simply need to grab the 'juju' package from the
-PPA:
+To install Juju on Ubuntu 16.04 LTS (xenial), you simply need to grab the
```
degville: Apologies for pedantry, but I think we usually capitalise 'Xenial' (at least when we mention it like this - same with Trusty below).

src/en/getting-started-general.md
```diff
@@ -36,6 +36,18 @@ sudo apt install juju
 Using the stable PPA resource guarantees you will always have access to the very latest
 stable version of Juju.
 
+To install Juju on Ubuntu 14.04 LTS (trusty), you have one additional step, to
+install LXD from the backports repository:
```
degville: Trusty?

src/en/getting-started-general.md
matthewhelmke: Excellent point. My mind must have been wandering. I have fixed both.

PR #1601, opened 2017-01-20
use interactive add-cloud as default

evilnick:


src/en/clouds-maas-manual.md
```diff
@@ -0,0 +1,68 @@
+Title: Manually adding MAAS clouds
+
+# Manually adding MAAS clouds
+
+MAAS clouds may be registeres with juju using the interactive
```
matthewhelmke: ...may be registered...

src/en/clouds-maas-manual.md
```diff
@@ -0,0 +1,68 @@
...
+`add-cloud` command. If you wish to define these clouds in a YAML
+configuration file instead, this is also possible. The YAML file
+describing a MAAS cloud is straightforward - all
+that varies is the endpoint and the name, since they use the same
+authentication method. For example, below is the file maas-clouds.yaml:
+
```
matthewhelmke: Consider, for brevity, although what you have is fine:
...`add-cloud` command. It is also possible to define these clouds using a YAML file containing a straightforward description. All that varies is the endpoint and the name, since they use the same authentication method. Here is an example:

src/en/clouds-maas-manual.md
```diff
@@ -0,0 +1,68 @@
...
+      type: maas
+      auth-types: [oauth1]
+      endpoint: http://172.18.42.10/MAAS
+   prodmaas:
+      type: maas
+      auth-types: [oauth1]
+      endpoint: http://prodmaas/MAAS
+```
+
+This defines three MAAS (region) controllers. To add a MAAS cloud from this
```
matthewhelmke: This example YAML defines...
To use this file to add a MAAS cloud to Juju, use:

src/en/clouds-maas-manual.md
```diff
@@ -0,0 +1,68 @@
...
+
+This will add both the 'prodmaas' and 'devmaas' clouds, which you can confirm
+by running:
+ 
+```bash
+juju list-clouds
+```
+
+This will list the newly added clouds, preceded with the prefix 'local:' which
+denotes that these are local clouds added by the user:
```
matthewhelmke: I don't see 'local:' in the output example.

evilnick:
@matthewhelmke that's good - it was one of the outdated files that i repurposed :) I should have looked more carefully. i will fix this up shortly

evilnick:
@matthewhelmke okay, changes made

PR #1616, opened 2017-01-26
moved howtos into foldouts

### LINKED ISSUE #770: Leadership howto
The developer-leadership page now uses an FAQ way of dealing with points of note - it is quite long though and would probably benefit from using the foldout style for the answers, to make it easier to see the list of things addressed

Also, the plural of HowTo is not HowTo's

matthewhelmke:
Fixes #770 
Simple fix. The plural mentioned in the original issue had already been fixed, here I am only putting the howto responses into foldout style, as discussed, to close out the issue.
@juju/docs

### DIFF
--- a/src/en/developer-leadership.md
+++ b/src/en/developer-leadership.md
@@ -98,7 +98,7 @@ on users of `leader-set`.
 
 ### How do I...
 
-Click on the expander to see details for each command.
+Click on the expander to see details for each.
 
 ^# ...share one-shot configuration among units?
   
--- a/src/en/developer-leadership.md
+++ b/src/en/developer-leadership.md
@@ -98,7 +98,9 @@ on users of `leader-set`.
 
 ### How do I...
 
-### ...share one-shot configuration among units?
+Click on the expander to see details for each command.
+
+^# ...share one-shot configuration among units?
 
 Assuming your own implementations of `create_settings` and `valid_settings`, you
 can use the two pseudo-python snippets below:
@@ -128,7 +130,7 @@ can use the two pseudo-python snippets below:
     exiting without error, to wait for the `leader-settings-changed` which
     should arrive soon).
 
-### ...share varying configuration among units?
+^# ...share varying configuration among units?
 
 You should be able to use the exact same constructs as above, in the same way;
 you just might want to call `set_shared_settings` in a few more places. If you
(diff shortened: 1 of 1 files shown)


src/en/developer-leadership.md
```diff
@@ -98,7 +98,9 @@ on users of `leader-set`.
 
 ### How do I...
 
-### ...share one-shot configuration among units?
+Click on the expander to see details for each.
+
+^# ...share one-shot configuration among units?
 
 Assuming your own implementations of `create_settings` and `valid_settings`, you
```
evilnick: I think the indents are off - have you checked this renders ok?

src/en/developer-leadership.md
```diff
@@ -167,7 +169,7 @@ primitives. As a charm author, this is unlikely to be the best use of your time
 -- extending the `hacluster` charm to cover your use case is likely to be more
 efficient.
 
-### ...render it likely that a long-lived process runs on one unit at a time?
+^# ...render it likely that a long-lived process runs on one unit at a time?
```
evilnick: I don't really like 'render it likely'

src/en/developer-leadership.md
matthewhelmke: It renders okay locally, probably because of the different css style when using the foldouts makes them look subordinate. Want these changed to `^###` ? I'm happy to do so.

src/en/developer-leadership.md
matthewhelmke: Agreed. I also don't like that it is followed by a note box that says:

> Note: this approach is not reliable. It may be good enough for some workloads, but don't use it unless you understand the forces in play and the worst possible consequences for your users...

Wasn't sure how deep into editing existing text I wanted to go when the issue I was addressing just asked for foldouts. However, there is much editing to be done.

src/en/developer-leadership.md
matthewhelmke: Wait...hold on. Nope. I'll fix this.

src/en/developer-leadership.md
matthewhelmke: I am changing this to ...run a long-lived process on one unit at a time.
The Note: box will do the hand waving about reliability.

matthewhelmke:
One more commit coming...I don't like what the indents did to the right margin, so I'm fixing that throughout.

PR #1617, opened 2017-01-27
moved reactive states explanation before event cycle info

### LINKED ISSUE #809: [2.0 docs] Reactive states should be explained before event cycle
The event cycle page is a very good page, the links to the reference are awesome! This page is exactly what I was looking for when I first started writing Charms. It didn't exist back then, I'm glad it does now!

However, I'm not sure about the ordering of that page. With the reactive approach, events aren't front and center anymore. The most important thing now is the handeling of reactive states. I think it would be better if this page started off talking about the reactive states and then continued talking about the hooks a lot further down the page.

matthewhelmke:
Fixes #809 
I did some massaging of text here and there to try and make the flow work in this direction.
I also fixed the formatting--we didn't need every heading on the page to be a top-level heading.
@juju/docs

### DIFF
--- a/src/en/developer-event-cycle.md
+++ b/src/en/developer-event-cycle.md
@@ -1,6 +1,74 @@
-Title: Event Cycle  
+Title: Event Cycle
 
 # Event Cycle
+
+When charming with the reactive framework, it’s possible to use
+developer-created states and boolean logic to run code when the states represent
+something meaningful to your deployment, such as when a database is connected,
+and when it is available, or when the primary workload becomes available for
+use. These are additional “events” that are abstracted from the hook event cycle
+outlined later in this page.
+
+## Handling reactive states
+
+Many states can be performed during a single hook execution, as these abstracted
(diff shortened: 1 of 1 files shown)


src/en/developer-event-cycle.md
```diff
@@ -1,14 +1,82 @@
...
+When charming with the reactive framework, it’s possible to use
+developer-created states and boolean logic to run code when the states represent
+something meaningful to your deployment, such as when a database is connected,
+and when it is available, or when the primary workload becomes available for
+use. These are additional “events” that are abstracted from the hook event cycle
+outlined later in this page.
+
+## Handling reactive states
+
+Many states can be performed during a single hook execution, as these abstracted
```
evilnick: it's not your fault, but this doesn't make much sense...

matthewhelmke: I agree with you. The page would benefit greatly from a full edit, particularly with assistance from a subject matter expert.

PR #1631, opened 2017-02-08
Prometheus

### LINKED ISSUE #1624: Instructions needed for setting up Prometheus monitoring
In Juju 2.1 each controller will expose Prometheus metrics over an HTTPS endpoint. The endpoint requires authorisation, so we need some docs on how to set this up. Below is something I wrote for the 2.1-beta5 release notes.

# Configuring Prometheus to scrape Juju Controller Metrics

Juju controllers now expose Prometheus metrics over an HTTPS endpoint. To feed these metrics into Prometheus, you will need to add a new scrape target to Prometheus. The metrics endpoint requires authorisation, so you will need to configure a user and password for Prometheus to use:

```
$ juju add-user prometheus
$ juju change-user-password prometheus
new password: <password>
type new password again: <password>
```

For the “prometheus” user to be able to access the metrics endpoint, you must grant the user access to the controller model:

```
$ juju grant prometheus read controller
```

Juju serves the metrics over HTTPS, with no option of degrading to HTTP. You can configure your Prometheus to skip validation, or you can store the controller’s CA certificate in a file for Prometheus to verify the server’s certificate against:

```
$ juju controller-config ca-cert > /path/to/juju-ca.crt
```

(NOTE: in 2.1-beta5 controller-config would print out the string in YAML format, which means it would include a YAML multi-line prefix. For 2.1-beta5 you will have to remove the YAML formatting. See: https://bugs.launchpad.net/juju/+bug/1661506.)

We can now add a scrape target to Prometheus. Modify prometheus.yml, adding the following scrape target:

```yaml
scrape_configs:
  job_name: juju
    metrics_path: /introspection/metrics
    scheme: https
    static_configs:
      targets: ['<controller-address>:17070']
    basic_auth:
      username: user-prometheus
      password: <password>
    tls_config:
      ca_file: /path/to/juju-ca.crt
```

matthewhelmke:
Fixes #1624 
Review requested from @axw as well as @juju/docs

### DIFF
--- a/src/en/howto-prometheus.md
+++ b/src/en/howto-prometheus.md
@@ -2,18 +2,15 @@ Title: Configure metrics gathering with Prometheus
 
 #  Configure metrics gathering with Prometheus
 
-
 ## Overview
 
-While it is possible to [write charms that collect metrics][developer-metrics], some
-will want the power and options available from a full monitoring solution. For
-convenience, starting with Juju 2.1 each Juju controller provides an HTTPS
-endpoint to expose [Prometheus][prometheus] metrics. To feed these metrics into
-Prometheus, you must add a new scrape target to your already installed and
-running Prometheus instance. For this use case, the only constraint on where
-Prometheus is running is that Prometheus must be able to contact the Juju
-controller's API server address/port.
-
+While it is possible to [write charms that collect metrics][developer-metrics] about
+charms, it is also possible to collect metrics about Juju itself. Starting with
+Juju 2.1 each Juju controller provides an HTTPS endpoint to expose [Prometheus][prometheus]
+metrics. To feed these metrics into Prometheus, you must add a new scrape
+target to your already installed and running Prometheus instance. For this use
+case, the only constraint on where Prometheus is running is that Prometheus
+must be able to contact the Juju controller's API server address/port.
 
 The Juju controller's metrics endpoint requires authorisation, so create a
 user and password for Prometheus to use:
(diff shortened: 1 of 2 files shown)


src/en/howto-prometheus.md
```diff
@@ -0,0 +1,68 @@
+Title: Configure metrics gathering with Prometheus
+
+#  Configure metrics gathering with Prometheus
+
+
+## Overview
+
+While it is possible to [write charms that collect metrics][developer-metrics], some
```
axw: The first two sentences give me the impression that we're not providing something that's better than the existing charm metrics. The new Prometheus integration is orthogonal. Charm metrics are metrics about charms, Prometheus metrics are metrics about Juju itself (operation of controllers and agents).

I think this should be reworded a bit. It's probably worth mentioning/cross-referencing still, but I would like to see mention the distinction made.

src/en/howto-prometheus.md
```diff
@@ -0,0 +1,68 @@
...
+Juju serves the metrics over HTTPS, with no option of degrading to HTTP. You
+can configure your Prometheus instance to skip validation, or enter this to
+store the controller’s CA certificate in a file for Prometheus to verify the
+server’s certificate against:
+
+```
+juju controller-config ca-cert > /path/to/juju-ca.crt
+```
+
+!!! Note: in 2.1-beta5 `controller-config` will print out the string in YAML
```
axw: Does this belong in the docs? Do we normally call out beta-specific things? FYI, the bug is fixed for 2.1-rc1.

src/en/howto-prometheus.md
matthewhelmke: Ahh, I misunderstood. Thank you. I'll clear this up.

src/en/howto-prometheus.md
matthewhelmke: Since it is fixed for rc1, no, this doesn't need to be here. Thanks.

matthewhelmke:
@axw thank you for the helpful review comments. I made changes following them.

matthewhelmke:
Thanks, @axw.
@evilnick or @degville want to take a quick read through for grammar/spelling/etc?

PR #1633, opened 2017-02-08
added clarification of add-user and register

### LINKED ISSUE #1621: Juju register command page improvement
From https://bugs.launchpad.net/juju/+bug/1660886, it is not always clear whether a newly added user needs to know anything else about the controller that they have been added to in order to register and login. For example, do they need to know controller IP, etc.

It would be great to clarify that the registration command that is provided once user is added is sufficient and contains all necessary information as per comment in the bug.

matthewhelmke:
Fixes #1621 
@juju/docs


matthewhelmke:
That is actually a good question that I believe I can clarify. I'll push another commit shortly.

PR #1642, opened 2017-02-09
added multiline value action example

### LINKED ISSUE #1622: Add multiline value action example
https://bugs.launchpad.net/juju/+bug/1661015 has multiline value example that may be nice to have on the action page

matthewhelmke:
Fixes #1622 
@juju/docs

### DIFF
--- a/src/en/actions.md
+++ b/src/en/actions.md
@@ -102,8 +102,8 @@ override the parameters within the file by providing them directly.
 repo: myproject
 sure: no
 ```
-With the above example `params.yaml` file, we could remove the `myproject` git repository
-with the following command:
+With the above example `params.yaml` file, we could remove the `myproject` git
+repository with the following command:
 
 ```bash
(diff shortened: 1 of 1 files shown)


src/en/actions.md
```diff
@@ -109,6 +109,29 @@ with the following command:
 juju run-action git/0 remove-repo --params=params.yaml sure=yes
 ```
 
+If you have an action that requires multiple lines, use YAML quoting to make
+sure the whitespace is not collapsed into one line, like in this example where
+`foo` is an action and the parameter `bar` is defined in the example 
+`actions.yaml`, shown just after the example:
+
```
degville: I think the above sentence could do with a slight rewording to explain that the YAML quoting is why there's both " and ' below, and to perhaps remove the multiple uses of the word 'example' (which is confusing my tiny brain).

matthewhelmke: Good thoughts. I'll commit an edit shortly.

PR #1658, opened 2017-02-15
Get started with Juju 2.1

degville:
A new doc that steps through two of the major new features in Juju 2.1 - interactive add-cloud and migrate. 

@juju/docs


src/en/getting-started-21.md
```diff
@@ -0,0 +1,168 @@
+Title: Get started with Juju 2.1
+TODO: add changes in output to model and show-machine
+
+
+# Get started with Juju 2.1
```
evilnick: What's new in 2.1

degville: I totally agree. I was originally going to called it What's new as I was worried about Google picking this up as a general getting started guide. I'll make those changes now.

PR #1667, opened 2017-02-16
2.1 LXD credentials

### LINKED ISSUE #1636: LXD credentials
In Juju 2.1, we've had to make some changes to cloud credential handling for LXD. If you're just bootstrapping and adding models on your laptop there is no change in workflow, but there is a change if you want to add models from another machine.

Users are now expected to have a "certificate" credential for creating LXD models. If you're on the LXD host, bootstrap and add-model will both auto-generate a credential as needed, assuming you have access to the LXD Unix socket.

If you intend to add LXD models from another machine (i.e. not the LXD host), you must add credentials to that other machine manually. On the LXD host, you can run `juju autoload-credentials` and select the LXD/localhost option. Then export the credential from the LXD host using `juju credentials localhost --format=yaml > localhost-credentials.yaml`, copy that to the remote host, and on the remote host run `juju add-credential localhost -f <localhost-credentials.yaml>`.

degville:
fixes #1636 

- adds details on generating/importing certificate credentials for LXD
- adds some credential clarifications
- updates add-model with changes for 2.1

@juju/docs

### DIFF
--- a/src/en/credentials.md
+++ b/src/en/credentials.md
@@ -2,12 +2,19 @@ Title: Juju credentials
 
 # Cloud credentials
 
-In order to access your cloud (with the exception of [LXD][lxd]), Juju will
-need to know how to authenticate itself. We use the term “credentials” to 
-describe the tokens or keys or secrets used - a set of credentials 
-is represented by a _credential name_ that is used to refer to those 
-credentials in subsequent commands.
+In order to access your cloud, Juju will need to know how to authenticate
+itself. We use the term “credentials” to describe the tokens or keys or secrets
+used - a set of credentials is represented by a _credential name_ that is used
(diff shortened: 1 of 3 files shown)


src/en/credentials.md
```diff
@@ -169,7 +175,7 @@ aws     bob*, carol
 google  wayne
 ```
 
-The asterisk '*' denotes the default credential, which will be used for the
+The asterisk ''' denotes the default credential, which will be used for the
 named cloud unless another is specified.
```
matthewhelmke: Is it still an asterisk, or should another word replace 'asterisk'?

degville: Ooh, good spot thanks Matthew. I think that's a vim fail.

PR #1675, opened 2017-02-22
added info about manual configuration of loop devices with LXD

### LINKED ISSUE #1665: Mention manual configuration for testing with loop devices in LXD
As of Juju 2.0, we no longer have a way to enable loop devices in the local provider, and we're not likely to bring it back soon (it's dangerous). Can we please get a small section added to replace the old docs that were removed with https://github.com/juju/docs/pull/1661.

To add loop devices to your container, you need to add devices to your container or profile.

```
...
devices:
  loop-control:
    major: "10"
    minor: "237"
    path: /dev/loop-control
    type: unix-char
  loop0:
    major: "7"
    minor: "0"
    path: /dev/loop0
    type: unix-block
  loop1:
    major: "7"
    minor: "1"
    path: /dev/loop1
    type: unix-block
...
  loop9:
    major: "7"
    minor: "9"
    path: /dev/loop9
    type: unix-block
```

The above is enough to expose the loop devices into the container, and for the container to acquire one of them using "losetup". It is not enough to enable the container to mount filesystems on the loop devices. For that, the simplest thing to do is to make the container privileged:

```
config:
  security.privileged: "true"
```

Each container uses the default profile, but also a model-specific profile with the name juju-<model-name>. Editing a profile will affect all of the containers using it, so you can add loop devices to all LXD containers by editing the default profile, or you can scope it to a model.

matthewhelmke:
Fixes #1665 
@juju/docs

### DIFF
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -290,8 +290,53 @@ juju upgrade-charm postgresql --storage pgdata=10G
 
 ### Known limitations
 
-- Currently LXD (localhost) does not support mounting loopback devices 
-  for storage.
+LXD (localhost) does not officially support mounting loopback devices for
+storage. However, with some configuration you can make this work.
+
+Each container uses the default profile, but also uses a model-specific profile
+with the name juju-<model-name>. Editing a profile will affect all of the
+containers using it, so you can add loop devices to all LXD containers by
+editing the default profile, or you can scope it to a model.
+
+To add loop devices to your container, add loop device entries to the default
+or model-specific profile, like this:
+
+
+```bash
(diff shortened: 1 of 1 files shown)


matthewhelmke:
That is a good idea, @degville. What do you think of keeping it under `known limitations` but adding `LXD loop devices` as a subheading?

degville:
I think that would work too. I feel it's slightly better to avoid
mentioning limitations as it's a little negative. But it's a tiny
difference. Happy either way.

matthewhelmke:
I accept that reasoning. :^)

PR #1680, opened 2017-02-23
21 networking update

### LINKED ISSUE #1638: Add 2.1 network changes info
From juju-dev lists
--------------

As part of our goal of giving operators and charms explicit mechanisms to specify what they want wrt networking and be able to give reliable operations, we've made a few changes in 2.1 that push us closer to the goal of reliable and repeatable operations.

Most of the visible changes in 2.1 are around how we handle machines that have more than one network interface. As a quick background, a network "space" in Juju is a collection of subnets that are intended to be "interchangeable" (eg have the same firewall rules, etc). The idea is that as an Operator, you can tell us "put my database into the 'internal-only' space", and Juju will do the work to find a machine that has access to one of those subnets, and tell the charm about the network configuration for the specific machine that it ends up on.

Most of the changes in 2.1 are around how we deal with applications that are deployed into containers, and how that interacts with spaces. The key points I believe are:
We are now stricter about requiring operators to inform us of what spaces they want the application to be in. If we are about to provision a container on a machine that has more than one space, but we do not have information about what spaces that container needs to be in, we will treat it as a provisioning failure, rather than guessing.
(in 2.0, any container we created would be have a network interface for every interface on the host, thus it was always in all spaces, however that still lead to bugs where we weren't sure which one to actually tell the application to use, because the operator didn't specify, and Juju was left to guess.)
In 2.0, we also would create a bridge for every network interface, even if you never intended to create a container connected to that space. For 2.1, we now wait to create the bridge until we see a request for a container that is using that space. That should ultimately create a better experience for operators. Charms that are deployed to bare metal no longer have the overhead of going via a bridge and a network card that had to be put into promiscuous mode (in case you add a container later).
For 2.2, I'd like to see us move this a step further, where any application that would end up on a machine with multiple spaces, but does not have a binding indicating how it should be configured would end up as a provisioning error, instead of Juju guessing. 
Bindings are supported on a per-endpoint bases (you can configure mysql so that the 'admin' endpoint is accessed on a separate subnet than 'db'.) However, charms will need to be updated to support that level of flexibility. The existing "unit-get private-address" doesn't have the context for us to know which binding to give it, so they will need to be updated to use "network-get --primary-address BINDING" (eg network-get --primary-address db).
For Operators, they can specify the binding of every endpoint with:
juju deploy --bind "admin=admin-space db=internal-space"
As a short-cut for the common case of wanting everything in the same space, you can just specify:
juju deploy --bind space
In bundles you can also specify bindings with a bindings stanza. eg:
    mysql:
      charm: "./xenial/percona-cluster"
      num_units: 1
      bindings:
        access: internal-api
        shared-db: internal-api

There is also a shortcut in bundles by doing:
    mysql:
      charm: "./xenial/percona-cluster"
      num_units: 1
      bindings:
        "": internal-api

Ultimately we'd probably like a better syntax around a default binding, 
but this syntax is compatible with 2.0. The real difference is we're more likely toenforce correct values in 2.1. 
So while a bundle may need updating, after updating it should be compatible with 2.0 and 2.1
There are some aspects that we were not able to get to in time for 2.1. Our ultimate goal is to provide these primatives across allproviders. Some things are likely to be small enough to make it into 2.1.1 (We're hoping to address several bugs with how spaces interact with bundles and charms on AWS,vSphere and manually provisioned machines.) Others are more likely to end up as 2.2 items.

degville:
fixes #1638 

- integrated updates for spaces and networking changes introduced with juju 2.1.
- fixed old and outdated commands
- revised and reformatted much of the text
- further clarifications and details
- spaces/commands tested

I think the title of 'How to configure more complex networks using spaces' should be changed to 'Complex networks and spaces'

@juju/docs

### DIFF
--- a/src/en/charms-bundles.md
+++ b/src/en/charms-bundles.md
@@ -321,9 +321,35 @@ This is the equivalent of deploying with:
 juju deploy cs:precise/mysql-27 --bind "server=database cluster=internal"
 ```
 
+Prior to Juju 2.1, all deployed machines were regarded as potential hosts for
+containers, and as a result, all network interfaces connected to those
+machines were bridged by default. Subsequently, bundles created before the
+release of Juju 2.1 might have assumed that a container has access to all of
+the same spaces as the host machine and will no longer deploy cleanly. 
+
+These bundles will need to be updated to be more specific about the
+bindings required. This allows the operator to specify exactly which charm
+defined endpoints should end up in specific places:
+
+```bash
+juju deploy mysql --bind "db:db-space db-admin:admin-space default-space"
+```
+
+Using the `binding` section in the bundle's YAML file, the above deploy
+commands can be mirrored in bundle format with the following:
+
+```yaml
+mysql:
+   charm: "cs:mysql"
+   num_units: 1
+   bindings:
+     "": default-space
+     db: db-space
+     db-admin: admin-space
+```
+
 It is not currently possible to declare a default space in the bundle for all
 application endpoints. The workaround is to list all endpoints explicitly.
-
 
 ## Sharing your Bundle with the Community
 
--- a/src/en/charms-bundles.md
+++ b/src/en/charms-bundles.md
@@ -328,8 +328,11 @@ release of Juju 2.1 might have assumed that a container has access to all of
 the same spaces as the host machine and may no longer deploy cleanly with Juju.
 
 These bundles will need to be updated to be more specific about the
-bindings required. This allows the operator to specify exactly which charm
-defined endpoints should end up in specific places:
+bindings required, allowing the operator to specify exactly which charm
+defined endpoints should end up in specific places.
+
+The following `deploy` command connects charm endpoints to specific spaces and
+includes a default option for any interfaces not specified:
 
 ```bash
 juju deploy mysql --bind "db:db-space db-admin:admin-space default-space"
--- a/src/en/charms-bundles.md
+++ b/src/en/charms-bundles.md
@@ -322,14 +322,15 @@ juju deploy cs:precise/mysql-27 --bind "server=database cluster=internal"
 ```
 
 Prior to Juju 2.1, all deployed machines were regarded as potential hosts for
-containers, and as a result, all network interfaces connected to those
-machines were bridged by default. Subsequently, bundles created before the
-release of Juju 2.1 might have assumed that a container has access to all of
-the same spaces as the host machine and may no longer deploy cleanly with Juju.
+containers, and as a result, all network interfaces connected to those machines
+were bridged by default. Consequently, bundles created before the release of
+Juju 2.1 might assume that a container has access to all of the same spaces as
+the host machine and may no longer deploy cleanly with more recent versions of
+Juju.
 
 These bundles will need to be updated to be more specific about the
-bindings required, allowing the operator to specify exactly which charm
-defined endpoints should end up in specific places.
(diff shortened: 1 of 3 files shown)


src/en/charms-bundles.md
```diff
@@ -321,10 +321,39 @@ This is the equivalent of deploying with:
 juju deploy cs:precise/mysql-27 --bind "server=database cluster=internal"
 ```
 
+Prior to Juju 2.1, all deployed machines were regarded as potential hosts for
+containers, and as a result, all network interfaces connected to those
+machines were bridged by default. Subsequently, bundles created before the
```
evilnick: Shouldn't that be 'Consequently' rather than 'subsequently'?

src/en/charms-deploying.md
```diff
@@ -281,43 +281,63 @@ are supplied, then the remaining units will be assigned as normal to a new,
...
-bindable endpoint that is not a relation, see [Charm metadata](./authors-charm-metadata.html).
+```
+--constraints spaces=db-space,^storage,^dmz,internal
+```
+
+Juju will provision instances connected to (with IP addresses on) one of the
+subnets of both db-space and internal spaces, and NOT connected to either the storage
+or dmz spaces.
+
+For more information regarding constraints in general, see "juju help
```
evilnick: juju help topics are deprecated - but there is a reference page on constraints: ./reference-constraints

src/en/network-spaces.md
```diff
@@ -1,35 +1,42 @@
 Title: How to configure more complex networks using spaces
+TODO: Could do with a re-write and further simplification
```
evilnick: could you add "Diagram needed" to the TODO

src/en/network-spaces.md
```diff
@@ -1,35 +1,42 @@
...
-
-As an example, consider an environment divided into three segments with
+Juju models networks using "spaces". A space is made up of one or more routable
+subnets with common ingress and egress rules. The Juju operator can model this
+topology in such a way that applications gain the required network connectivity
+without generating complex network IP maps that are not portable. This gives
+the operator much better and finer-grained control over all networking aspects
+of a model and its application deployments. 
+
+Spaces represent sets of disjoint subnets that are available for running cloud
```
evilnick: is 'disjoint' the word you meant to use here?

src/en/network-spaces.md
```diff
@@ -1,35 +1,42 @@
...
+instances that may span one or more availability zones ("zones") and there are
```
evilnick: I would lose the 'and' and start a new sentence

src/en/charms-bundles.md
degville: yep

src/en/charms-deploying.md
degville: I'll change this. It's from the old version.

src/en/network-spaces.md
degville: yep.

src/en/network-spaces.md
degville: This is from the original version, but you're right, I'll just remove it.

src/en/network-spaces.md
degville: will do - thanks!

evilnick:
I think this is mergeable if you have finished tweaking

degville:
@evilnick I have - I'll leave it alone now. Thanks!

PR #1684, opened 2017-02-23
removed 2.1 from latest release URLs

degville:
@juju/docs


evilnick:
Also, i remembered! This is a bug in the markdown parser! i think there is an issue about it

PR #1685, opened 2017-02-23
further fixes for URLs in 2.1 release notes

degville:
@juju/docs


degville:
Sorry about this, but it's still broken. I'll change them all to relative links and we can perhaps look into what's happening. Looks to me like the process is simply inserting <version> after /docs regardless.

PR #1688, opened 2017-02-24
added info about the new remove-storage command

### LINKED ISSUE #1687: Add 'remove-storage' command
From: https://github.com/juju/juju/pull/7032

## Description of change

Add the remove-storage command. This command takes
one or more storage instance, filesystem, or volume
IDs, and destroys the corresponding entities. If
a storage instance is specified (e.g. "pgdata/0"),
then the associated volume/filesystem will also be
destroyed.

To disambiguate volume and filesystem IDs, the
user must specify those IDs prefixed with the
entity type, e.g. filesystem-0, volume-0.

## QA steps

juju bootstrap aws
juju deploy postgresql --storage pgdata=ebs (wait for unit to become idle)
juju remove-storage pgdata/0
observe "pgdata-storage-detaching" hook run on postgresql/0
observe pgdata/0 is removed
observe volume 0 is detached, destroyed and removed

## Documentation changes

Yes, docs need the new remove-storage command.

matthewhelmke:
Fixes #1687 
@juju/docs

### DIFF
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -242,11 +242,11 @@ juju deploy postgresql --storage pgdata=ebs,10G cache=ebs-ssd
 
 If you remove a unit that has storage, the storage is removed with the unit.
 
-The long term plan is to enable storage to be managed separately from units,
-for types of storage that can be persistent, such as EBS volumes, enabling the
-storage can be reused by new units. We have not fully implemented this ability.
-One of the first steps in managing the storage life cycle separately is
-allowing you to remove existing storage, which you can do today.
+The long term plan is to enable storage to be managed separately from units
+for types of storage that can be persistent, such as EBS volumes. This will
+enable the storage to be reused by new units. We have not fully implemented
+this ability. One of the first steps in managing the storage life cycle
+separately is allowing you to remove existing storage, which you can do today.
 
 When you remove the storage attached to a running charm, `remove-storage` will
 run the storage-detaching hook, detach the storage from the machine, and remove
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -256,11 +256,11 @@ can be found in the output of `juju storage`, shown here:
 
 ```bash
 [Storage]
-Unit          Id        Location     Status    Message
-postgresql/0  pgdata/0  /srv/pgdata  attached 
+Unit          Id        Type        Pool  Provider id                                       Size   Status    Message
+postgresql/0  pgdata/0  filesystem  gce   us-east1-b--78e14381-d247-48d7-8273-c3c951a079d9  10GiB  attached
 ```
 
-To remove this storage by specifying the unit storage instance, use:
+To remove this storage, use:
 
 ```bash
 juju remove-storage pgdata/0
@@ -274,6 +274,8 @@ juju remove-storage pgdata/0 pgdata/1
 ```
 
 If you remove a unit that has storage, the storage is removed with the unit.
+Future versions of Juju will cause the storage to be detached but remain in
+the model.
 
 ### Upgrading with storage constraints
 
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -240,8 +240,21 @@ juju deploy postgresql --storage pgdata=ebs,10G cache=ebs-ssd
 
 ### Remove storage
 
-Specify one or more unit storage, volume, or filesystem IDs, which can be found
-in the output of `juju storage`, shown here:
+If you remove a unit that has storage, the storage is removed with the unit.
+
+The long term plan is to enable storage to be managed separately from units,
+for types of storage that can be persistent, such as EBS volumes, enabling the
+storage can be reused by new units. We have not fully implemented this ability.
+One of the first steps in managing the storage life cycle separately is
+allowing you to remove existing storage, which you can do today.
+
+When you remove the storage attached to a running charm, `remove-storage` will
+run the storage-detaching hook, detach the storage from the machine, and remove
+the storage from the model, destroying the cloud storage resources in the
+process.
+
+To remove existing storage, you must specify one or more unit storage, volume, or
+filesystem IDs; these can be found in the output of `juju storage`, shown here:
 
 ```bash
 [Storage]
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -240,16 +240,14 @@ juju deploy postgresql --storage pgdata=ebs,10G cache=ebs-ssd
 
 ### Remove storage
 
-If you remove a unit that has storage, the storage is removed with the unit.
-
-The long term plan is to enable storage to be managed separately from units
-for types of storage that can be persistent, such as EBS volumes. This will
-enable the storage to be reused by new units. We have not fully implemented
-this ability. One of the first steps in managing the storage life cycle
-separately is allowing you to remove existing storage, which you can do today.
+Future versions of Juju will manage storage entirely separately from units.
+This will enable types of storage which can persist (e.g. EBS volumes) to be
+re-used by new units or workloads. This is not fully implemented in the current
+version of Juju, but one of the first steps is to be able to remove storage,
+which is currently possible.
 
 When you remove the storage attached to a running charm, `remove-storage` will
-run the storage-detaching hook, detach the storage from the machine, and remove
+run the [storage-detaching hook][storagedetatching], detach the storage from the machine, and remove
 the storage from the model, destroying the cloud storage resources in the
 process.
 
@@ -284,6 +282,8 @@ at the same time, you can refer to more than one in the command, like this:
 ```bash
 juju remove-storage filesystem-2/3
 ```
+
+If you remove a unit that has storage, the storage is removed with the unit.
 
 ### Upgrading with storage constraints
 
@@ -359,4 +359,5 @@ the storage feature read
 [writing charms that use storage](./developer-storage.html).
 
 [model-config]: ./models-config.html#list-of-model-keys
+[storagedetatching]: ./reference-charm-hooks#[name]-storage-detatching
 [upgrade-charm]: ./commands.html#upgrade-charm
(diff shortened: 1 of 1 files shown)


src/en/charms-storage.md
```diff
@@ -18,6 +18,7 @@ juju add-storage
 juju show-storage
```
evilnick: isn't this just 'juju storage' now?

src/en/charms-storage.md
```diff
@@ -304,6 +305,15 @@ config:
   security.privileged: "true"
 ```
 
+### Remove storage
```
evilnick: I think this should go further up, after the 'Deploying...' section

src/en/charms-storage.md
```diff
@@ -304,6 +305,15 @@ config:
...
+
+Specify one or more unit storage, volume, or filesystem IDs, which can be found
+in the output of `juju storage`, like this:
+
+```bash
+juju remove-storage pgdata/0
```
evilnick: It would be good to provide more examples (e.g. one of each type if possible). E.g. demonstrate the use of 'show storage' with some output, then examples below of removing the storage based on different types of id.

src/en/charms-storage.md
```diff
@@ -304,6 +305,15 @@ config:
...
+```
+
```
evilnick: Are there any caveats for this? What happens if I remove storage from a running charm? Does the storage automatically get removed if I remove the charm that's using it?

src/en/charms-storage.md
matthewhelmke: Doh! Didn't catch that. Thanks!

src/en/charms-storage.md
matthewhelmke: Okay

src/en/charms-storage.md
matthewhelmke: That's a good idea!

src/en/charms-storage.md
matthewhelmke: Hmm...I'll see what I can find out.

matthewhelmke: Paging @axw. When you have a few minutes, can you help me answer the questions @evilnick poses above? I think anticipating and answering these up front in our docs will help our users.

axw: - If the storage is attached to a charm, then we'll:
  - run the storage-detaching hook
  - detach the storage from the machine
  - remove the storage from the model (destroying the cloud storage resources in the process)
- If you remove a unit that has storage, nothing happens *yet*. Per [the spec](https://docs.google.com/document/d/1rl5NP8qWDD-wYgqmV-MlKPfvyL0I6BzbMZKf7n-tspw/edit#heading=h.fw0oe8lv4893), what we intend is for the storage to be detached (storage-detaching hook fires, and storage detaches from machine), but remain in the model. You'll then be able to redeploy the application with that storage, or remove the storage with this command.

matthewhelmke: Thank you

matthewhelmke:
@evilnick, see what you think of the latest edits and whether they want another round of enhancement.

src/en/charms-storage.md
```diff
@@ -237,6 +238,53 @@ default values:
 juju deploy postgresql --storage pgdata=ebs,10G cache=ebs-ssd
 ```
 
+### Remove storage
+
+If you remove a unit that has storage, the storage is removed with the unit.
```
evilnick: I would put this as a note at the end, and let the following paragraph be the intro.
Also - is this actually what happens currently?

src/en/charms-storage.md
```diff
@@ -237,6 +238,53 @@ default values:
...
+```bash
+juju remove-storage volume-0
+juju remove-storage filesystem-0
+```
+
+If you have and want to remove the storage from multiple volumes or filesystems
+at the same time, you can refer to more than one in the command, like this:
+
+```bash
+juju remove-storage filesystem-2/3
```
evilnick: this example doesn't really explain the mechanics. Maybe you need more than one. 
e.g. I assume :

      juju remove-storage filesystem-0/2 

removes 0 AND 2, but maybe it removes 0,1 and 2.

src/en/charms-storage.md
```diff
@@ -237,6 +238,53 @@ default values:
 juju deploy postgresql --storage pgdata=ebs,10G cache=ebs-ssd
 ```
 
+### Remove storage
+
+If you remove a unit that has storage, the storage is removed with the unit.
+
+The long term plan is to enable storage to be managed separately from units
```
evilnick: I would suggest we rephrase this in terms of Juju. E.g. 

Future versions of Juju will manage storage entirely separately from units. This will enable types of storage which can persist (e.g. EBS volumes) to be re-used by new units or workloads. This is not fully implemented in the current version of Juju, but one of the first steps is to be able to remove storage, which _is_ currently possible.

src/en/charms-storage.md
```diff
@@ -237,6 +238,53 @@ default values:
...
+for types of storage that can be persistent, such as EBS volumes. This will
+enable the storage to be reused by new units. We have not fully implemented
+this ability. One of the first steps in managing the storage life cycle
+separately is allowing you to remove existing storage, which you can do today.
+
+When you remove the storage attached to a running charm, `remove-storage` will
+run the storage-detaching hook, detach the storage from the machine, and remove
```
evilnick: `storage-detaching`.  Might be worth a link to the hooks page ./reference-charm-hooks#[name]-storage-detaching

src/en/charms-storage.md
matthewhelmke: As I understand it, yes, this is what happens now. I _think_ what @axw was saying above in the "if you remove a unit that has storage, nothing happens _yet_" is that the long-term persistent storage story has not yet been implemented, so the new stuff (storage remaining in the model even through the unit is destroyed) doesn't yet apply.

src/en/charms-storage.md
matthewhelmke: I like that.

src/en/charms-storage.md
matthewhelmke: Yes.

src/en/charms-storage.md
matthewhelmke: Ooh, great point. The source info I used doesn't cover this. @axw, is @evilnick assuming the correct syntax here for non-sequential listed storage filesystems, and does that remove BOTH or ALL from the first to the last number?

axw: filesystem-0/2 is the ID of a single filesystem. It's not a range.

axw: I'm toying with an alternative design which would make the "filesystem-" "volume-" prefix unnecessary. Please hold off for now, until the design is finalised.

matthewhelmke: Okay. We will keep this as it is, unmerged, until we hear from you, @axw. Thanks!

matthewhelmke: @axw, please review what I have here when you get time. I have updated based on our email conversation. FYI, as I tested, the output of 'juju status' in Juju 2.1.1 on my test machine is unchanged, so I did not edit that part.

As 'detach-storage' and 'attach-storage' are not quite complete, I think the note in the Remove Storage section is sufficient for today and I will create a new issue for the other two commands.

matthewhelmke:
@axw, please review what I have here when you get time. I have updated based on our email conversation. FYI, as I tested, the output of 'juju status' in Juju 2.1.1 on my test machine is unchanged, so I did not edit that part.

As 'detach-storage' and 'attach-storage' are not quite complete, I think the note in the Remove Storage section is sufficient for today and I have created issue #1716 for the other two commands.

axw:
> FYI, the output of juju status in Juju 2.1.1 on my test machine is unchanged, so I did not edit that part.

The change is on the develop branch. I'll deploy postgresql to my controller and share the new output inline.

src/en/charms-storage.md
```diff
@@ -237,6 +238,43 @@ default values:
...
+When you remove the storage attached to a running charm, `remove-storage` will
+run the [storage-detaching hook][storagedetatching], detach the storage from the machine, and remove
+the storage from the model, destroying the cloud storage resources in the
+process.
+
+To remove existing storage, you must specify one or more storage IDs; these
+can be found in the output of `juju storage`, shown here:
+
+```bash
+[Storage]
```
axw: Here's updated output from the develop branch, deployed to GCE:

```
[Storage]
Unit          Id        Type        Pool  Provider id                                       Size   Status    Message
postgresql/0  pgdata/0  filesystem  gce   us-east1-b--78e14381-d247-48d7-8273-c3c951a079d9  10GiB  attached  
```

src/en/charms-storage.md
```diff
@@ -237,6 +238,43 @@ default values:
...
+Unit          Id        Location     Status    Message
+postgresql/0  pgdata/0  /srv/pgdata  attached 
+```
+
+To remove this storage by specifying the unit storage instance, use:
```
axw: "by specifying the unit storage instance" doesn't seem to be adding anything to this sentence?

src/en/charms-storage.md
```diff
@@ -237,6 +238,43 @@ default values:
...
+```
+
+If you have and want to remove multiple storage instances at the same time,
+you can refer to more than one in the command, like this:
+
+```bash
+juju remove-storage pgdata/0 pgdata/1
+```
+
+If you remove a unit that has storage, the storage is removed with the unit.
```
axw: Maybe add a comment saying that in a future version of Juju, removing the unit or application will cause the storage to be detached, but remain in the model?

anastasiamac:
@matthewhelmke , @evilnick 
As @axw mentioned, this is currently on develop branch. It will also be behind a development flag on 2.2GA as an experimental feature under development.

src/en/charms-storage.md
matthewhelmke: Ahh, good point. That was a holdover from when we had multiple options available. Now that this is the only option, I'll remove the extra words.

matthewhelmke:
@evilnick for review

matthewhelmke:
@evilnick bump

Issue #1704, opened 2017-03-09
charm tools snap install


jameinel:
https://jujucharms.com/docs/2.1/tools-charm-tools

Currently says "sudo snap install charm"
but you actually have to "sudo snap install charm --classic"

PR #1705, opened 2017-03-09
Update tools-charm-tools.md

jameinel:
Fix the install string for the snap install.

The only other issue I ran into is that "--classic" is only supported if your snapd is new enough. It has been backported, but I did have to do "sudo apt upgrade" to get a new enough snapd. Probably we don't have to worry about that.


matthewhelmke:
Fixes #1704

PR #1719, opened 2017-03-16
typo in folding syntax in release notes

degville:
@juju/docs


matthewhelmke:
I was under the impression we were intentionally leaving the most recent entry unfolded. I fixed this a while ago and undid the fix after noticing the 1.x series also has the first entry in the release notes page unfolded, making me certain it must have been intentional. @evilnick??

evilnick:
No, they aren't intentionally broken

PR #1726, opened 2017-03-17
added a note about juju using a self-signed ssl cert

### LINKED ISSUE #1697: "juju gui" users must be informed about self-signed certificates
See https://github.com/juju/juju-gui/issues/2553

matthewhelmke:
Fixes #1697 
@juju/docs


matthewhelmke:
[One of the comments here](https://github.com/juju/juju-gui/issues/2553) says that it happens "when Juju is bootstrapped with a self-signed certificate..." I can't recall any way to bootstrap Juju using a CA-signed cert, and without knowing of one, I think it must always be the case.

PR #1728, opened 2017-03-17
Backporting #1726 from master into 2.1

matthewhelmke:
added a note about juju using a self-signed ssl cert


matthewhelmke:
@evilnick, please give this a quick look to make sure I did the cherry-pick correctly before we merge.

PR #1735, opened 2017-03-22
added links to help users with adding creds

matthewhelmke:
@evilnick, is this what you were looking for?


matthewhelmke:
I'll do that next.

Issue #1734, opened 2017-03-22
docs templates for local rendering need updating


evilnick:

evilnick:
I'm calling this done

PR #1737, opened 2017-03-22
add styles to local rendering

evilnick:
this is work towards #1734 
It does work, but:

 - there is unnecessary cruft that can be cut out
 - the navigation is unusable really(all the targets are closed)

To test:

```
git clone -b local https://github.com/evilnick/docs
cd docs
make && make serve
```

Then point a browser at http://localhost:8000/en/about-juju.html


evilnick:
The navigation issue will require a change to the mdbuild tool. i will have a look at some point, but  not today

matthewhelmke:
Noting for completeness: In addition to what you have already mentioned, I notice that you haven't yet had a chance to add in styling around images to separate them from the background. I couldn't find a good example of an admonition, but I'm guessing those are also not yet in here.

Good start!

evilnick:
No - the styles we have here have no bearing at all on the rendered docs - it is purely for offline generation. If I add a style here, it still looks the same on jujucharms.com

The juju docs already have admonitions!

PR #1739, opened 2017-03-23
Update local builds to match reality

evilnick:
Pulls in a load of new files and some tweaks to replicate the nav hacks in jujucharms.com

A lot of this is probably not needed, but easier to get it working and then prune out the cruft

fixes #1737


matthewhelmke:
LGTM
🍻

evilnick:
ta

PR #1749, opened 2017-03-24
juju login needs a -u flag

### LINKED ISSUE #1747: Need to update docs for change in Juju login
See: https://github.com/juju/juju/pull/7092

matthewhelmke:
Will fix #1747 

I've made the obvious change to add `-u` to `juju login <user>` examples.

The commands.md page will be autogenerated after release by extracting help text, so any updates made to the help for `juju login` will be reflected in that automatically.

What I am unsure about is what does `juju login` do if we do not supply a username. Also, I'm not certain whether the users-models.md page needs an update.

@rogpeppe would you please review the small changes I've already made and then help me with the questions here?

### DIFF
--- a/src/en/users-creating.md
+++ b/src/en/users-creating.md
@@ -82,7 +82,7 @@ And to log back in, either on the same client system or not, using the same
 user we added earlier:
 
 ```bash
-juju login jon
+juju login -u jon
 ```
 
 Once a user logs in they become the current user. The following is a quick way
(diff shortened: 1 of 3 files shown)


rogpeppe:
The changes look fine to me.

> What I am unsure about is what does juju login do if we do not supply a username. Also, I'm not certain whether the users-models.md page needs an update.

If you don't provide any arguments, it ensures that you're logged into the current controller (it's a no-op when you already are).

If you provide an argument, it logs you into a public controller by contacting the controller named by the argument (this can be a DNS name or a known controller alias) and adding it as a known controller locally. This functionality was previously provided by a variant of the "juju register" command, but that behaviour will be deprecated in the future (it's possible that "juju register" itself might go away entirely).

matthewhelmke:
Thank you, @rogpeppe !

matthewhelmke:
@evilnick ready for review

PR #1753, opened 2017-03-27
add new controller config keys page

### LINKED ISSUE #1699: cannot find a controller config keys reference doc
Moved from https://github.com/CanonicalLtd/jujucharms.com/issues/419

I'm not finding a definitive reference for all valid config keys to pass as "--config myfile.yaml" when bootstrapping a controller. The bootstrap help text shows a few usage examples, but I'm looking for something that is documented similar to the models-config doc.

There is a handy table of all valid config gets for models:
https://jujucharms.com/docs/2.1/models-config

To be consistent, it'd be good to see a corresponding reference table for controller config, maybe even at a predictably consistent URL such as:
https://jujucharms.com/docs/2.1/controllers-config

matthewhelmke:
Current commit is a first draft and is missing info. Eventually, this will fix #1699.

@mitechie who is the best person to ask for assistance filling in the missing details?

### DIFF
--- a/src/en/controllers-config.md
+++ b/src/en/controllers-config.md
@@ -1,0 +1,74 @@
+Title: General configuration options
+TODO: Check accuracy of key table
+      Make the table more space-efficient. Damn it's bulbous.
+
+
+# Configuring controllers
+
+A Juju controller is the management node of a Juju cloud environment. It houses
+the database and keeps track of all the models in that environment.
+
+Controller configuration consists of a collection of keys and their respective
+values. An explanation of how to both view and set these key:value pairs is
+provided below. Notable examples are provided at the end.
+
+
+## Getting and setting values
+
+You can display the current model settings by running the command:
+
+```bash
+juju controller-config
+```
+
+This will include all the currently set key values - whether they were set
+by you, inherited as a default value or dynamically set by Juju. 
+
+A key's value may be set for the current model using the same command:
+
+```bash
+juju controller-config auditing-enabled=true
+```
+
+It is also possible to specify a list of key-value pairs:
+  
+```bash
+juju controller-config auditing-enabled=true set-numa-control-policy=false
+```
+
+!!! Note: Juju does not currently check that the provided key is a valid
+setting, so make sure you spell it correctly.
+
+To return a value to the default setting the `--reset` flag is used,
+specifying the key names:
+  
+```bash
+juju controller-config --reset auditing-enabled
+```
+
+## List of controller keys
+
+The table below lists all the controller keys which may be assigned a value. Some
+of these keys deserve further explanation. These are explored in the sections
+below the table.
+
+| Key                        | Type   | Default  | Valid values             | Purpose |
+|:---------------------------|--------|----------|--------------------------|:---------|
+api-port                     | string |          |                          | The port to use for connecting to the API
+auditing-enabled             | bool   | false    | false/true               | Set whether auditing will be used. See [additional info below](#auditing).
+ca-cert                      | string |          |                          | The full contents of certificate created and used by the controller
+controller-uuid              | string |          |                          | The uuid of the controller
+set-numa-control-policy      | bool   | false    | false/true               | Set whether... See [additional info below](#numa-control-policy).
+state-port                   | string |          |                          | The port to use for connecting to the API
+
+### Auditing
+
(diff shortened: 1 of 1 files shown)


matthewhelmke:
@howbazaar, it will very likely be faster and easier for you to look at this PR and comment than for us to meet in a hangout.

matthewhelmke:
Gentle nudge, @howbazaar . If possible, we would really love to get this reviewed so we can push it.

src/en/controllers-config.md
```diff
@@ -0,0 +1,69 @@
...
+juju controller-config
+```
+
+This will include all the currently set key values - whether they were set
+by you, inherited as a default value or dynamically set by Juju. 
+
+A key's value may be set for the current model using the same command:
+
+```bash
+juju controller-config auditing-enabled=true
```
howbazaar: Probably worth noting that many of the settings are read only once the controller has been bootstrapped.

src/en/controllers-config.md
```diff
@@ -0,0 +1,69 @@
...
+|:---------------------------|--------|----------|--------------------------|:---------|
+api-port                     | integer | 17070    |                          | The port to use for connecting to the API
+auditing-enabled             | bool   | false    | false/true               | Sets whether the controller will record auditing information
+autocert-dns-name            | string |          |                          | Sets the DNS name of the controller. If a client connects to this name, an official certificate will be automatically requested. Connecting to any other host name will use the usual self-generated certificate.
+autocert-url                 | string |          |                          | Sets the URL used to obtain official TLS certificates when a client connects to the API. By default, certificates are obtained from LetsEncrypt. A good value for testing is "https://acme-staging.api.letsencrypt.org/directory".
+allow-model-access           | bool   |          |                          | Sets whether the controller will allow users to connect to models they have been authorized for even when they don't have any access rights tot he controller itself.
```
howbazaar: s/tot he/to the/

src/en/controllers-config.md
```diff
@@ -0,0 +1,69 @@
...
+ca-cert                      | string |          |                          | The certificate of the CA that signed the controller's CA certificate, in PEM format
+controller-uuid              | string |          |                          | The key for the UUID of the controller
+identity-public-key          | string |          |                          | Sets the public key of the identity manager
```
howbazaar: Is there other information on identity managers anywhere?

src/en/controllers-config.md
```diff
@@ -0,0 +1,69 @@
...
+identity-url                 | string |          |                          | Sets the URL of the identity manager
+mongo-memory-profile         | string | MongoProfLow |                      | Sets whether mongo uses the least possible memory or another mongo memory profile
```
howbazaar: You should specify what the possible values are.

MongoProfLow isn't the value, the two values are "low" and "default"

src/en/controllers-config.md
howbazaar: Acutally, checked the code, you can't actually set any values. Only get them.

matthewhelmke: Oh! Good to know. Thanks!

matthewhelmke: @howbazaar, so the settings can (only) be set during bootstrap? How is that done, by passing parameters in the bootstrap command?

src/en/controllers-config.md
matthewhelmke: Yes, in https://jujucharms.com/docs/stable/users-models

src/en/controllers-config.md
matthewhelmke: That is the area where I was hoping most for your kind assistance, @howbazaar. I don't know the valid or default values for the blank lines.

evilnick:
@matthewhelmke controller config are set when bootstrapping with the --config option. We already have examples of it https://jujucharms.com/docs/stable/controllers-creating#change-timeout-and-retry-delays

And also, those values  ^ should be in the table!

matthewhelmke:
Thanks @evilnick

matthewhelmke:
Note to self: when this content is reviewed and approved, add a link back from [here](https://jujucharms.com/docs/stable/controllers-creating#change-timeout-and-retry-delays) to this new page.

matthewhelmke:
Okay, I'll do that now, @evilnick.

PR #1784, opened 2017-04-05
deleted an unnecessary sentence and added a link with description

matthewhelmke:
The deleted sentence was a holdover from the configuring models page and didn't make sense in the configuring controllers page.

The link was added to the main Juju controllers page to the new config page, paralleling the link from the Juju models page to the models config page. The description may need to be massaged, but it is a start.

### DIFF
--- a/src/en/controllers.md
+++ b/src/en/controllers.md
@@ -79,7 +79,7 @@ Common tasks are summarized below.
    as whether the controller will record auditing information. Configuration
    must happen as the controller is bootstrapped.
 
-   This is a complex subject. See [Configuring controllers](./controllers-config.html).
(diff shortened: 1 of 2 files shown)


src/en/controllers.md
```diff
@@ -73,6 +73,16 @@ Common tasks are summarized below.
 
 
 
+^# Configure a controller
+
+   There are several settings which can be adjusted from their defaults, such
+   as whether the controller will record auditing information. Configuration
+   must happen as the controller is bootstrapped.
+
+   This is a complex subject. See [Configuring controllers](./controllers-config.html).
```
evilnick: Can we instead say:

This topic is covered in more depth in the [configuring controllers section]

matthewhelmke: Definitely.

PR #1790, opened 2017-04-10
Added info about debug-hooks working with actions

### LINKED ISSUE #1722: Clarify how to start Juju with desired mongo profile
As per https://bugs.launchpad.net/juju/+bug/1673314, users can choose what memory profile bootstrapped mongo will use.

matthewhelmke:
Fixes #1722 
@juju/docs

The additions to the two files are identical, much like the sections of the two files into which the info was added were already identical. This is a great example of where we could use an `include` feature someday, should we have time available to create the enhancement.

### DIFF
--- a/src/en/actions.md
+++ b/src/en/actions.md
@@ -178,3 +178,21 @@ actions:
   unit: git/0
 ```
 
+### Debugging actions
+
+To debug actions, use the `debug-hooks` command, like this:
+ +
+ +```bash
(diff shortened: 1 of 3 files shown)


evilnick:
They are identical because *someone* created a new page instead of updating or removing the old one. Part of the general eco mess.

However, this could also do with mentioning in the actions documentation

matthewhelmke:
I can do that. New commit to come later today.

PR #1791, opened 2017-04-10
Removed wip page that was not needed

### LINKED ISSUE #1774: https://jujucharms.com/docs/1.25/wip-users suggests using a command that doesn't exist
Moved from https://github.com/CanonicalLtd/jujucharms.com/issues/432

Juju 1.25.6, attempting to run password changes, the instructions state to use "juju system" but this returns with "ERROR unrecognized command: juju system".

The same page also doesn't specify what should 'bob' should do with the .server file when it's been sent after creating the new user.

matthewhelmke:
Fixes #1774 

We have a few more "work in progress" pages in 1.25 that we should examine and possibly remove, too.


evilnick:
Possibly. I think the centos stuff and spaces is valid though

PR #1792, opened 2017-04-11
Enhanced info on openstack cloud creds

matthewhelmke:
I read a recent question in the Juju mailing list. Our page already answered the user's question, but Ian commented in the list thread and his comment gave me the idea to enhance our page slightly. Hence, this PR.

@juju/docs

### DIFF
--- a/src/en/help-openstack.md
+++ b/src/en/help-openstack.md
@@ -10,8 +10,9 @@ information to add it to the list of known clouds.
 
 Using the Juju `add-cloud` command, it is easy to add your OpenStack clouds to
 Juju's list of known clouds. The command is interactive, and will ask for
-a name,endpoint,authorisation method(s) and regions to use. A sample session
-is shown below.
+a name,endpoint,authorisation method(s) and regions to use. (If you are unsure
+about endpoints and regions, they will be listed in the .novarc file, if you
+have one for your OpenStack.) A sample session is shown below.
 
 Running...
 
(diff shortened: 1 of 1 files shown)


src/en/help-openstack.md
```diff
@@ -20,6 +20,10 @@ juju add-cloud
 ```
 ...will enter the interactive mode. Enter the desired values to continue.
 
+!!! Note:
+To use the authentication from OpenStack, find the URL in your novarc the
```
evilnick: Thanks for adding extra info. I read this 3 times though and it still sounds badly worded. Perhaps:

If you have a `novarc` file for your OpenStack, the information you require will be found in there


Though also:
 * it is really jarring to have this note in the middle of the steps
 * If they have a .novarc file, they should just use autoload-credentials

matthewhelmke: That rewording is an improvement. I've had coffee since writing the original and agree with your assessment.

As I consider a better location for the information, we must remember that they need this information during the interactive `add-cloud`. Putting the note after the interactive bits is slightly late, IMO. Perhaps moving it above the command itself would be better. That is, unless I misunderstand the process and you can autoload-credentials BEFORE you add the cloud to Juju or can use autoload-credentials without adding the cloud at all.

evilnick: Not sure, I've not tried it.


Okay, looking at it again, I think this info should go just after the sentence "...regions to use." 

And probably say 

If you are unsure about endpoints and regions, they will be listed in the .novarc file for you OpenStack, if you have one.

Does that make it clearer. I mean, it doesn't even need to be a note, it could be in brackets at the end of that sentence

matthewhelmke: I like that better. I'll get a new commit ready soon.

PR #1793, opened 2017-04-11
Added info about vpc-id on AWS

### LINKED ISSUE #1745: Missing information on how to provide vpc-id value for aws models
When creating a new model on JAAS you may want to to provide a `vpc-id` via the CLI if you want it to be on a specific VPC. This is done with `juju add-model --config vpc-id=<id>` but this doesn't config option appear to be documented anywhere, in the docs, nor in the CLI help output.

matthewhelmke:
Fixes #1745 
@juju/docs

### DIFF
--- a/src/en/help-aws.md
+++ b/src/en/help-aws.md
@@ -100,6 +100,9 @@ Features supported by Juju-owned instances running within AWS:
 
 - Juju deploys to what Amazon refers to as *m3.medium* instances by default. you
   can specify different instance types by using [constraints][constraints].
+
+- You can add a new model to a specific virtual private cloud (VPC) from the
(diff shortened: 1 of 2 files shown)


src/en/help-aws.md
```diff
@@ -100,6 +100,9 @@ Features supported by Juju-owned instances running within AWS:
 
 - Juju deploys to what Amazon refers to as *m3.medium* instances by default. you
   can specify different instance types by using [constraints][constraints].
+
+- You can add a new model to a specific virtual private cloud (VPC) from the
```
evilnick: trying not to be picky, but I think it would be more correct to say  "you can create a new model in a specific..."

PR #1796, opened 2017-04-11
JAAS - remove specific image

evilnick:


evilnick:
@matthewhelmke sorry - meant to do this too

Although the image is useful, it is perhaps too specific and is highly likely to change, so probably best to cut it

PR #1797, opened 2017-04-12
JAAS page refinements

evilnick:
First part of work

* Remove alternative JAAS page	
* Remove JAAS/GUI section on model deletion
* Add tip on model management through JAAS


evilnick:
thx

PR #1802, opened 2017-04-14
Tweak the copy a little bit

mitechie:
This tweaks the copy a little bit. It attempts to de-emphasize JAAS the controller, focus the wording a bit, and phrase things slightly more technically correct.

### DIFF
--- a/src/en/getting-started-jaas.md
+++ b/src/en/getting-started-jaas.md
@@ -4,21 +4,21 @@ Title: Getting started with Juju as a Service
 
 Juju as a Service (JAAS) is the fastest and easiest way to model and deploy
 your cloud-based applications. You can deploy Kubernetes to GCE in 10 minutes
-or get OpenStack running in Azure in 15 minutes, all from your web browser.
+and build a Hadoop with Spark solution in Azure, all from your web browser.
 
 When you use JAAS, Canonical manages the Juju infrastructure. This frees you
-to concentrate on your software and solutions. With JAAS you can deploy,
-configure, and operate your applications on the largest public clouds:
+to concentrate on your applications. With JAAS you can deploy, configure, and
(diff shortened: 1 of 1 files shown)


src/en/getting-started-jaas.md
```diff
@@ -244,9 +244,8 @@ you can manually copy and paste the unique authorisation URL from the command
 output.
 
 After successful authentication, you will be asked to enter a descriptive name
-for the JAAS controller, giving you access to the same controller used by the
-JAAS web interface. This means any models or applications you have already
-deployed are now accessible from the command line.
+for the JAAS controller. This means any models or applications you have
+already deployed are now accessible from the command line.
```
pmatulis: Providing a name for the controller does not mean models/applications are now accessible from the CLI. Maybe just:

"Now|Henceforth, any models or applications..."

mitechie:
Thank you for the feedback. Updated the three items. PTAL

src/en/getting-started-jaas.md
```diff
@@ -244,8 +244,8 @@ you can manually copy and paste the unique authorisation URL from the command
 output.
 
 After successful authentication, you will be asked to enter a descriptive name
-for the JAAS controller. This means any models or applications you have
-already deployed are now accessible from the command line.
+for the JAAS controller. This means you can operate your models or
+applications in JAAS from the command line.
```
pmatulis: Sorry to continue with this but what does "This means" refer to? I still think you should break the logic here and just say:

"Now|At-this-time|Henceforth|Going-forward|Once-authenticated, you can..."

mitechie: No apology needed. I didn't look at it right from your last comment. I've updated it and agree.

PR #1808, opened 2017-04-21
various minor language tweaks

evilnick:


src/en/controllers-gui.md
```diff
@@ -85,7 +85,7 @@ components of your project work together.
 
 ## Building
 
-Another use for the GUI is building and managing an environment in an intuitive
+Another use for the GUI is building and managing an model in an intuitive
```
degville: an model

src/en/reference-numbering.md
```diff
@@ -2,8 +2,8 @@ Title: Machine and unit numbering
 
 # Machine and Unit numbering
 
-Within an environment juju keeps track of sequences for machines and units.
-(Prior to 1.25 just machines). This means that for the life of an environment 
+Within an environment Juju keeps track of sequences for machines and units.
```
degville: should this use of 'environment' still be here?

src/en/controllers-gui.md
evilnick: d'oh!

src/en/reference-numbering.md
evilnick: Yeah, I was unsure. In some places this uses enivrionment instead of model, but some places it is less definitely talking about a Juju entity and more about a thing. I will change it here though

evilnick:
thanks

PR #1821, opened 2017-04-28
Status history pruning age and size is configurable in 2.2beta3.

degville:
also, added full stops to list/table items as these - along with many lists and tables - were a mixture of both no stops and stops, and I think we should be consistent, one way or another. 

For discussion; my suggestion is the following:
no full stop/period: sentences/entries with less that 5 words.
yes to full stop/period: Longer sentences, multiple sentences, and lists or tables where there are both short and long entries/sentences.


pmatulis:
I use a stop whenever there is a full sentence (subject-verb-object).

PR #1828, opened 2017-05-03
Add new LXD zfs default

### LINKED ISSUE #1822: New loop device default value
From https://github.com/CanonicalLtd/jujucharms.com/issues/455

On the docs, Getting started with Juju section, the recomendation is let all default and option bellow:
Size in GB of the new loop device (1GB minimum) [default=10GB]: 20.
however, when i installed here, the default value is now 100 GB:
its like:
Size in GB of the new loop device (1GB minimum) [default=100GB]?:

evilnick:
Fixes #1822


degville:
I know this is a small issue, but I get the impression that the default amount varies depending on available storage. I've just tried this to check on a fresh updated VM of Xenial, and it suggests `default=15`.

evilnick:
ok, I will add a note instead

evilnick:
Actually, that just clutters things up for no gain. I will just mark it as won't fix

PR #1838, opened 2017-05-11
Replace instances of "service" with "application"

mitechie:
I've updated all the instances I could find that seemed to require updating. There's some that don't need to be renamed as they reference something that's not specifically a juju deployed service (e.g. service outage)

In some cases, where the use of service was heavy, I reworded things slightly.

I also moved the subordinate-service page to be subordinate-applications and added a redirect rule (updating the existing ones) and checking any links to the page that are also updated. 

Finally, @evilnick I made SURE to git mv on the file but GH still shows it as a complete new file in the diff below. I apologize and maybe there's some better trick that I'm missing to make this more clear.


src/base.tpl
```diff
@@ -79,7 +79,7 @@ combo load -->
 <meta name="twitter:creator" content="@ubuntucloud">
 <meta name="twitter:domain" content="jujucharms.com">
 <meta name="twitter:title" content="Documentation - Juju">
-<meta name="twitter:description" content="Learn about the using Juju through the Juju GUI or CLI and see how it makes creating, configuring, deploying and managing services in the cloud simple.">
+<meta name="twitter:description" content="Learn about the using Juju through the Juju GUI or CLI and see how it makes creating, configuring, deploying and operating software in the cloud simple.">
```
evilnick: This comes from jujucharms.com, so maybe it should be updated there :)

src/en/about-juju.md
```diff
@@ -2,12 +2,11 @@ Title: What is Juju?
...
-oriented architectures and application oriented deployments. Juju allows you to
-deploy, configure, manage, maintain, and scale cloud applications quickly and
-efficiently on public clouds, as well as on physical servers, OpenStack, and
-containers. You can use Juju from the command line or through its beautiful
-GUI.
+Juju is a state-of-the-art, open source modelling tool for operating software
+in the cloud. Juju allows you to deploy, configure, manage, maintain, and
+scale cloud applications quickly and efficiently on public clouds, as well as
+on physical servers, OpenStack, and containers. You can use Juju from the
+command line or through its beautiful GUI.
```
evilnick: This was the peer-reviewed-by-dozens text from the shared document a few months back. I am okay for updating it, but I checked the other place it is used ( https://jujucharms.com/how-it-works) and it seems that has been updated too, and now says this:

'What is Juju?

Juju is an open source application modelling tool that allows you to deploy, configure, scale and operate cloud infrastructures quickly and efficiently on public clouds such as AWS, GCE, and Azure along with private ones such as MAAS, OpenStack, and VSphere.

Its store allows access to a wide range of best practice solutions which you can deploy with a single command. You can use Juju from the command line or through its powerful graphical representation of the model in the GUI.'

src/en/about-juju.md
```diff
@@ -41,11 +40,12 @@ in seconds, on any cloud or server or virtual machine.
...
-application delivers, regardless of the machine on which it runs. So the Juju
-charm for an application includes (amongst other things) all the logic for
-writing configuration files for that application - that logic itself can be
-written in whatever language or tool the author of the charm prefers.
+files. Juju works a layer above that by focusing on the long term operations
+required to keep that software running over time, regardless of the machine on
+which it runs. So the Juju charm for an application includes (amongst other
+things) all the logic for writing configuration files for that
+application - that logic itself can be written in whatever language or tool
+the author of the charm prefers.
```
evilnick: similarly 

"Puppet and Chef are great tools for configuration management.

Juju provides a higher level of abstraction, supplying the tools to manage the full scope of operations beyond deployment and configuration management, regardless of the machine on which it runs.

One of the main advantages of Juju is its dynamic configuration ability, which allows you to re-configure services on the fly, add, remove, or change relationships between services, and scale in or out with ease, sharing the operational knowledge and making the most of the wider community."

src/en/authors-charm-hooks.md
```diff
@@ -2,7 +2,7 @@ Title: Charm hooks
 
 # Charm hooks
 
-A service unit's direct action is entirely defined by its charm's hooks. Hooks
+An application's unit's direct action is entirely defined by its charm's hooks. Hooks
```
evilnick: no 's on application

src/en/authors-charm-benchmarks.md
```diff
@@ -1,4 +1,4 @@
-Title: Benchmarking Juju services  
+Title: Benchmarking Juju applications
```
evilnick: All of these author-* docs are to lesser or greater extents, legacy docs now. They were supposed to be updated by eco but never were. I guess it doesn't hurt to fix up service/application, but we didn't bother in the past as a lot of them are orphaned and effectively unmaintained anyway. I think fixing up the legacy docs is a larger conversation/task

src/en/authors-charm-hooks.md
```diff
@@ -88,12 +88,12 @@ intervention on juju's part.
 should be used to ensure that the charm's software is not running, and will not
 start again on reboot.
 
-This hook is called when a service removal is requested by the client. It should
+This hook is called when a application removal is requested by the client. It should
```
evilnick: an application
also look for the other mulitples of times when s/service/application doesn't work :)

src/en/charms.md
```diff
@@ -75,7 +75,7 @@ background.
 
 As we said at the beginning, we are going to scale up this application to cope 
 with a lot of traffic. To do that we will use a common TCP/HTTP load balancing 
-service, HAProxy, so we should also deploy a charm for that now:
```
evilnick: I believe this is one of the ones we left previously, since it is valid to call it a load-balancing service, but I guess we can also refer to it as an application

src/en/glossary.md
```diff
@@ -2,17 +2,18 @@ Title: Glossary
 
 # Glossary of terms
```
evilnick: This glossary page is not exposed any more. this is another legacy issue - it wasn't deleted because a lot of the authors-* pages refer to it and it would therefore break links. 

The *actual* glossary page is https://jujucharms.com/docs/2.1/juju-concepts

I suppose it doesn't hurt to change it, but it will be deleted anyhow someday

tools/nav.txt
```diff
@@ -38,15 +38,15 @@
 +models.html@Models
```
evilnick: This file isn't used any more, so thanks for reminding me it exists, I can delete it now

htmldocs/.htaccess
```diff
@@ -18,11 +18,11 @@ RewriteRule ^service-config.html$ charms-config.html [R=301,L]
 RewriteRule ^charm-store.html$ authors-charm-store.html [R=301,L]
 RewriteRule ^charm-quality.html$ authors-charm-quality.html [R=301,L]
 RewriteRule ^charm-tests.html$ authors-testing.html [R=301,L]
-RewriteRule ^subordinate-services.html$ authors-subordinate-services.html [R=301,L]
+RewriteRule ^subordinate-services.html$ authors-subordinate-applications.html [R=301,L]
+RewriteRule ^authors-subordinate-services.html$ authors-subordinate-applications.html [R=301,L]
```
mitechie: I'm not sure, it came up in my find and replace. I assumed when we were talking about setting up redirects that this is the location they were setup.

src/base.tpl
mitechie: ah, ok. I'll submit a PR against that repo for this then.

src/en/authors-charm-benchmarks.md
mitechie: Yes, and I can see about helping with the larger one. As long as they're published searchable docs, however, I feel like updating them in line is reasonable.

src/en/glossary.md
mitechie: ok, let's go ahead and update it and I'll do some follow up work to remove the references in the author docs as a follow up PR

tools/nav.txt
mitechie: thanks, I had ping'd about it in irc the other day and it didn't seem like it was part of anything.

mitechie:
I can easily break this into multiple PRs if that's helpful. These do tend to stand on their own and I can group in a couple of files at a time or what not. Let me know what's most helpful.

src/en/about-juju.md
mitechie: Ok, I understand but I also think that 

    service-oriented architectures and application oriented deployments

Doesn't speak that clearly and directly to the value of Juju and I'd like to make this change. If I need signoff from someone I'll go get it.

src/en/about-juju.md
mitechie: Same here, I think that 

    long term operations

vs

    the service the-application delivers

is more in line with our current messaging.

src/en/authors-charm-hooks.md
mitechie: updated

src/en/authors-charm-hooks.md
mitechie: updated to an, I must be blind to the one you're referring to. Can I get a hint?

htmldocs/.htaccess
evilnick: I think they are managed elsewhere these days, but am not 100% sure.

src/en/about-juju.md
evilnick: I agree, I just think we ought to be consistent, so if we change it here we should also change it there. I will ping Clara and see if we need some special okays!

src/en/authors-charm-benchmarks.md
evilnick: it is reasonable, certainly. Let's do this for now,

src/en/glossary.md
evilnick: ok

evilnick:
Yes, I think splitting it would help a lot. Maybe into three bits:

 *  everything starting author-*
 *  everything reference-*
 *  everything else

That would be a great help!

src/base.tpl
mitechie: PR is up but can't land due to usso outage atm. Will land once things settle.

mitechie:
Ok, I've broken it down into 4 PRs

Just move the files and remove nav.txt
https://github.com/juju/docs/pull/1841

Everything author-
https://github.com/juju/docs/pull/1842

Everything reference-
https://github.com/juju/docs/pull/1843

Everything else
https://github.com/juju/docs/pull/1844

Closing this one and moving conversations to the others.

PR #1843, opened 2017-05-11
Rename service for the reference- pages

mitechie:


src/en/reference-status.md
```diff
@@ -8,18 +8,18 @@ following:
...
+|waiting      |yellow         | The unit is unable to progress to an active state because a application to which it is related is not running. That application might be in error, blocked, waiting or maintenance. This unit expects it will be able to proceed towards active as soon as the things it is concerned about are resolved there - it needs no human attention (otherwise its status would be blocked).|
+|active |green| This unit believes it is correctly offering all the applications it is primarily installed to provide. A application can be active even if additional relations are being configured and set up. For example, a Ceph OSD application that is connected to its monitors is active, even if you then relate it to Nagios; setting up a  subordinate does not impinge on the OSD functionality which is its primary purpose for existing.|
```
evilnick: An application

src/en/reference-status.md
```diff
@@ -8,18 +8,18 @@ following:
...
+|waiting      |yellow         | The unit is unable to progress to an active state because a application to which it is related is not running. That application might be in error, blocked, waiting or maintenance. This unit expects it will be able to proceed towards active as soon as the things it is concerned about are resolved there - it needs no human attention (otherwise its status would be blocked).|
```
evilnick: an application

src/en/reference-status.md
```diff
@@ -8,18 +8,18 @@ following:
...
-|active |green| This unit believes it is correctly offering all the services it is primarily installed to provide. A service can be active even if additional relations are being configured and set up. For example, a Ceph OSD service that is connected to its monitors is active, even if you then relate it to Nagios; setting up a  subordinate does not impinge on the OSD functionality which is its primary purpose for existing.|
+| blocked      | blue           | This unit needs manual intervention because of a problem associated directly with the unit itself. This may be due to bad or inconsistent configuration of the application, disk space being unavailable for the unit, or missing relations which are essential to application operations. The message should tell a human administrator what is required to unblock the application.<br> For example: “Please provide an object store. Please configure the minimum speed required by users of this application. Please link this application to a database.”|
```
evilnick: it would read better as 'which are essential to the application's operation.'

src/en/reference-status.md
```diff
@@ -8,18 +8,18 @@ following:
...
...t is actively doing stuff in preparation for providing those applications. This is a “spinning” state, not an error state. It reflects activity on the unit itself, not on peers or related units. The Juju agent will set the unit to maintenance when it begins the process of installing the charm - downloading it, unpacking it and calling the install hook.<br>The unit may transition to the maintenance state at any time, in any hook. The maintenance state implies that the unit has no reason to think that it will not provide applications soon. For example, the software might be formatting block devices, or replicating data from peer units, or handling an action which for whatever reason means that it is not actively providing application, but it is not aware of anything on its own config or in related unit config that will prevent it from bringing up its applications.<br>Counterparts to this unit should expect that application calls to this unit will fail during the period it is in the maintenance state, but unless they have an urgent need to move to another counterpart, they should just wait for this unit to finish churning.<br> Being in maintenance does not require human intervention.|
```
evilnick: some of these uses are a bit strained now, but particularly "is not actively providing application" definitely shouldbe changed. I suggest "providing the application".

This is the fun you can have when you swap something which can be a noun or a verb for something which only works as a noun ;)

mitechie: Thanks good point. I've reworded these and tried to split a bit between "the application" in juju terms and "the services" that a deployed application is meant to provide once up and running. I think it reads a bit better this way.

mitechie:
Thanks for the hints. I had misread that you meant that in that exact file but I see it was spread into another file. If course I think these are the ones I broke down and just did a regex replace with service/application vs manually doing each one as it seemed fairly repeat/programatic text. Doh!

Thanks for the diligence and patience to point them out in these files. I've pushed an updated with lots of "an" and some rewording in the status one.

mitechie:
Actually, I did a

    git grep "a application"

and found a few other references to clean up. This hits all those with a newly sync'd master and should remove the last of them.

evilnick:
Ok, cool. I will give this a once over then hopefully merge and backport

evilnick:
:tada:

PR #1853, opened 2017-05-15
change oracle-compute cloud type to oracle

hmlanigan:


hmlanigan:
to match juju PR: https://github.com/juju/juju/pull/7334
this change will go out with juju 2.2-rc1

degville:
Thanks for this! We actually updated oracle-compute to oracle with the release of b4, but with b4 the output from `juju add-cloud` was/is still showing `oracle-compute`,

We'll update this as soon as rc1 is released - I've noted the `vsphere` error you fixed too.

PR #1872, opened 2017-05-24
Replace provider with cloud/public cloud

degville:
as per feedback from @evilnick, replaced provider with cloud/public cloud for principle help docs.


degville:
Thanks. And yes, there are many more. I'll create another issue for those as that's going to be a more involved job and these were easy fixes for the principle help docs.

PR #1891, opened 2017-06-06
Update OpenStack Provider configuration and bootstrap details.

hmlanigan:
Clarify the process for setting up and bootstrapping a private OpenStack Cloud.


src/en/howto-privatecloud.md
```diff
@@ -1,6 +1,6 @@
-Title: Setting up private clouds with Simplestreams
+Title: Setting up image streams for private clouds with Simplestreams
```
axw: Maybe "image metadata" rather than "image streams", for people who are unfamiliar with simplestreams?

src/en/howto-privatecloud.md
```diff
@@ -17,11 +17,14 @@ required by the end user. However, those setting up a private
 cloud, or who want to change how things work (eg use a different
 Ubuntu image), can create their own metadata.
 
-This page explains how to use Juju and additional tools to generate
-this Simplestreams metadata and configure OpenStack to use them.
+There are two ways to accomplish this.  One to use Juju and additional
```
axw: One *is* to use

javacruft:
This part of the documentation mixes operations that the user of a cloud can perform (say in the instance where the cloud does not have streams data), with the operators that the operator of the cloud can perform with no clear distinction between the two.

For example, the glance-simplestreams-sync charm is only consumable on Juju deployed OpenStack Clouds, and by the operators who manage that environment.  Normal cloud users will also not be able to created endpoints in the keystone service catalog.

This is more a comment on some of the existing content, rather than specific to this change.

hmlanigan:
@javacruft I took a shot at making it more clear what type of permissions where needed for the different parts image metadata instructions.  I'm not quite sure of the correct terminology for different permissions levels, please comment on improvements there.

src/en/howto-privatecloud.md
```diff
@@ -17,14 +17,23 @@ required by the end user. However, those setting up a private
...
-this Simplestreams metadata and configure OpenStack to use them.
+There are a few ways to accomplish this based on the OpenStack
+configuration and your level of permissions with juju and/or the
+OpenStack deployment.
 
+* If you are a general user start with [Create image metadata with Juju][general].
+* If you have admin or operator permissions in the OpenStack deployment, start
+with [Create image metadata with Juju][general] and continue with
+[Upload the Simplestreams Metadata to Swift][swift].
+* If the OpenStack deployment was done with juju and you are permissions, use the [glance-simplestreams-sync charm][gsscharm].
```
gnuoy: Maybe "If the OpenStack deployment was done with juju and you have permissions to deploy charms alongside the OpenStack charms then"

src/en/howto-privatecloud.md
```diff
@@ -110,7 +119,23 @@ ls ~/simplestreams/*/streams/*
...
+environment will be able to create a service and view endpoints used by the following 
+instructions.
+
+### Requirements
+
+ - [image metadata has been created with Juju][general].
+ - python-openstackclient
+ - python-swiftclient
+
+### Create a Swift container and upload image metatdata
```
gnuoy: I know this is really beyond the scope of this change but I think we should generalise this so that it talks about an object store rather than swift. I used these instructions with a swift-free deployment (using ceph rados gateway) and it worked perfectly.

src/en/howto-privatecloud.md
```diff
@@ -230,36 +255,47 @@ The output from the previous command will be similar to the following:
...
+!!! Note: you can verify the url before bootstrap with
+wget $SWIFT_URL/simplestreams/images/streams/v1/index.json
+
 Enter the following command to register the endpoint with the Simplestreams
 service:
 
 ```bash
-openstack endpoint create --region $REGION --publicurl $SWIFT_URL/simplestreams/images \
+openstack endpoint create --region $REGION \
+   --publicurl $SWIFT_URL/simplestreams/images \
```
gnuoy: I don't think this is right. This bases the product-streams publicurl on the object-stores internal url ($SWIFT_URL contains the internal url). The public url should be based on the object-stores public url.

src/en/howto-privatecloud.md
```diff
@@ -230,36 +255,47 @@ The output from the previous command will be similar to the following:
...
-Finally, use either the network label or the UUID with the 'network' configuration
-option when bootstrapping a new controller:
+!!! Note: You must have permissions to deploy charms in the juju model running
+OpenStack to utilize this method for image metatdata management.
 
 
-```bash 
-juju bootstrap openstack --config image-metadata-url=$SWIFT_URL --config network=<network id>
-```
+### Deploying the charm to your OpenStack Cloud
```
gnuoy: Could we change "the charm" to something more specific like "the Glance simple streams charm" or "the metadata management charm". I'm just worried about the distinction between deploying charms using your new mystack cloud versus deploying charms to the underlying infrastructure and I think the title "Deploying the charm to your OpenStack Cloud" could refer to either.

src/en/howto-privatecloud.md
```diff
@@ -230,36 +255,47 @@ The output from the previous command will be similar to the following:
...
+
+[Glance Simplestreams Sync][glance-simplestreams-sync]
+
+It is recommended to use Swift as juju will automatically look for a
```
gnuoy: I don't really understand this comment. It is recommended to use Swift as opposed to what ? Also could we generalise and say an object store rather than swift ?

src/en/howto-privatecloud.md
gnuoy: Just to be clear, please don't block on this comment. The text is fine as is but it would be nice to generalise the wording at a later date.

hmlanigan: The upper level heading is now "Upload the Simplestreams Metadata to an object store" and I added a line that swift is used for the example, but any object store can be used.

src/en/howto-privatecloud.md
hmlanigan: using both public url and internal url now.

src/en/howto-privatecloud.md
hmlanigan: there is a charm config variable use_swift which will set the product-streams service etc.  clarified the statement.

src/en/howto-privatecloud.md
hmlanigan: updated.

PR #1928, opened 2017-06-15
Fix command formatting

evilnick:
@degville this should fix the formatting issues, re-add the links for 'see also' etc


evilnick:
No, I took a bit of time to update the generating script, and add in a bunch of caveats for the commands which don't behave properly (check usage of add-storage against help for example)

I will add the command generation script to the doc-utils now. It is still a horrible kludge but it sort of works

PR #1942, opened 2017-06-20
Add link to model config options to Creating controllers.

hmlanigan:
Make it a little easier to find all config keys available for bootstrap.


evilnick:
:cake:

PR #1961, opened 2017-06-28
updated about-charms image

### LINKED ISSUE #1811: About Juju image 'about-charms' is not consistent with new GUI
About 1/2 way down this page: https://jujucharms.com/docs/stable/about-juju

degville:
fixes #1811 

- updated about-charms image to reflect current GUI design and remove the use of services.

The content of the image should be updated if/when we look at the about text.

@juju/docs


evilnick:
Actually, as there is only one truth for images, the backport label doesn't apply!

PR #1971, opened 2017-07-07
review clouds-maas page

pmatulis:


src/en/clouds-maas.md
```diff
@@ -4,89 +4,114 @@ Title: Using MAAS with Juju
...
-...
-mainmaas            0                 maas        Metal As A Service
+Cloud        Regions  Default          Type        Description
+.
+.
+.
+maas-cloud         0                   maas        Metal As A Service
 ```
 
-Before you bootstrap this cloud, it is necessary to add the relevant
```
evilnick: i don't know why you would remove this part, especially as you have now moved something else between these pieces.

src/en/clouds-maas.md
```diff
@@ -4,89 +4,114 @@ Title: Using MAAS with Juju
...
+You are now ready to create a Juju controller (this will utilize a node in the
+MAAS):
 
 ```bash
-juju bootstrap mainmaas mainmaas-controller
+juju bootstrap maas-cloud maas-cloud-controller
 ```
 
-Above, the Juju controller was called 'mainmaas-controller'.
+Above, the name given to the new controller was 'maas-cloud-controller'.
```
evilnick: This is where you could explain node usage. E.g. "MAAS will allocate a node from it's pool to run the controller on. If you want to make sure a specific node is used for this, see constraints etc etc etc

src/en/clouds-maas.md
```diff
@@ -4,89 +4,114 @@ Title: Using MAAS with Juju
...
+.
+maas-cloud         0                   maas        Metal As A Service
 ```
 
-Before you bootstrap this cloud, it is necessary to add the relevant
-credentials, as explained below.
+### Manually defining MAAS clouds
+
+Alternatively, it is possible to manually define a single or multiple MAAS
+clouds with a file and then add a cloud by referring to such a file (still with
```
evilnick: I would cut 'and then...).' This only gives a partial explanation which isn't enough to actually use the feature, so you might as well just go to the source of truth

src/en/clouds-maas.md
```diff
@@ -4,89 +4,114 @@ Title: Using MAAS with Juju
...
+
+Enter maas-oauth:
+
+Credentials added for cloud maas-cloud.
 ```
 
-When prompted for "maas-oauth", you should paste your MAAS API key. Your API
-key can be found in the Account/User Preferences page in the MAAS web
-interface, or by using the MAAS CLI:
+We've called the new credential 'maas-cloud-creds' and used a MAAS user's API
```
evilnick: These couple of lines seem more confusing than the original. "when prompted for "maas-oauth" just seems much less opaque.

src/en/clouds-maas.md
```diff
@@ -4,89 +4,114 @@ Title: Using MAAS with Juju
...
+sudo maas-region apikey --username=$PROFILE
 ```
 
-**Note:** Juju does not echo this key back to the screen.
+Where $PROFILE is to be replaced by the MAAS username.
+
+## Create the Juju controller
 
-Now you can create a Juju controller with the bootstrap command:
+You are now ready to create a Juju controller (this will utilize a node in the
```
evilnick: 1. It is 'utilise'
2. I think it is probably better to explain it underneath (see next note)

src/en/clouds-maas-manual.md
```diff
@@ -64,4 +63,4 @@ testmaas           0                 maas        Metal As A Service
 It is necessary to add credentials for these clouds before bootstrapping them.
 See the [Documentation on MAAS credentials here][maas-credentials].
 
-[maas-credentials]: ./clouds-maas.html#adding-your-maas-credentials
+[maas-credentials]: ./clouds-maas.html#credentials
```
pmatulis: OK

src/en/clouds-maas.md
pmatulis: To be consistent, we would have to do this at the end of every section of every page where there are steps: "Before doing X you must do Y". I can keep it if you want.

src/en/clouds-maas.md
pmatulis: OK

src/en/clouds-maas.md
pmatulis: OK

src/en/clouds-maas.md
pmatulis: OK

src/en/clouds-maas.md
pmatulis: OK

src/en/clouds-maas.md
evilnick: No, we don't really need to add it to everything, just keep it here. The specific issue here, and why this was added, is because people using the tool automation didn't realise (and the tool doesn't tell them) that they also have to add credentials.

src/en/clouds-maas.md
```diff
@@ -99,14 +99,22 @@ Where $PROFILE is to be replaced by the MAAS username.
...
 Above, the name given to the new controller was 'maas-cloud-controller'.
+MAAS will allocate a node from its pool to run the controller on. If you want
+to make sure a specific node is used for this, use constraints (see
+[Create a controller with constraints][create-a-controller-with-constraints]).
+
+## Next steps
+
+You can now start deploying Juju charms and/or bundles to your MAAS cloud.
+Continue with Juju by visiting the [Models][models] and
+[Introduction to Juju Charms][charms] pages.
```
evilnick: :+1:

src/en/clouds-maas.md
pmatulis: I think what you're saying is that because `add-cloud` is interactive the user may presume that it does everything needed to use a cloud. I'll add it back.

evilnick: :+1:

PR #1972, opened 2017-07-12
Add bits on dynamic/persistent storage

axw:
... and refresh everything in general.
This drops references to my hacked
postgresql charm, simplifies some of
the usage examples, and adds documentation
on the new commands and behaviour.


evilnick:
Many thanks for this work!
there are a few language style things we may tweak, but I would rather merge this ASAP and we will play with it later.

PR #1975, opened 2017-07-13
getting-started / JAAS review

### LINKED ISSUE #1977: Review needed: getting-started

pmatulis:
Fixes #1977 

Notes:
 - Buttons sometimes use monofont and sometimes single-quoted. What is the standard for Juju docs? MAAS uses single-quotes.
 - In the JAAS introduction, the 3 compatible clouds link to their respective Juju pages. This may be a source of confusion. Users may follow these links and get lost. Initially, I expected these to be links to upstream (e.g. https://aws.amazon.com).
 - As noted on IRC, this getting-started page and the jaas-cli page duplicate information.

### DIFF
--- a/src/en/getting-started.md
+++ b/src/en/getting-started.md
@@ -2,29 +2,33 @@ Title: Getting started with Juju
 
 # Getting started with Juju
 
-Juju works across public clouds, private clouds, and locally with LXD-based
-deployments. Juju works through a controller that will process all of the
-events that occur. These instructions will walk you through using the hosted
-Juju as a Service (JAAS) controller and requires you to have public cloud
-credentials ready. If you'd prefer to test it out locally, see the [Getting
-started with Juju and LXD][tut-lxd] guide.
+Juju works with public clouds, private clouds, and locally with LXD. A central
+Juju controller (machine) is used to process all of the events that occur
+between the Juju client and the workload machines.
+
+The instructions on this page will show you how to get started with Juju by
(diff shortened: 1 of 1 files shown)


src/en/getting-started.md
```diff
@@ -68,19 +72,18 @@ Press the green 'Start building' button to get started...
 
 </table>
 
-
 You build your [model][models] by adding and relating applications from the
-[Charm Store][charmstore]. Press the green '+' symbol in the middle of the UI
-to start finding applications in the Charm Store.
+[Charm Store][charmstore]. Press the green '+' symbol in the middle of the web
+UI to search for applications in the Store.
```
evilnick: I would leave 'Charm Store'. We seem to have so many stores!

src/en/getting-started.md
```diff
@@ -195,13 +199,12 @@ details  {
    Juju via a snap package. See the [install][install] for more information.
 
 
-^# From the Ubuntu PPA
+^# For Ubuntu
```
evilnick: I think i prefer it as it was - this makes it seem like this is the part for Ubuntu users, when in reality the approved/default method is installing the snap

src/en/getting-started.md
pmatulis: OK

src/en/getting-started.md
pmatulis: OK

PR #1991, opened 2017-07-21
Review storage and update commands.md

### LINKED ISSUE #1716: Add information on attach-storage and detach-storage
This is related to #1687 as a continuation of the persistent storage story.

As of today, the `detach-storage` and `attach-storage` commands have landed, but the backends do not yet operate as documented in the spec. The `attach-storage` command will just fail with "not implemented", and `detach-storage` will detach but also destroy.

### LINKED ISSUE #1990: Review charms-storage

pmatulis:
Fixes #1990 
Fixes #1716

### DIFF
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -99,7 +99,7 @@ juju deploy postgresql --storage pgdata=ebs-ssd,100G
 ```
 
 Deploy Ceph OSD with 3x100GiB volumes per unit for data storage, and 1x10GiB
-per unit for journaling:
+per unit for journalling:
 
 ```bash
 juju deploy ceph-osd --storage osd-devices=100G,3 --storage osd-journals=10G
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -239,17 +239,8 @@ Destroy already detached storage 'pgdate/0' (remove it from the model):
 juju remove-storage pgdata/0
 ```
 
-If an attempt is made to remove storage that is currently in use (it is
-attached) Juju will refuse to comply and will emit a warning:
-
-```no-highlight
-failed to remove pgdata/0: cannot destroy storage "pgdata/0": storage is
-attached
-
-Use the --force flag to remove attached storage, or use
-"juju detach-storage" to explicitly detach the storage
-before removing
-```
+If an attempt is made to remove storage that is currently in use (i.e. it is
+attached) Juju will return an error.
 
 ### Cross-model storage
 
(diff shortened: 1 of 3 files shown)


src/en/charms-storage.md
```diff
@@ -2,109 +2,122 @@ Title: Using storage with Juju charms
...
+```bash
+juju deploy postgresql --storage pgdata=ebs-ssd,100G
 ```
 
-The `--storage` flag may be specified multiple times, to support
-specifying constraints for multiple stores. For example, the [Ceph OSD](https://jujucharms.com/ceph-osd/)
-charm supports two stores: osd-devices, and osd-journals. You can
-specify constraints for these separately:
+Deploy Ceph OSD with 3x100GiB volumes per unit for data storage, and 1x10GiB
+per unit for journaling:
```
evilnick: journalling

src/en/commands.md
```diff
@@ -464,7 +464,13 @@ Click on the expander to see details for each command.
 
 ^# add-storage
 
-    **Usage:** ` juju add-storage [options] <unit name> <charm storage name> | <charm storage name>=<storage constraints> ... `
+   **Usage:** ` juju add-storage [options] 
+<unit name> <storage directive> ...
```
evilnick: you might want to check this renders okay. This was another bit that script is supposed to rewrite (as above) because it is a) silly and b) breaks the foldout style

src/en/commands.md
```diff
@@ -839,7 +882,7 @@ Click on the expander to see details for each command.
                 Default region is specified by the CLOUDSDK_COMPUTE_REGION environment
                 variable.
 
-             3. On Windows, %APPDATA%\\gcloud\\application_default_credentials.json
+             3. On Windows, %APPDATA%\\gcloud\application_default_credentials.json
```
evilnick: do both not need to be escaped?

src/en/commands.md
```diff
@@ -4494,19 +4637,271 @@ Click on the expander to see details for each command.
 
    Supplying one key name returns only the value for the key. Supplying key=value
    will set the supplied key to the supplied value, this can be repeated for
-   multiple keys.
+   multiple keys. You can also specify a yaml file containing key values.
+   The following keys are available:
```
evilnick: ooh! this is useful!

src/en/charms-storage.md

src/en/commands.md
pmatulis: This was painful.

src/en/commands.md
pmatulis: I don't understand it. I only added the first one.

PR #1999, opened 2017-08-01
Add the bundle resources documentation.

ExternalReality:


src/en/charms-bundles.md
```diff
@@ -384,3 +433,4 @@ Freenode) who can assist. You can also use the
 [juju-list]: https://lists.ubuntu.com/mailman/listinfo/juju
 [constraints-docs]: ./charms-constraints.html
 [discover-config-options-docs]: ./charms-config.html#discovering-application-configuration-options
+[charm-resources-docs]: https://jujucharms.com/docs/2.1/developer-resources
```
evilnick: this should be a relative link to the docs (so that it points to the correct version), e.g. "./developer-resources"

PR #2001, opened 2017-08-02
storage: add ceph install sub-page; more review and examples

pmatulis:


evilnick:
Cool, will check in the morning

src/en/charms-storage.md
```diff
@@ -386,19 +412,21 @@ provider currently supports a single pool configuration attribute:
     - 'oracle' (volume type is 'default')
     - 'oracle-latency' (volume type is 'latency').
 
+### LXD (lxd)
+
```
axw: do you intend to add something here about the new 'lxd' storage provider?

src/en/charms-storage.md
```diff
@@ -82,36 +87,35 @@ effect:
...
+### juju add-storage
 
 ```bash
-juju deploy postgresql --storage pgdata=100G
+juju add-storage <unit> <label>[=<pool>,<size>,<count>]
 ```
 
-Deploy PostgreSQL with one instance (count) of 100GiB, via the charm's 'pgdata'
-storage label, using the 'ebs-ssd' storage pool:
+As with `juju add-unit` the parameters used are taken from the `juju deploy`
```

src/en/charms-storage.md
```diff
@@ -82,36 +87,35 @@ effect:
...
+storage label, using the default storage pool:
 
 ```bash
-juju deploy postgresql --storage pgdata=ebs-ssd,100G
+juju deploy postgresql --storage pgdata=100G
 ```
 
-Deploy Ceph OSD with 3x100GiB volumes per unit for data storage, and 1x10GiB
-per unit for journalling:
+A more explicit, but equivalent, command is:
```

src/en/charms-storage.md
pmatulis: Yep, that's coming in the next PR. I have some questions. :speech_balloon: Will send email.

src/en/charms-storage.md
```diff
@@ -18,49 +23,50 @@ management abilities. Related commands are listed below, along with a brief
...
 
 [`create-storage-pool`][commands-create-storage-pool]
 : Creates or defines a storage pool.
 
 [`detach-storage`][commands-detach-storage]
-: Detaches storage from a unit.
+: Detaches a storage instance from a unit.
+
+[`remove-storage`][commands-remove-storage]
+: Removes a storage instance from a model.
```
evilnick: I think it would help disambiguation if this command and the one above explicitly said what happened to the actual storage (I know this is covered later, but this is higher up the page and may be the only things skim readers see)

src/en/charms-storage.md
```diff
@@ -1,4 +1,5 @@
-Title: Using storage with Juju charms
+Title: Using Juju Storage
+TODO:  bug tracking: https://pad.lv/1708212
 
 # Using Juju Storage
```
evilnick: Does this need 'Juju'?

src/en/charms-storage.md
```diff
@@ -164,60 +172,72 @@ For example, as described above, attaching an EBS volume to an EC2 instance
...
-above example, the EBS volume was created in the same availability zone as the
-instance (a requirement).
+!!! Note:
+    Currently, it is not possible to add storage to the model without also
+    attaching it to a unit.
 
 Charms can specify a maximum number of storage instances. In the case of the
-postgresql charm, a maximum of one is allowed for 'pgdata'. If an attempt is
+charm 'postgresql', a maximum of one is allowed for 'pgdata'. If an attempt is
 made to exceed it, Juju will return an error.
```
evilnick: is this error message worth including?

src/en/charms-storage.md
```diff
@@ -164,60 +172,72 @@ For example, as described above, attaching an EBS volume to an EC2 instance
...
+#### Examples
+
+To create a 32GiB EBS volume and attach it to unit 'ceph-osd/0' as its OSD
+storage:
+
+```bash
+juju add-storage ceph-osd/0 osd-devices=ebs,32G,1
+```
+
+Above, the volume was created in the same availability zone as the instance (a
```
evilnick: This seems a roundabout way of saying "Storage will always be created in the same availability zone as the unit it is being added to"

src/en/charms-storage.md
```diff
@@ -164,60 +172,72 @@ For example, as described above, attaching an EBS volume to an EC2 instance
...
+!!! Important:
+    Detaching storage from a unit does not destroy the storage.
+
+### Persistence
+
+As we saw, detaching storage does not destroy the storage. In addition, when a
+unit is removed from a model, and the unit has dynamic storage attached, the
+storage will be detached and left intact. This allows detached storage to be
+re-attached to an existing unit using `juju attach-storage`, or to a new unit
+using the `--attach-storage` flag of `juju deploy` or `juju add-unit`:
```
evilnick: not sure why there is a ':' at the end here

src/en/charms-storage.md
```diff
@@ -226,22 +246,27 @@ juju deploy postgresql --attach-storage pgdata/0
...
+juju remove-storage osd-devices/3
 ```
 
-If an attempt is made to remove storage that is currently in use (i.e. it is
-attached) Juju will return an error.
+To upgrade the OSD journal of Ceph unit 'ceph-osd/0' from magnetic to solid
+state (SSD) and remove the uneeded original journal 'osd-journals/0':
 
+```bash
+juju add-storage ceph-osd/0 osd-journals=ebs-ssd,8G,1
```
evilnick: Does this mean that you can 'swap out' existing storage simply by adding new storage to the same slot? This would warrant further explanation I think, though I guess it depends on the charm and how it is using the storage

src/en/charms-storage-ceph.md
```diff
@@ -0,0 +1,123 @@
...
+ceph-osd/0  osd-journals/2  block  ebs   vol-0a328ced71048ce0b  8.0GiB  attached  
+ceph-osd/1  osd-devices/3   block  ebs   vol-0eaada2acc2dc851b  32GiB   attached  
+ceph-osd/1  osd-devices/4   block  ebs   vol-08d7fefb838d72217  32GiB   attached  
+ceph-osd/1  osd-journals/5  block  ebs   vol-0ca0e2ec9fb200e8a  8.0GiB  attached  
+ceph-osd/2  osd-devices/6   block  ebs   vol-0d0a9a3a8cb4c4464  32GiB   attached  
+ceph-osd/2  osd-devices/7   block  ebs   vol-0e0b54e0b70bb8ebd  32GiB   attached  
+ceph-osd/2  osd-journals/8  block  ebs   vol-07de85abbdd395e1b  8.0GiB  attached 
+```
+
+Here we see the two OSDs and one journal per OSD system, for a total of nine
```
evilnick: OSDs here is somewhat ambiguous - you are referring to the OSD storage, but it could easily be read as a Ceph OSD unit

src/en/charms-storage.md
```diff
@@ -164,60 +172,72 @@ For example, as described above, attaching an EBS volume to an EC2 instance
...
+Assuming the storage provider supports it, storage can be created and attached
+to a unit using `juju add-storage`. Juju will ensure the storage is allowed to
+attach to the unit's machine.
 
-Juju will ensure the storage is allowed to attach to the unit's machine. In the
-above example, the EBS volume was created in the same availability zone as the
-instance (a requirement).
+!!! Note:
+    Currently, it is not possible to add storage to the model without also
+    attaching it to a unit.
```
mitechie: This isn't true though for 2.3. There's the import-storage command.

src/en/charms-storage-ceph.md
```diff
@@ -0,0 +1,123 @@
+Title: Installing Ceph
+
+# Installing Ceph
+
+Here we present one way to install a rudimentary Ceph cluster with Juju backed
+by an AWS cloud. See [Ceph documentation][upstream-ceph-documentation] to find
+out more about Ceph.
```
mitechie: Have you reached out to @ChrisMacNaughton about any suggestions/best practice for the setup that might be worth including?

src/en/charms-storage-ceph.md
```diff
@@ -0,0 +1,123 @@
...
+juju deploy -n 3 ceph-osd --storage osd-devices=ebs,32G,2 --storage osd-journals=ebs,8G,1
+juju add-relation ceph-mon ceph-osd
+```
+
+This will take a while. Presented below is the final output to the `juju status`
+command:
+
+```no-highlight
+Model  Controller      Cloud/Region   Version       SLA
+ceph   aws-controller  aws/us-east-1  2.3-alpha1.1  unsupported
```
mitechie: I'd go ahead and trim the "alpha1.1 off the version just so that this stays up to date a bit more.

src/en/charms-storage-ceph.md
```diff
@@ -0,0 +1,123 @@
+Title: Installing Ceph
```
mitechie: would this work better as a tutorial on setting up a standalone ceph cluster with Juju?

src/en/charms-storage-ceph.md
```diff
@@ -0,0 +1,123 @@
...
+
+Here we see the two OSDs and one journal per OSD system, for a total of nine
+volumes.
+
+This Ceph cluster is used as a basis for some examples provided in
+[Using Juju Storage][charms-storage].
+
+## Ceph Object Gateway
+
+To allow clouds such as OpenStack and Amazon to access the Ceph cluster learn
```
mitechie: This seems funny. Why would Amazon access the ceph cluster? I assume this is more about end user access of the cluster?

src/en/charms-storage-ceph.md
```diff
@@ -0,0 +1,123 @@
...
+about the [Ceph Object Gateway][upstream-ceph-object-gateway] (RADOS gateway).
+To add a Ceph Object Gateway, formerly known as a 'RADOS gateway':
+
+```bash
+juju deploy ceph-radosgw
+juju add-relation ceph-radosgw ceph-mon
+```
+
+Some manual configuration is necessary to complete the gateway install - see
```
mitechie: That docs page has a lot of material in it and it's not 100% clear what's needed. Can/should we setup a sample config file and try to script this based off this setup at all?

mitechie: I think it'd be much more awesome if we could leave off where we stuck some content into the storage array and left the user with something that had some sample client setup communicating with it.

src/en/charms-storage.md
mitechie: I think so. All VMs come with some sort of storage and so just sticking with "Juju Storage" sets good context.

src/en/charms-storage.md
```diff
@@ -18,49 +23,50 @@ management abilities. Related commands are listed below, along with a brief
...
-The advanced storage features of Juju naturally depend upon charms that have
-storage options. This document will focus on the
-[PostgreSQL charm][postgresql-charm] which uses these features to store the
-database contents separately from the root filesystem.
+- 'pool': class of storage (e.g. magnetic, SSD)
+- 'size': size of each volume
+- 'count': number of volumes
 
-### Storage constraints
+The default pool (e.g. 'ebs' for AWS, 'cinder' for OpenStack) is given by:
```
mitechie: So above "pool" is described as "magnetic/ssd" but here it's "ebs" which can be both magnetic and ssd. This seems a bit confusing.

src/en/charms-storage.md
pmatulis: Good question. I just changed the page 'Title:' according to the existing top header.

src/en/charms-storage.md

src/en/charms-storage.md
pmatulis: No. :laughing:

src/en/charms-storage.md
pmatulis: Right. It's because it's already been said two times by this time. Here we're just stating what happened in this example. I was actually going to remove this sentence altogether.

src/en/charms-storage.md
pmatulis: It's collateral damage caused by a refactoring of the text that didn't get swept up by a thoroughgoing sub-review. I will replace it by a full stop.

src/en/charms-storage.md
pmatulis: Your question tells me the page isn't clear (my failure). Yes, the charm is the limiting factor to everything you can do to the application, including storage. To me, this is answered in the introductory section:

> Certain applications can benefit from advanced storage configurations and if a charm exists for such an application Juju can declare such requirements both at deploy time and during the lifetime of the application.

> The level of sophistication is limited by the charm; a charm may support multiple storage options (e.g. persistent storage, additional cache). All this allows the user to allocate resources at a granular level. This page will refer to the PostgreSQL charm and the Ceph OSD charm.

It's interesting that I appended the first sentence before reading your review note. Does it help?

This note under 'Storage constraints' also helps nail it:

> 'label' is a string taken from the charm itself. It encapsulates a specific storage option/feature.

It's difficult to explain in detail because of the charm-specific nature of what actually happens. In this specific case I could have just added a second OSD journal and things would work, yet from my Ceph days in the Support group this was never done. There was always a single journal. So this example perhaps has a little extra realism to it. Hopefully that enriches the reader experience and does not inadvertently detract from it.

Regardless, when a journal is added Ceph does indeed do a fair amount of processing before it is ready for usage (as it does when an OSD is added) but I didn't mention that and I don't think I should.

src/en/charms-storage-ceph.md
pmatulis: OK, will replace by "two OSD volumes and one journal volume per OSD system".

src/en/charms-storage.md
pmatulis: Will update this when I look at that feature. Is it not `import-filesystem`?

src/en/charms-storage.md
evilnick: yeah, that makes sense

src/en/charms-storage.md
evilnick: Yes, there is plenty of material to suggest that charms are custom, but the ability to swap storage is one which might not immediately spring to mind. If you can do this with the ceph charm it is worth pointing out that it is a feature of the charm before giving that example I think

src/en/charms-storage-ceph.md
evilnick: :+1:

src/en/charms-storage-ceph.md
pmatulis: No, but I'll assume he gets your mention and replies if he has anything to add. To be clear, I didn't intend this to become a sophisticated "how to install Ceph with Juju". If it's warranted we could have such a thing elsewhere in the Juju docs (or as a tutorial?).

src/en/charms-storage-ceph.md
pmatulis: Will update again at the next review pass.

src/en/charms-storage-ceph.md
evilnick: There is certainly space for a standalone tutorial covering this and more. The purpose of this page was merely to set up a minimal Ceph cluster so we could use it as a guided example. The simpler it is and the less detail it goes into the better, as a) people will need to follow it and b) we will need to maintain it in docs. It could easily link to a tutorial here if one were to exist

src/en/charms-storage.md
mitechie: Yes, sorry, you're correct. import-filesystem

src/en/charms-storage-ceph.md
mitechie: This starts to feel a bit tutorial-ish and regardless I think that it's worth asking the storage expert if there's anything just plain wrong/bad practice that's setup in a doc like this.

src/en/charms-storage-ceph.md
pmatulis: I would like to keep this rudimentary Ceph page as a sub-page to the storage page. We can definitely add a more sophisticated set of instructions elsewhere if anyone has time. I agree that  the Juju docs is not the place for in-depth treatments of third-party software. Not only is it inappropriate but it makes maintenance very heavy.

src/en/charms-storage-ceph.md
pmatulis: Yes, I could have left this out but it's a very common thing to do with Ceph these days. I will clarify it a bit.

pmatulis: Actually, I'm thinking of just ditching this section altogether. What do you think?

src/en/charms-storage-ceph.md
pmatulis: See previous comment.

src/en/charms-storage.md
pmatulis: Agreed. I just removed the top stuff. Now just:

> 'pool': class of storage

PR #2010, opened 2017-08-09
storage: general review and LXD storage provider

pmatulis:
I made some changes to the models.md page thinking I was going to do more, but didn't.

### DIFF
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -424,16 +424,16 @@ LXD-based models have access to the 'lxd' storage provider. The LXD provider
 does not currently have any specific configuration options.
 
 !!! Note:
-    LXD on all supported versions of Ubuntu provide the 'lxd' storage provider
-    feature with the exception of Ubuntu 16.04 LTS (Xenial). With Xenial,
-    either the 'backports' repository (`sudo apt install -y lxd/xenial-backports`)
-    or a PPA (`sudo apt-add-repository -yu ppa:ubuntu-lxc/lxd-stable; sudo apt
-    install -y lxd`) will be required.
+    To get an LXD version on either Ubuntu 14.04 LTS (Trusty) or Ubuntu 16.04
+    LTS (Xenial) that has the 'lxd' storage provider feature the
+    [LXD PPA][ppa-lxd] (`sudo apt-add-repository -yu ppa:ubuntu-lxc/lxd-stable;
+    sudo apt install -y lxd`) will be required.
 
 Every LXD-based model comes with a minimum of one LXD-specific Juju storage
 pool called 'lxd'. If ZFS and/or BTRFS are present when the controller is
-created then 'lxd-zfs' and 'lxd-btrfs' will also be available. The following
-output to the `juju storage-pools` command shows all three LXD-specific pools:
+created then pools 'lxd-zfs' and/or 'lxd-btrfs' will also be available. The
+following output to the `juju storage-pools` command shows all three
+Juju LXD-specific pools:
 
 ```no-highlight
 Name       Provider  Attrs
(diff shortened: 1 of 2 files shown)


src/en/charms-storage.md
```diff
@@ -133,6 +135,10 @@ rootfs   rootfs
 tmpfs    tmpfs     
 ```
 
+!!! Note:
+    The name given to a default storage pool will often be the same as the
+    name of the storage pool upon which it is based.
```
axw: I think you mean "same as the name of the storage provider upon which it is based"? in this case the default pool is "ebs", which  has the same name as the provider that services that pool

src/en/charms-storage.md
```diff
@@ -230,6 +233,10 @@ using `juju remove-storage`.
 If an attempt is made to either attach or remove storage that is currently in
 use (i.e. it is attached to a unit) Juju will return an error.
 
+Finally, a model cannot be destroyed while storage volumes remain without
+passing a special option (`--destroy-storage`). Naturally, this applies to
```
axw: or --release-storage?

src/en/charms-storage.md
```diff
@@ -268,6 +275,13 @@ juju detach-storage osd-journals/0
 juju remove-storage osd-journals/0
 ```
  
+To destroy a controller (and its models) along with all existing storage
+volumes:
+
+```bash
+juju destroy-controller lxd-controller --destroy-all-models --destroy-storage
+```
```
axw: maybe also give an example of using --release-storage?

src/en/charms-storage.md
```diff
@@ -415,6 +430,72 @@ provider currently supports a single pool configuration attribute:
 
 ### LXD (lxd)
 
+LXD-based models have access to the 'lxd' storage provider. The LXD provider
+does not currently have any specific configuration options.
```
axw: The LXD provider does have configuration options: `driver` and `lxd-pool`. The former is the LXD storage driver, the latter is the name to give to the corresponding storage pool in LXD. Any other configuration you specify will be passed along to LXD when creating the LXD storage pool (e.g. zfs.pool_name in the builtin lxd-zfs pool).

src/en/charms-storage.md
```diff
@@ -415,6 +430,72 @@ provider currently supports a single pool configuration attribute:
...
++------------+-------------+--------+------------------------------------+---------+
+| juju       |             | dir    | /var/lib/lxd/storage-pools/juju    | 0       |
++------------+-------------+--------+------------------------------------+---------+
+| juju-btrfs |             | btrfs  | /var/lib/lxd/disks/juju-btrfs.img  | 0       |
++------------+-------------+--------+------------------------------------+---------+
+| juju-zfs   |             | zfs    | /var/lib/lxd/disks/juju-zfs.img    | 0       |
++------------+-------------+--------+------------------------------------+---------+
+```
+
+To be clear, the three Juju-related pools above are for storing *volumes* that
```
axw: might also be worth pointing out that the pools are named according to the lxd-pool attribute in the juju storage pool configuration (and there's an implied lxc-pool="lxd" if none is specified)? i.e. lxd <-> juju, lxd-btrfs <-> juju-btrfs, lxd-zfs <-> juju-zfs

src/en/charms-storage.md
```diff
@@ -154,25 +160,22 @@ See [IOPS][wikipedia-iops] (Wikipedia) for background information.
...
+of storage, however, cannot be dynamically managed. For instance, Juju cannot
+disassociate MAAS disks from their respective MAAS nodes. These types of static
+storage can only be requested at deployment time and will be removed when the
+machine is removed from the model.
 
-Some types of storage, however, cannot be dynamically managed. For instance,
-Juju cannot disassociate MAAS disks from their respective MAAS nodes. These
-types of static storage can only be requested at deployment time and will
-be removed when the machine is removed from the model.
+Certain cloud providers may also impose certain restrictions when attaching
```
evilnick: the second 'certain' is certainly redundant :)

src/en/charms-storage.md
```diff
@@ -415,6 +430,72 @@ provider currently supports a single pool configuration attribute:
...
 
+LXD-based models have access to the 'lxd' storage provider. The LXD provider
+does not currently have any specific configuration options.
+
+!!! Note:
+    To get an LXD version on either Ubuntu 14.04 LTS (Trusty) or Ubuntu 16.04
+    LTS (Xenial) that has the 'lxd' storage provider feature the
+    [LXD PPA][ppa-lxd] (
+    `sudo apt-add-repository -yu ppa:ubuntu-lxc/lxd-stable; sudo apt install -y lxd`
+    ) will be required.
```
evilnick: Can we not rather specify a version of LXD, and then say this is currently available from the ppa?

src/en/charms-storage.md
pmatulis: Yep.

src/en/charms-storage.md
pmatulis: I'm not entirely sure of an exact version that provides all the described features. I originally had written 2.9. I can also just remove the commands.

src/en/charms-storage.md
pmatulis: Yes, storage provide. Will fix.

src/en/charms-storage.md
pmatulis: So is '--release-storage' functionally equivalent to detaching (`juju detach-storage`) storage from **all** units in a model?

src/en/charms-storage.md

src/en/charms-storage.md
evilnick: well, we could just say 2.16

pmatulis: Alright, I'll put "at least 2.16" and remove the commands.

src/en/charms-storage.md
pmatulis: I actually thought I was belabouring this point:

> TABLE containing Juju pool names and LXD pool names (with the exception of 'juju' explained below).

> As can be inferred from the above output, for each Juju storage pool based on the 'lxd' storage provider there is a LXD storage pool that gets created. It is these LXD pools that will house the actual volumes.

> The LXD pool corresponding to the Juju 'lxd' pool doesn't get created until the latter is used for the first time (typically via the juju deploy command). It is called simply 'juju'.

Should I add a summary table (your *id est*) under the above sentence?

src/en/charms-storage.md
pmatulis: Yeah, I fell victim to a copy & paste from the previous sections.

I understand passing storage constraints but not configuration options.

Why expose `driver` when it's integral to `lxd-pool`? IIUC, doing so would allow for incompatible pairs (e.g. btrfs driver with a zfs-based pool)?

Anyway, I don't know how to pass these configuration options. Can you give a few explicit examples, including one where options are passed to LXD?

src/en/charms-storage.md
axw: --release-storage means to remove it from the Juju database without destroying the corresponding cloud storage. The storage will be detached first.

src/en/charms-storage.md
axw: > Why expose driver when it's integral to lxd-pool? IIUC, doing so would allow for incompatible pairs (e.g. btrfs driver with a zfs-based pool)?

Juju creates the LXD storage pools if they don't exist. If they *do* exist, Juju will ensure that they have the same attributes as specified by the Juju user.

> Anyway, I don't know how to pass these configuration options. Can you give a few explicit examples, including one where options are passed to LXD?

This is done by creating a Juju storage pool. e.g.

```go
juju create-storage-pool lxd-zfs lxd driver=zfs lxd-pool=juju-zfs zfs.pool_name=juju-lxd
```

src/en/charms-storage.md
axw: Ah, fair enough, I glanced over that. I think it's fine as is.

src/en/charms-storage.md
axw: FWIW, I'd say the minimum is 2.15: https://insights.ubuntu.com/2017/07/12/storage-management-in-lxd-2-15/. There have been fixes since then, and there's no reason not to use 2.16, so may as well just say 2.16.

src/en/charms-storage.md
pmatulis: Thank you.

src/en/charms-storage.md
pmatulis: Thank you.

src/en/charms-storage.md
pmatulis: Good deal.

pmatulis:
@axw Wanna take a final look?

src/en/charms-storage.md
```diff
@@ -198,6 +197,10 @@ a different number. In any case, if detaching storage from a unit would bring
 the total number of storage instances below the minimum, Juju will return an
 error.
 
+It is not possible to add storage to a model without also attaching it to a
```
axw: maybe "add *new* storage"? the following sentence ("... you can add storage to a model...") reads like it contradicts this

pmatulis: Thanks.

PR #2012, opened 2017-08-11
Updated help-azure to use latest Azure CLI from NPM

### LINKED ISSUE #2000: Azure help is outdated
I went through setting up azure on juju yesterday, including having to manually add credentials for ci usage. The current directions reference to install an edge snap, which results in an out of date version of azure cli tools being installed. The instructions still seem to work for the intents of getting your information to feed into juju via autoload credentials (though az login seems broken), however the manual setup is incorrect. Due to the use of the older cli, it fails in non-intuitive ways, and juju and azure won't work well.

Let's update this page to reflect an updated and working method. I referenced the following upstream information to get things to work. https://docs.microsoft.com/en-us/azure/xplat-cli-connect and https://docs.microsoft.com/en-us/azure/cli-install-nodejs for information on obtaining the newer cli.

degville:
fixes #2000 

- updated help-azure to use latest Azure CLI from NPM
- updated manual credentials process
- tested both the above
- updated image to remove reference to 'Free Trial'.


matthewhelmke:
:-)

src/en/help-azure.md
```diff
@@ -26,44 +27,53 @@ Before you can use Azure with Juju, you will need to import your Azure
 account credentials into Juju. Retrieving those credentials is easy, thanks to
 Microsoft's new [Azure Cloud Shell][azureshell].
 
-This can be installed using *[snap][snapcraft]*:
+Azure Cloud Shell can be installed using *[npm][npminfo]*. Enter the following
```
axw: "Azure Cloud Shell" is something else. That's the shell you can get in the Azure portal web UI. This is just the "Azure CLI".

src/en/help-azure.md
```diff
@@ -26,44 +27,53 @@ Before you can use Azure with Juju, you will need to import your Azure
...
+command to install npm if it isn't already on your system:
 
 ```bash
-sudo snap install azure-cli --classic --edge
+sudo apt install -y nodejs-legacy npm
```
axw: Sorry I didn't see this happening earlier, but if we're updating these docs, please be aware that there's an entirely new Python-based Azure CLI (Azure CLI 2.0): https://github.com/Azure/azure-cli. This is the CLI recommended by Microsoft now, and supersedes the older NodeJS-based implementation.

src/en/help-azure.md
degville: Thanks - I'll change this.

src/en/help-azure.md
degville: I'll update the document to install and use Azure CLI 2.0 - thanks!

degville:
@axw I've updated this to use Azure CLI 2. It all works for me but I'd be grateful if you could scan over the updates when you get a chance. Thanks!

src/en/help-azure.md
```diff
@@ -128,157 +144,163 @@ In the sections below, we will assign each of these a variable name.  When you
...
 ```
 
-Use the APP_ID to create an Active Directory (Kerberos) server principal:
+We'll be using the value that follows **appId** as **APP_ID** and **tenant** as
+**TENANT_ID** (used later). Also make a note of `Object ID`, which we'll refer
+to as **OBJ_ID**
 
-!!! Note: Replace our variable here with the actual value you learned above.
-Do this throughout the rest of this page when you see variables listed.
+Now grant permissions to the principal with the following command, replacing
```
axw: Pretty sure you can fold this into the `create-for-rbac` command with the --role flag.

degville:
Thanks! I added role to create-for-rbac . I'm going to merge this now...

PR #2019, opened 2017-08-23
Add doc on how to change an application's series

hmlanigan:
to Getting things done.

### DIFF
--- a/src/en/howto-applicationupdateseries.md
+++ b/src/en/howto-applicationupdateseries.md
@@ -1,0 +1,117 @@
+Title: Change an application's series (prior to 2.3)
+
+#  Change an application's series (prior to 2.3)
+
+## Overview
+
+When deploying an application with Juju you can specify the series it
+will use, or use the default series for the model. Prior to Juju 2.3,
+any units which are subsequently added to the application always use
+the series specified at deployment
+
+With version 2.3 or later it's possible let juju know you'd like to add
+units with a different series moving forward by running the command:
+`juju update-series <application> <series>`
+
+Here we'll cover the manual steps on how to do this with earlier versions of the software.
+
+## Prerequisites
+
+  1. Check that the current charm revision in use for the application supports the
+  series you wish to use in the future.  This checked on in the Juju Charm Store.
+
+  2. Name of the application.
+
+  3. Name of the model where the application lives.
+
+  4. Admin privilages to the controller where the model lives.
+
+## Update
+
+Start by switching to the juju controller where the changes will happen using
+`juju switch`.
+
+Next, the database on the controller will need to be modified.  You'll need to
+know which model the application is deployed to, the application name and the
+series you'd like to use moving forward.
+
+An example script can be found [here][appendix].
+
+```bash
+mongo-update-app-series.sh -m default -a ghost -s xenial
+```
+```no-highlight
+MongoDB shell version: 3.2.12
+connecting to: 127.0.0.1:37017/juju
+2017-08-23T16:17:57.249+0000 W NETWORK  [thread1] SSL peer certificate validation failed: unable to get local issuer certificate
+2017-08-23T16:17:57.249+0000 W NETWORK  [thread1] The server certificate does not match the host name 127.0.0.1
+Connection to 10.106.28.244 closed.
+```
+
+Verify by checking the status of the application, getting the output in
+yaml format.  Output from: `juju status ghost --format yaml` will look similar to:
+```yaml
+...
+applications:
+  ghost:
+    charm: cs:ghost-20
+    series: xenial
+    os: ubuntu
+...
+```
+
+## Appendix: example script to update juju database
+
+```bash
+#!/bin/bash
+
+cntlr_model="controller"
+machine=0
+tmpfile="/tmp/series.js"
+usemsg="Usage: mongo-update-series -m <model> -a <application> -s <series>"
+
+while getopts "hm:a:s:" option; do
+	case $option in
+		a)
+			application=$OPTARG
+			;;
+		m)
+			appmodel=$OPTARG
+			;;
+		s)
+			series=$OPTARG
+			;;
+		h)
+			echo $usemsg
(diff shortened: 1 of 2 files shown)


src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,113 @@
...
+
+```
+$ juju switch localhost-test:admin/default
+```
+
+Next the database on the controller will need to be modified.  An example script
+can be found [here][appendix].
+
+```
+$ mongo-update-app-series.sh -m default -a ghost -s xenial
```
mitechie: can we package this up as a juju plugin? It could even be a plugin that follow the 2.3 semantics such that if you're on 2.3 it works per docs and if earlier it behaves the same way once the plugin is there?

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,113 @@
+Title: Change an application' series post deployment
```
mitechie: since this is really about how to do it pre-2.3 maybe title this in that way? Otherwise just use the normal documentation? So this turns into "how to change an application series pre-juju2.3" or something ?

hmlanigan: I've been going back and forth on this one myself.  Changed.

src/en/howto-applicationupdateseries.md
hmlanigan: It's not in the cards right now, however it's open to discussion in the future.

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,117 @@
+Title: Change an application's series pre-juju2.3
+
+#  Change an application's series pre-juju2.3
+
+## Overview
+
+When deploying an application with juju you can specify the series it
```
evilnick: I suggest:

When deploying an application with Juju you can specify the series it
will use, or use the default series for the model. Prior to Juju 2.3,
any units which are subsequently added to the application always use 
the series specified at deployment.

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,117 @@
+Title: Change an application's series pre-juju2.3
+
+#  Change an application's series pre-juju2.3
```
evilnick: I think "(prior to 2.3)" works slightly better

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Overview
+
+When deploying an application with juju you can specify the series it
+will use, or use the default series for the model.  Prior to juju 2.3,
+the application would always use the series specified at deployment
+when adding units with no way to change it.
+
+With juju 2.3 or later it's possible let juju know you'd like to add
```
evilnick: With version 2.3 or later, it's possible let Juju know..."

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,117 @@
...
+units with a different series moving forward by running the command:
+`juju update-series <application> <series>`
+
+Here we'll cover the manual steps on how to do this pre-juju2.3.
```
evilnick: "...how to do this with earlier versions of the software."

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Prerequisites
+
+1. Check that the current charm revision in use for the application supports the
```
evilnick: To make a proper numbered list in markdown, these should be indented (two spaces is usual) and always start with a "1."

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,117 @@
...
+series you wish to use in the future.  This checked on in the juju charm store.
```
evilnick: Juju Charm Store

src/en/howto-applicationupdateseries.md
```diff
@@ -0,0 +1,117 @@
...
+
+2. Name of the application.
+
+3. Name of the model where the application lives.
+
+4. Admin privilages to the controller where the model lives.
+
+## Update
+
+Start by switching to the juju controller where the changes will happen using
```
evilnick: Juju controller

evilnick:
Thanks for this. Let me know if you want to make the changes and retarget to 2.2 or we can take the information from this PR and do it if you prefer

src/en/howto-applicationupdateseries.md
hmlanigan: why is juju capitalized here and not other places?

src/en/howto-applicationupdateseries.md
hmlanigan: updated

src/en/howto-applicationupdateseries.md
hmlanigan: updated

src/en/howto-applicationupdateseries.md
hmlanigan: updated

src/en/howto-applicationupdateseries.md
hmlanigan: updated

src/en/howto-applicationupdateseries.md
hmlanigan: updated

src/en/howto-applicationupdateseries.md
hmlanigan: updated

hmlanigan:
I think I've accomplished the rebase to 2.2?

evilnick:
It looks like you have :smile: 
I will go over this again in the morning. thanks!

hmlanigan:
@evilnick, please review together with PR 2045, they are separate but go together.

evilnick:
there were still a few things that weren't updated, but I will add the fixes

PR #2021, opened 2017-08-25
Update juju status and other command output

pmatulis:
- I looked for all files containing `juju status` output. I corrected that as well as other types of output in those files.
 - An exception is controllers-backup.md . The backup stuff needs a thorough review (TODO added).
 - The tut-google.md page went deep into GUI stuff for some reason. Anyway, I noticed that when I followed the URL from `juju gui` the GUI was completely empty (no models nor apps). I had to comment out the existing image, no functional loss. Should definitely be revisited (TODO added).
 - Some minor rewording/corrections, some small amount of text removed, one broken link fixed.
 - Reformatted Note: entries

Fixes #1324 (output is different for 2.3)
Fixes #2011 (PR #2022 fixes this in master)

PR #2022, opened 2017-08-25
added clarification for firewall

pmatulis:
See PR #2021 for this change in 2.2.

### DIFF
--- a/src/en/tut-lxd.md
+++ b/src/en/tut-lxd.md
@@ -125,10 +125,11 @@ answers. IPv6 networking (the last question) is not required for Juju.
 
 LXD is now configured to work with Juju.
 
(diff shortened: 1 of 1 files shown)


evilnick:
Ok, I will approve this, but this is exactly the wrong way around to do things!

pmatulis:
I don't mind changing it. What is a better way?

evilnick:
there isn't now. The better way would have been to do the original fix here, in this PR, and backport it to 2.2

It's not a major thing, but if we change this text again there is likely to be a conflict between master and 2.2 now

pmatulis:
Oh that's what you mean. Yeah, I started out the 'status output' PR thinking I could go to master and backport everything but then I saw that the output for 2.3 and 2.2 differ. Then all hell broke loose.

PR #2025, opened 2017-08-28
charm upgrade config link

### LINKED ISSUE #1966: Charm upgrade docs should link to charm-config page for more detail
You can supply config as part of upgrade but it isn't really mentioned in the help text for the command and the current upgrade page also doesn't explain formats etc. This is already done in the charm-config page, so a link there would be useful

thanks @tvansteenburgh :)

pmatulis:
Fixes #1966

### DIFF
--- a/src/en/developer-upgrade-charm.md
+++ b/src/en/developer-upgrade-charm.md
@@ -27,7 +27,9 @@ settings are removed; and any new settings take defaults as though freshly
 deployed. Settings may be added or updated by specifying the `--config` flag,
 pointing to a YAML-encoded application config file (see
 [Application configuration][charms-config] for general information on
-configuring a charm).
+configuring a charm). It is important to read the documentation for the charm's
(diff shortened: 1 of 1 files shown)


src/en/developer-upgrade-charm.md
```diff
@@ -23,7 +24,10 @@ No other factor is used in determining compatibility: configuration settings in
 particular are converted completely naively, such that any settings from the
 original charm that share a name and type are preserved; any incompatible
 settings are removed; and any new settings take defaults as though freshly
-deployed.
+deployed. Settings may be added or updated by specifying the `--config` flag,
+pointing to a YAML-encoded application config file (see
+[Application configuration][charms-config] for general information on
+configuring a charm).
```
evilnick: It is probably worth mentioning that the *new* version of the charm may actually require additional config or settings so it is worth checking the charm's own documentation.

PR #2045, opened 2017-09-01
Add doc on how to change an machine's series 

hmlanigan:
to Getting things done.

### DIFF
--- a/src/en/howto-machineupdateseries.md
+++ b/src/en/howto-machineupdateseries.md
@@ -1,0 +1,143 @@
+Title: Change a machine's series (prior to 2.3)
+
+#  Update juju's understanding of a machine's series (prior to 2.3)
+
+## Overview
+
+If you want to let juju know that you've upgraded the series on an existing
+machine, and using juju 2.3 or later, you can run the command:
+`juju update-series <machine> <series>` 
+
+The update-series command will update the series saved for the given machine,
+as well as any units on that machine.  When a unit's series is updated, the
+config-changed hook will be triggered for that unit.
+
+Here we'll cover the manual step on how to do this prior to juju 2.3. Not
+covered is how to trigger the config-changed hook.
+
+!!! Note: 
+    Charms may have to be upgraded to take advantage series change.
+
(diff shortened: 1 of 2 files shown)


src/en/howto-machineupdateseries.md
```diff
@@ -0,0 +1,137 @@
...
+
+5. Admin privilages to the controller where the model lives.
+
+## 
+
+Start by switching to the juju controller where the changes will happen.
+
+```
+$ juju switch localhost-test:admin/default
+```
```
pmatulis: We use syntax highlighting using a "type" with 'bash', 'yaml', 'json', and 'no-highlight' being the most common. We also do not include the prompt.

>\```bash
>juju switch localhost-test:admin/default
>\```

We also separate input from output. Typically the output (which uses the 'no-highlight' type) is prepended by something like:

>This will give output closely resembling:

Try being creative with the above. It gets tiresome really quickly by using the same everywhere. An option is to sometimes include the command (if it's simple like `juju status` for example) within the preceding text itself (e.g. "Output to the `juju status` command will look like:").

src/en/howto-machineupdateseries.md
```diff
@@ -0,0 +1,137 @@
...
+
+If using juju 2.3 or later, you can run the `juju update-series` command.
+
+The `juju update-series <machine> <series>` will update the series saved
+for the given machine, as well as any unit's on that machine.  When a
+unit's series is updated, the config-changed hook will be triggered.
+
+This guide will cover how to do this prior to juju 2.3
+
+!!! Note: charms may have to be updated to take advantage series info change.
```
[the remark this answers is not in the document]
hmlanigan: change made.  'upgraded' used instead as that is closer to the related juju command.

src/en/howto-machineupdateseries.md
hmlanigan: Added the highlighting types.  Tried to make the output more interesting?

hmlanigan:
moved to 2.2 and updated title to match pr 2019.

PR #2064, opened 2017-09-06
replace entire example charm config, includes is_default key

### LINKED ISSUE #2023: [2.3] "default" is changed to "is_default"
As per a fix for https://bugs.launchpad.net/juju/+bug/1709692, the output for "juju config", from 2.3 release will specify whether the application configuration parameter value "is_default" instead of "default"

pmatulis:
Fixes #2023


pmatulis:
Yes.

PR #2065, opened 2017-09-07
add new upgrade charm page to user guide

### LINKED ISSUE #1965: Add user docs for charm upgrade
these currently exist only in the developer docs - we should add this to the user section as well

pmatulis:
Fixes #1965

### DIFF
--- a/src/en/charms-upgrading.md
+++ b/src/en/charms-upgrading.md
@@ -1,0 +1,28 @@
+Title: Upgrading applications
+
+# Upgrading applications
+
+An application is upgraded by upgrading its corresponding charm. Since a charm
+is always associated with a *channel* and a charm's version can fluctuate
+within that channel the following rules apply:
+
+- If the channel has a version **newer** than the deployed version then the latter
+  is **upgraded** to the former.
+- If the channel has a version **older** than the deployed version then the latter
+  is **downgraded** to the former.
+
+Channels can be specified with the `upgrade-charm` command. For example:
+
+```bash
+juju upgrade-charm mysql --channel edge
+```
+
+See [Deploying applications][deploy-charm_channels] for an overview of channels
+and the [Charm developer guide][dev-upgrade-charm] for even more details on the
+mechanics of charm upgrades.
+
+
+<!-- LINKS -->
+
+[dev-upgrade-charm]: ./developer-upgrade-charm.html
+[deploy-charm_channels]: ./charms-deploying.html#channels
--- a/src/en/charms-upgrading.md
+++ b/src/en/charms-upgrading.md
@@ -21,6 +21,25 @@ You can query for charm information, such as channels and revisions, with the
 `charm` utility, which is available via the Charm Tools software. See the
 [Charm Tools][charm-tools] page for guidance.
 
+## Crossgrading an application
+
+Crossgrading an application refers to upgrading an application by switching out
+the current charm and replacing it with a new local charm. This is accomplished
+by way of the `--switch` option. This differs from upgrading an existing charm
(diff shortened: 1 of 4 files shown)


src/en/charms-upgrading.md
```diff
@@ -0,0 +1,28 @@
+Title: Upgrading applications
+
+# Upgrading applications
+
+An application is upgraded by upgrading its corresponding charm. Since a charm
+is always associated with a *channel* and a charm's version can fluctuate
+within that channel the following rules apply:
+
+- If the channel has a version **newer** than the deployed version then the latter
```
evilnick: This seems more complicated an explanation than previously, mainly because of 'latter' and 'former'

src/en/charms-upgrading.md
```diff
@@ -0,0 +1,28 @@
...
+  is **upgraded** to the former.
+- If the channel has a version **older** than the deployed version then the latter
+  is **downgraded** to the former.
+
```
evilnick: It might be useful to give some example here

src/en/charms-upgrading.md
```diff
@@ -0,0 +1,28 @@
...
+Channels can be specified with the `upgrade-charm` command. For example:
+
+```bash
+juju upgrade-charm mysql --channel edge
```
evilnick: We also need to mention the use of '--switch', '--path' and '--revision'

src/en/charms-upgrading.md
pmatulis: OK

src/en/charms-upgrading.md
pmatulis: Pain.

src/en/charms-upgrading.md
pmatulis: The `switch` option does not upgrade an application. It *replaces* it with an entirely new one, to my amazement. Logically, this should be a part of charms-deploying since it is deploying a new application. Anyway, the linked developer-upgrade-charm.md file talks about it.

pmatulis: As for `--path`, how does one discover the "original path" if the application was originally deployed from the Store?

> Note that the directory containing the charm must match what was originally used to deploy the charm as a superficial check that the updated charm is compatible.

evilnick: Nevertheless, these are all options to this command! If you don't want to do this now, we need to at least create an issue for it. These things need to be added to this page (even if they make no sense or are discussed elsewhere)

pmatulis: Lots of pages do not cover every option every time a command is mentioned. That's what the commands.md page is for. I'll put it there anyways since you must see something I do not.

Do you know the answer to my `--path` question?

pmatulis:
@evilnick Please check.

src/en/charms-upgrading.md
```diff
@@ -21,6 +21,25 @@ You can query for charm information, such as channels and revisions, with the
...
+the current charm and replacing it with a new local charm. This is accomplished
+by way of the `--switch` option. This differs from upgrading an existing charm
+locally (with the `--path` option) due to the charm being considered entirely
+new.
+
+This is considered a dangerous operation since Juju has only limited
+information with which to determine compatibility. The operation will succeed
+so long as the following conditions are met:
+
+- The new charm must declare all relations that the application is currently
```
evilnick: I know this is technically correct, but it I think it could be misunderstood. Perhaps " the new charm must support..." or "be compatible with"

pmatulis: OK

PR #2073, opened 2017-09-08
remove legacy constraints

### LINKED ISSUE #1337: Remove or edit legacy constraint info
In [reference-constraints.md](https://github.com/juju/docs/blob/master/src/en/reference-constraints.md), there is a section entitled Legacy constraints. The information there is primarily from the pre-1.0 days and should be reviewed.

pmatulis:
Fixes #1337 

I was told by the Juju team that these "legacy constraints" are probably not used in 2.x but, for sure, they are not used in Juju 2.2. I propose that we remove them for 2.2 only. That is, unless someone else definitely knows whether they are also not used for 2.0 and 2.1.

### DIFF
--- a/src/en/reference-constraints.md
+++ b/src/en/reference-constraints.md
@@ -1,5 +1,4 @@
-Title: Juju constraints
-TODO:  Consider removing or editing Legacy section
+Title: Constraints
 
 # Constraints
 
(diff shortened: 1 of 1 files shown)


evilnick:
yes, I agree. 2.1 will cease to be an issue soon anyhow

PR #2078, opened 2017-09-11
Fix Yaml formatting in release notes

evilnick:
Annoyingly tricky to fix yaml formatting within a list within a foldout


evilnick:
@degville I thought you could give this the nod as it needs to chase the other fixes
(it works)

Issue #1981, opened 2017-07-17
Juju Doc Should Include a MAAS Compatibility Matrix


dsjpoulain:
The Juju 1.x doc should make it clear that Juju 1.x only works with MAAS 1.x, whereas the Juju 2 doc should state that Juju 2 can work with MAAS versions 1 and 2.

evilnick:
It doesn't really need a matrix as such, just a note on the 1.25 docs that it doesn't work with MAAS 2.0

dsjpoulain:
Thanks Nick. I'd still strongly recommend adding an explicit mention in the 2.x docs that Juju 2.x works with both versions of MAAS, as this would avoid any potential confusion from users, esp. those coming from Juju 1.25 - we've seen such confusion in the field on a number of occasions.

evilnick:
OK, cool. Will add to both then !

PR #2082, opened 2017-09-11
Added compatibility note to clouds-maas

evilnick:
Part of work for #1981

### DIFF
--- a/src/en/clouds-maas.md
+++ b/src/en/clouds-maas.md
@@ -5,7 +5,7 @@ Title: Using MAAS with Juju
 
 Juju works closely with [MAAS][maas-site] to deliver the same experience
 on bare metal that you would get by using any other cloud. Note that the 
-Juju 2.x software is compatible with both the 1.x and 2.x series of MAAS.
+Juju 2.x series is compatible with both the 1.x and 2.x series of MAAS.
 
 ## Add a MAAS cloud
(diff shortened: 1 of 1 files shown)


src/en/clouds-maas.md
```diff
@@ -4,7 +4,8 @@ Title: Using MAAS with Juju
 # Using a MAAS cloud
 
 Juju works closely with [MAAS][maas-site] to deliver the same experience
-on bare metal that you would get by using any other cloud.
+on bare metal that you would get by using any other cloud. Note that the 
+Juju 2.x software is compatible with both the 1.x and 2.x series of MAAS.
```
pmatulis: Awkward.

Maybe:

"Note that the Juju 2.x series is compatible..."

evilnick: fine

evilnick:
@pmatulis if the PR is from someone on the docs team, and you trust them to make requested changes, I think it is okay to follow what I now call the " @degville protocol" - make the comments inline but if there is no need to check the changes, give it an "approved" anyhow. I think everyone is sensible enough not to merge things which require a further look. This for example, would save you the time of coming back here and clicking on things to see that I changed one word.

pmatulis:
I forgot to click "Approve". That's all.

PR #2118, opened 2017-09-14
Add relevant notes about 2.2 reload-spaces command

### LINKED ISSUE #1920: Add docs for new 'reload-spaces' command
* check scope (MAAS/Oracle?)
*

evilnick:
Fixes #1920


evilnick:
Yeah, I didn't really want to repeat, but there is absolutely no point having a page specifically for that simple command, or redirecting people somewhere else just to tell them 3 lines of info.

I might have another look at the placement on the spaces page, but there was no other obvious place to put it, which is I guess why the other one is there

PR #2123, opened 2017-09-18
charm tools correction and clarification

### LINKED ISSUE #2120: correction: charms tools download directory
As per https://github.com/juju/charmstore-client/issues/55

pmatulis:
Fixes #2120 

It seems the standard is to use lowercase letters when referring to charms?


evilnick:
It seems so. The actual names are always lowercase so it avoids some confusion I suppose

PR #2136, opened 2017-09-21
2066 review resources

### LINKED ISSUE #2066: [2.3] review Resources concept
Command `juju charm resources` becomes `juju charm-resources`.
See PR https://github.com/juju/juju/pull/7695

Command `juju attach` becomes `juju attach-resource`.
See PR https://github.com/juju/juju/pull/7693

pmatulis:
Fixes #2066

### DIFF
--- a/src/en/charms-resources.md
+++ b/src/en/charms-resources.md
@@ -44,8 +44,7 @@ wal-e     0
 These are resources that a deployed application is currently using across all
 its units and are discoverable with the `resources` command. In contrast to the
 `charm-resources` command, this command needs a controller (that it queries) as
-well as deployed charm. An application gains resources via the
-`attach-resource` command.
+well as deployed charm.
 
 For example, to show the resources of the postgresql application:
 
@@ -59,6 +58,23 @@ Sample output:
 [Service]
 Resource  Supplied by  Revision
 wal-e     charmstore   0
+```
+
+An application gains resources, typically by a charm developer, via the
+`attach-resource` command:
+
+```bash
+juju attach-resource postgresql wal-e=/home/ubuntu/resources/wal-e-upgrade.snap
+```
+
+Where the charm's `metadata.yaml` file contains the following stanza:
+
+```no-highlight
+"resources":
+  "wal-e":
+    "type": "file"
+    "filename": "wal-e.snap"
+    "description": "WAL-E Snap Package"
 ```
 
 ## Unit resources
--- a/src/en/charms-resources.md
+++ b/src/en/charms-resources.md
@@ -1,0 +1,104 @@
+Title: Resources
+
+# Resources
+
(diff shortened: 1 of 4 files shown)


src/en/charms-resources.md
```diff
@@ -0,0 +1,103 @@
+Title: Resources
+TODO:  Critical: Improve examples
+
+# Juju Resources
+
+A Juju *resource* is additional content/files that a charm needs in order to
```
evilnick: it doesn't *always* need them - they may be optional. Perhaps "that a charm can make use of or may require"

src/en/charms-resources.md
```diff
@@ -0,0 +1,103 @@
...
+run.
+
+There are three ways to examine resources:
```
evilnick: three types of resources?

src/en/charms-resources.md
```diff
@@ -0,0 +1,103 @@
...
+Resource  Revision
+wal-e     0
+```
+
+## Application resources
+
+These are resources that a deployed application is currently using across all
+its units and are discoverable with the `resources` command. In contrast to the
+`charm-resources` command, this command needs a controller (that it queries) as
+well as deployed charm.
```
evilnick: as a deployed charm

src/en/developer-resources.md
```diff
@@ -1,25 +1,24 @@
-Title: Writing charms using resources
+Title: Resources | Developer guide
 
-# Writing charms that use resources
+# Developer resources
```
evilnick: They aren't really 'developer resources'. Perhaps 'Using resources' would be better

src/en/charms-resources.md
pmatulis: OK

src/en/charms-resources.md
pmatulis: Not so simple. *Charm resources* are simply what resources a charm is defined with. Once the charm's application is running, you can look at the other two types, in a runtime fashion, from either a application or a unit standpoint. So they all seem intrinsically equivalent to me, hence "examining".

src/en/charms-resources.md
pmatulis: OK

src/en/developer-resources.md
pmatulis: OK

pmatulis: Should I leave the menu alone (*Juju Resources*)?

src/en/charms-resources.md
evilnick: ok

src/en/developer-resources.md
evilnick: the menu item is ok

evilnick:
thanks :cake:

PR #2139, opened 2017-09-22
review of ssh access

### LINKED ISSUE #1417: Unable to Juju ssh from a user with write access to the model
According to https://bugs.launchpad.net/juju/+bug/1629141, we need to clarify in the docs that 'juju add-ssh-key' run by the admin user that created the model gives the other user access via ssh.

### LINKED ISSUE #2138: Review SSH access

pmatulis:
This is a significant update on how SSH works with Juju.

Fixes #2138 
Fixes #1417 

I could not destroy a model with 'admin' access so that's why I removed that ability. I'm not sure if this is a bug.

### DIFF
--- a/src/en/users-auth.md
+++ b/src/en/users-auth.md
@@ -49,26 +49,27 @@ account that public keys are added when using the Juju SSH commands (
 the 'root' user (passwordless sudo privileges), the granting of SSH access must
 be done with due consideration.
 
-### Native ssh
+It is possible to connect to a Juju machine in one of two ways:
 
-When using the native (OpenSSH) `ssh` client, if one's public key has been
-installed into a model, then, as expected, a connection to the 'ubuntu' user
-account can be made. All that is needed is the corresponding keypair and
-adequate network connectivity. 
+- Via Juju, using the `juju ssh` command
+- Direct access, using a standard SSH client (e.g `PuTTY` on Windows or `ssh`
+  [OpenSSH] on Linux)
 
-For example, to connect to a machine with an IP address of 10.149.29.143:
+Connecting via Juju involves a second degree of security, as explained below.
 
-```bash
-ssh ubuntu@10.149.29.143
-```
+Regardless of the method used to connect, a public SSH key must be added to the
+model. In the case of direct access, it remains possible for a key to be added
+to an individual machine using standard methods (manually copying a key to the
+`authorized_keys` file or by way of a command such as `ssh-import-id` for
+Debian-based Linux distributions such as Ubuntu).
 
 ### Juju ssh
  
-However, when using the `juju ssh` command, Juju's own user rights system
-imposes a second degree of security that will permit access solely from a Juju
-user, and only one with sufficient permissions. How this works depends on
-whether the user is an admin or a non-admin. See [Juju users][users] for a
-breakdown of the different user types.
+When using the `juju ssh` command, Juju's own user rights system imposes an
+extra degree of security by permitting access solely from a Juju user, and only
+one with sufficient permissions. How this works depends on whether the user is
+an admin or a non-admin. See [Juju users][users] for a breakdown of the
+different user types.
 
 For example, to connect to a machine with an id of '0':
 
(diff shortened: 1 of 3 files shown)


src/en/users-auth.md
```diff
@@ -24,22 +25,90 @@ juju bootstrap aws mycontroller
...
+adequate network connectivity. 
+
+For example, to connect to a machine with an IP address of 10.149.29.143:
+
+```bash
+ssh ubuntu@10.149.29.143
+```
+
+### Juju ssh
+
```
evilnick: So, the problem I have here isn't with the words but rather with the whole concept here. Juju provides additional 'security', except it doesn't, because if you don't have the correct access rights, you can just use a standard ssh client and get access anyhow

src/en/users-auth.md
```diff
@@ -24,22 +25,90 @@ juju bootstrap aws mycontroller
...
+to a model then that key is placed on all machines (present and future) in that
+model.
+
+Each Juju machine provides a user account named 'ubuntu' and it is to this
+account that public keys are added when using the Juju SSH commands (
+`juju add-ssh-key` and `juju import-ssh-key`). Because this user is effectively
+the 'root' user (passwordless sudo privileges), the granting of SSH access must
+be done with due consideration.
+
+### Native ssh
```
evilnick: I don't think 'Native' quite describes it properly. Maybe 'Direct SSH access'? or something

src/en/users-auth.md
```diff
@@ -24,22 +25,90 @@ juju bootstrap aws mycontroller
...
+
+When using the native (OpenSSH) `ssh` client, if one's public key has been
```
evilnick: i would be slightly wary here - it isn't JUST the OpenSSH ssh client, it is any client which uses the SSH protocol. So I think we have to talk about SSH rather than just 'ssh'

src/en/users-auth.md
evilnick: Perhaps this needs to be highlighted  in the first place - if you want Juju to have better security, you shouldn't add any keys at all, and perhaps this should actually be the main thrust of this section - start with Juju first, then deal with adding keys and direct access later, with a warning suggesting this leaves the machines open to unfettered access.

src/en/users-auth.md
evilnick: I only mention this because on other platforms, e.g. Windows, you can get SSH access without installing OpenSSH

src/en/users-auth.md
pmatulis: OK

src/en/users-auth.md
pmatulis: Right, thanks.

src/en/users-auth.md
pmatulis: Understood, except for this:

> if you want Juju to have better security, you shouldn't add any keys at all

A key *always* needs to be added. The only thing that is optional, in terms of allowing an ultimate connection, is 'admin' access at the model level. If it is granted, the user can use either `juju ssh` or a standard client. If it isn't, the user can use a standard client.

Given the extra steps a user needs to go through in order to use `juju ssh` (one time stuff like install Juju and register the user, but also needing to always log in to the controller, as well as reminding the Juju admin to grant 'admin' access; which probably confers other rights that the user should not have like destroying the model) I would think that the majority of users will opt for direct access. I am going to open a Juju bug on the matter (to somehow enforce `juju ssh`).

pmatulis: https://bugs.launchpad.net/juju/+bug/1719682

evilnick: To clarify, you don't need to add a key - juju automatically adds it's own keys. that seems to be the only way of ensuring that people don't just use direct access all the time. I.e., if the admin user disables adding keys.

evilnick: addendum - i'm not sure if that holds true when you are using a juju client from more than one machine (e.g. multi user on multiple machines)

pmatulis: So I confirmed that my understanding, as described in this PR, is correct.

Re your suggestion to "list keys post controller-creation", the admin's keys cannot be listed via `juju ssh-keys`. They are filtered out. :boar:

PR #2174, opened 2017-09-28
update charm tools usage

### LINKED ISSUE #725: Charm Tools need usage of commands updated
The usage section of the Charm Tools document should include the usage of the `charm build` command.  

Charm Tools 1.8.0 has the `charm build` command and that is not included in the docs.

We should add that section.

pmatulis:
Fixes #725 

This is a bit of a trainwreck but we are going above and beyond the call of duty with this.

In retrospect (because this took so much time), I should have just copied and pasted the help text. I didn't do this at the outset because the existing text does give useful information that is not in the help text. So I kept some of the old text, which *looks* ok but I am not in the position to verify for sure.

I need to assume the help text from version 2.2 gives more accurate information than from that of pre-1.8, on which the current text is based. However, I did find a few errors in it and corrected them when transposing it to this text form (such as the help sometimes including the dash between the `charm` command and a sub-command or by sometimes omitting the `charm` command or when the provided syntax is for the wrong command).

Most of the transposed text (from the help) was copied verbatim (a lot hasn't changed) but I made a few adjustments due to poor wording and a few grammatical errors. I also did not include minute detail. A few systematic changes were made to most sub-commands. To give an idea, here are two examples:

help:
*The pull command downloads a copy...*
new text:
*Downloads a copy...*

help:
*build a charm from layers and interfaces*
new text:
*Builds a charm from layers and interfaces.*

2.2 does not include sub-command `pull-source` in the sub-command listing but it shows up if its individual help page is invoked so I decided to keep it in. The `test` sub-command, however, is gone from both so I removed it. In the end, 4 sub-commands were added and 1 was removed but maybe there are other new sub-commands that don't show up in the listing.

:warning: This PR will push us closer to removing the macOS and Windows sections from the Charm Tools page since the versions available for these platforms do not include the 4 new sub-commands.

### DIFF
--- a/src/en/tools-charm-tools.md
+++ b/src/en/tools-charm-tools.md
@@ -50,30 +50,20 @@ the Windows Control Panel
 
 # Usage
 
-Charm Tools comes packaged as a stand alone tool and provides the `charm` command.
-There are several tools available within charm tools itself. At any time you can
-run `charm` to view the available subcommands and all subcommands have
-independent help pages, accessible using either the `-h` or `--help` flags.
+All the functionality provided by the Charm Tools is accessed via the `charm`
+command and a sub-command argument. Omitting the argument will give a listing
+of all sub-commands. To view an individual help page for each sub-command run
+`charm help <sub-command>`.
+
+Click the triangle to reveal a summary of a sub-command.
 
 ^# add
 
       charm add [-h] [--description] [-b] [--debug] {tests,readme,icon}
 
-  Add is a generator function which can be used to extend a charm depending on
-  the subcommand issued:
+  Extends a charm depending on the options chosen:
 
-  ### icon
-
-  `icon` will create an `icon.svg` in the `CHARM_DIRECTORY`. This icon is
-  a template and should be customised by the charm author (see
-  [Charm Icons documentation](authors-charm-icon.html)).
-
-  ### readme
-
-  `readme` will create a `README.ex` in the `CHARM_DIRECTORY` from the template
-  charm.
-
-  ### tests
+  #### tests
 
   `tests` will create a tests directory for a charm and generate an example test
   using the Amulet framework. This command will ingest relation data from the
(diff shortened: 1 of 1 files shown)


src/en/tools-charm-tools.md
```diff
@@ -110,287 +129,183 @@ independent help pages, accessible using either the `-h` or `--help` flags.
...
+  - [cs:]charm
+  - [cs:]series/charm
+  - [cs:]~user/charm
+  - [cs:]~user/series/charm
+  - layer:layer-name
+  - interface:interface-name
+  
+  If the item is a layered charm, and the top layer of the charm has a repo
+  key in layer.yaml, the top layer repo will be cloned. Otherwise, the charm
+  archive will be downloaded and extracted from the charm store.
```
evilnick: Charm Store

src/en/tools-charm-tools.md
```diff
@@ -110,287 +129,183 @@ independent help pages, accessible using either the `-h` or `--help` flags.
...
 
-      charm pull trusty/wordpress-trusty
+^# release
 
-  Will copy the Trusty version of the wordpress charm (files) into directory
-  `wordpress-trusty` under the current directory.
+      charm release [options] <charm or bundle id> [--channel <channel>]
 
-  To select a channel, use the `--channel` option, for instance:
+  Publishes a charm or bundle to the charm store. Releasing is the action of
```
evilnick: Charm Store

src/en/tools-charm-tools.md
```diff
@@ -110,287 +129,183 @@ independent help pages, accessible using either the `-h` or `--help` flags.
...
-   - [cs:]~user/charm
-   - [cs:]~user/series/charm
-   - layer:layer-name
-   - interface:interface-name
+      charm pull [options] <charm or bundle id> [--channel <channel>] [<directory>]
 
-  If the item is a layered charm, and the top layer of the charm has a repository
-  key in `layer.yaml`, the top layer repository will be cloned. Otherwise, the charm
-  archive will be downloaded and extracted from the charm store.
+  Downloads a copy of a charm or bundle from the charm store into a local
```
evilnick: Charm Store

src/en/tools-charm-tools.md
```diff
@@ -401,22 +316,44 @@ independent help pages, accessible using either the `-h` or `--help` flags.
...
+
+
+^# push-term
+
+      charm push-term [options] <filename> <term id>
+
+  Creates a new Terms and Conditions document (revision).
+
+  For example, to create a new terms document with the content from file
+  `text.txt` and the name 'enterprise-plan' and returns the revision of the
```
evilnick: return

PR #2193, opened 2017-10-02
[2.x] add Go cookies variable

### LINKED ISSUE #1666: cookies location explanation
As per https://bugs.launchpad.net/bugs/1634945, please clarify for our users that .go-cookies file location can be controlled by setting an environment variable.

pmatulis:
So just apply this to 2.x?

Fixes #1666


evilnick:
yup. I don't think this solves their bug, because nobody is going to look there

Issue #1563, opened 2016-12-07
/authors-charm-store


arosales:
ref = https://jujucharms.com/docs/stable/authors-charm-store

Need to update the Charm Developer workflow for contributors and maintainers to the follow:

If you are a current charm maintainer, we recommend the following actions:
- Ensure your charm is in your preferred SCM, e.g. Launchpad or Github.
- Create a team in Launchpad, e.g. <charm>-team (https://launchpad.net/people/+newteam).  This spreads maintenance tasks to team members who can push charm updates to the team’s name space in the charm store.
- Ensure you set the charm/bundle home page and bug link URL using ‘charm set’ (https://jujucharms.com/docs/stable/tools-charm-tools).  This helps users know where they can find more information about your application and where bugs/issues should be raised.
- Release the latest revision of your charm into the team’s name space (https://jujucharms.com/docs/stable/authors-charm-store).
- Submit a new charm revision for review @ http://review.jujucharms.com to have it promulgated to the recommended name space. Remember to release new revisions of the charm when updates are made in the upstream repo or when appropriate in the application release process.

If you would like to make a contribution to a charm and you are not currently the maintainer, find the charm detail page at http://jujucharms.com.  You should be able to find links to the project homepage and bug tracker where you can contribute charm changes.

PR #2195, opened 2017-10-04
update charm review & promulgation

### LINKED ISSUE #2145: Charm review docs and all references need to go
The Charm review queue has been sunset (see https://lists.ubuntu.com/archives/juju/2017-August/009294.html) and therefore all docs referring to this should be removed and cleaned up.

pmatulis:
Fixes #2145 

It seems there is no more review process for new charms.

I kept the original charm-review-process.md file just in case and created a new page called charm-promulgation.md. We can remove the old file once things have settled down.

I do not know how the author is supposed to make sure their charm "is of good quality and is secure".

I do not see any current instructions pointing authors to the (now defunct) review queue. I don't know how they ever used the old process to begin with, unless they stumbled upon the link (to the review queue site) amidst the old review instructions (not meant for authors).

### DIFF
--- a/src/en/charm-promulgation.md
+++ b/src/en/charm-promulgation.md
@@ -1,0 +1,18 @@
+Title: Charm promulgation
+
+# Charm promulgation
+
+To promulgate a Juju charm a member of the
+[~charmers][launchpad-group-charmers] Launchpad group will:
+
+1. Download the proposed charm (`charm pull`).
+1. Perform static analysis on the charm (`charm proof`).
+1. Ensure the charm has a bugs-url and homepage (`charm set`).
+1. If steps 2 or 3 fails, the charm author will be asked to make changes.
+1. Promulgate the charm (`charm promulgate`).
+1. Grant write access to the promulgated charm (`charm grant`).
+
+
+<!-- LINKS -->
+
+[launchpad-group-charmers]: https://launchpad.net/~charmers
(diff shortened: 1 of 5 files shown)


pmatulis:
@tvansteenburgh What do you think? Also, there appears to be some useful stuff in the [charm-review-process](https://github.com/juju/docs/blob/master/src/en/charm-review-process.md) page. I don't feel comfortable flushing it all and I'm not proficient enough with the subject matter to know what to keep.

pmatulis:
To make things even more complicated, there is also [reference-reviewers](https://github.com/juju/docs/blob/master/src/en/reference-reviewers.md) and [authors-charm-policy](https://github.com/juju/docs/blob/master/src/en/authors-charm-policy.md).

pmatulis:
Affects #1563

tvansteenburgh:
@pmatulis This looks great, thanks. You're correct that there are some more updates needed. I've had that on my to-do list for a while but haven's actually done it (obviously). Along with the pages you mentioned, I also found these others that will need some updates.

https://jujucharms.com/community/charmers
https://jujucharms.com/docs/2.2/charm-unmaintained-process
https://jujucharms.com/docs/2.2/authors-charm-store
https://jujucharms.com/docs/2.2/tools-charm-tools

Anyway, I don't expect you guys to do that - @evilnick is right that it's on us to provide accurate docs around the new review process. I just need to get around to it. In any case, this is a great start, and I really appreciate that you did it. Thank you!

PR #2206, opened 2017-10-04
explain charm upgrade

### LINKED ISSUE #598: Expand on upgrade-charm docs
It would be great if the upgrade charm docs outline in details what Juju does to the existing charm when performing the upgrade. The docs https://jujucharms.com/docs/1.24/authors-charm-upgrades don't really tell you in any detail what happens. One can infer that it extracts the new charm into the charm path overwriting all files which match but it's not very clear. It's also missing that it then runs either the install or charm-upgrade hook afterwards.

pmatulis:
Fixes #598

### DIFF
--- a/src/en/charms-upgrading.md
+++ b/src/en/charms-upgrading.md
@@ -21,6 +21,21 @@ You can query for charm information, such as channels and revisions, with the
 `charm` utility, which is available via the Charm Tools software. See the
 [Charm Tools][charm-tools] page for guidance.
 
+## Upgrade mechanics
+
+Here is a summary of how charm upgrades work.
+
+The agent (running on each unit of the application being upgraded) unpacks the
+new version of the charm to a staging directory next to the original charm. It
+then points itself to the newly unpacked charm and deletes the old one, after
+which the expected hooks are fired using the newly unpacked charm code.
+
+The original charm files are not overwritten. The original charm continues to
+run if anything goes wrong with the retrieval or unpacking of the new charm.
+
+To be clear, the logic of the upgrade itself is contained within the charm.
+Juju simply unpacks the new charm as noted and fires the hooks. 
+
 ## Crossgrading an application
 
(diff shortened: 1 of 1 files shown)


src/en/charms-upgrading.md
```diff
@@ -21,6 +21,21 @@ You can query for charm information, such as channels and revisions, with the
...
+## Upgrade mechanics
+
+Here is a summary of how charm upgrades work.
+
+The agent (running on each unit of the application being upgraded) unpacks the
+new version of the charm to a staging directory next to the original charm. It
+then points itself to the newly unpacked charm and deletes the old one, after
+which the expected hooks are fired using the newly unpacked charm code.
+
+The original charm files are not overwritten. The original charm continues to
```
evilnick: I don't understand this part. So, the new charm is in a completely new directory, but the old charm is still running in the original directory? They can't both be running at once? At some point the new charm has to take over and kill the process of the old one, surely?

pmatulis: Just because there are two directories doesn't mean both their corresponding software is running simultaneously. Once the new one is unpacked, the original one stops running and the new one begins to run (hooks).

src/en/charms-upgrading.md
```diff
@@ -21,6 +21,21 @@ You can query for charm information, such as channels and revisions, with the
...
+run if anything goes wrong with the retrieval or unpacking of the new charm.
+
+To be clear, the logic of the upgrade itself is contained within the charm.
+Juju simply unpacks the new charm as noted and fires the hooks.
```
evilnick: specifically it initiates which hook?

src/en/charms-upgrading.md
pmatulis: I am not prepared to say what happens if the upgrade itself fails. I would like to hope that the old one will start to run again but I don't have that information.

PR #2207, opened 2017-10-04
Update interface developer docs for Endpoints

johnsca:
These changes update the interface developer docs to use the new Endpoint style for writing interface layers.  This PR is dependent on juju-solutions/charms.reactive#123 and must not be merged until then.

### DIFF
--- a/src/en/developer-layers-interfaces.md
+++ b/src/en/developer-layers-interfaces.md
@@ -24,36 +24,66 @@ API.
 ## Terminology
 
 Historically, the term "relation" has been used ambiguously when discussing
-Juju, charms, and applications deployed with Juju using charms.  To make things
+Juju, charms, and applications deployed with Juju using charms. To make things
 more clear and consistent, we use the following terms:
 
-  * "charm" This is the set of code that encapsulates the operational knowledge
+  * A **charm** is the set of code that encapsulates the operational knowledge
     that manages the life-cycle of an application when deployed.
 
-  * "application" This is a deployed instance of a charm.  It is hosted on a
+  * An **application** is a deployed instance of a charm. It is hosted on a
     machine, VM, or container, it has a copy of the charm code, and it is
     managed by a Juju controller.
 
-  * "interface" This is the protocol used by the charm(s) running on two
-    applications to communicate over a relation established between them.
+  * The **interface** is the protocol used by two charms to communicate over a
+    relation between them.
     This protocol is what interface layers codify.
 
-  * "endpoint" This is the name that a charm uses as a connection point at which
-    a relation can be made between two deployed applications.  An endpoint is
-    defined in a charm's `metadata.yaml` and specify a name, a type (one of
-    `requires`, `provides`, or `peers`), and an interface that relations
-    established at the endpoint will use.
+  * An **endpoint** is one charm's connection point at which a relation can be
+    made. An endpoint is defined in a charm's `metadata.yaml` and specify a
+    name, a **role** (one of `requires`, `provides`, or `peers`), and an
+    interface that relations established at the endpoint will use.
 
-  * "relation" This is an established connection between endpoints of two
-    applications.  Relations have a relation ID that is tracked by the Juju
+  * A **relation** is an established connection between endpoints of two
+    applications. Relations have a relation ID that is tracked by the Juju
     controller and relation data associated with each unit of each application
-    involved in the relation.  Relations can only be established between
+    involved in the relation. Relations can only be established between
     endpoints that specify the same interface, and a given endpoint can only
     have one relation to a specific application. However, a given application
     can be related to an application any number of times on different
     endpoints, as long as they specify the same interface, and even to itself
     if it has two endpoints of the same interface (or if the endpoint is of the
     type `peers`).
+
+Here's how these terms map to a charm's `metadata.yaml` and `layer.yaml` file.
+
+```yaml
+# metadata.yaml
+name: mattermost  # Charm name
+# ...
+requires:  # Endpoint role
+    postgres:  # Endpoint name
+        interface: pgsql  # Interface name
+provides:  # Endpoint role
+    website:  # Endpoint name
+        interface: http  # Interface name
+# ...
+```
+
+```yaml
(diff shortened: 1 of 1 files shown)


johnsca:
[(2) Endpoint: Documenting Interface Layers](https://trello.com/c/v9A3M9uV/66-2-endpoint-documenting-interface-layers)

src/en/developer-layers-interfaces.md
```diff
@@ -1,25 +1,60 @@
...
 
-Interface layers currently must be written in Python and extend the ReactiveBase
-class, though they can then be __used__ by any language using the built-in CLI
+Interface layers currently must be written in Python and extend the [Endpoint]
+class, though they can then be *used* by any language using the built-in CLI
 API.
 
+[Endpoint]: https://charmsreactive.readthedocs.io/en/latest/charms.reactive.relations.html#charms.reactive.altrelations.Endpoint
+
+## Terminology
```
galgalesh: :+1:

src/en/developer-layers-interfaces.md
```diff
@@ -1,25 +1,60 @@
...
+    machine, VM, or container, it has a copy of the charm code, and it is
+    managed by a Juju controller.
+
+  * "interface" This is the protocol used by the charm(s) running on two
+    applications to communicate over a relation established between them.
+    This protocol is what interface layers codify.
+
+  * "endpoint" This is the name that a charm uses as a connection point at which
+    a relation can be made between two deployed applications.  An endpoint is
+    defined in a charm's `metadata.yaml` and specify a name, a type (one of
```
galgalesh: I'd add a sample `metadata.yaml` with comments on each key/value pair explaining what denotes the interface name, the endpoint name, the endpoint type etc..

src/en/developer-layers-interfaces.md
```diff
@@ -141,41 +142,27 @@ We're now ready to implement the provider interface in `provides.py`.
...
+        for relation in self.relations:
+            relation.send['hostname'] = hostname or hookenv.unit_get('private_address')
+            relation.send['port'] = port
 ```
 
-!!! Note: You may notice we do not decorate for the -broken hook event. This
-is intentional.  It will only actually cause problems if you use set_state
-in the context of a relationship being broken, but it doesn't work like you'd
-expect so it's better to avoid it entirely
+!!! Note: You can only "send" data at the relation level. Remember, a relation
```
galgalesh: Best to also specify that data is only sent **after** the hook exits. (this also gets reset when hook fails, right?)

Maybe we can make this explicit by calling the dict `relation.to_send[]`, saying "if the hook exits cleanly, then the `to_send` dict will be sent over the relation".

src/en/developer-layers-interfaces.md
```diff
@@ -193,9 +180,11 @@ And in the reactive class, we will receive the RelationBase object, and invoke
...
 
 ```python
-@when('website.available')
-def configure_website(website):
-  website.configure(80)
+from charms.reactive import context
+
+@when('endpoint.website.joined')
+def configure_website():
+    context.endpoints.website.configure(80)
```
galgalesh: Does this "configure" all the relationships of that website? If so, great :+1:, then the name really fits the functionality.

src/en/developer-layers-interfaces.md
```diff
@@ -204,63 +193,68 @@ def configure_website(website):
...
-            # this unit's conversation has a port, so
-            # it is part of the set of available units
-            conv.set_state('{relation_name}.available')
+from charms.reactive import when_any, when_not
+from charms.reactive import set_flag, clear_flag
+from charms.reactive import Endpoint
+
+
+class HttpRequires(Endpoint):
+    @when_any('endpoint.{relation_name}.changed.hostname',
```
galgalesh: shouldn't this be `{endpoint_name}` instead of `{relation_name}`?

src/en/developer-layers-interfaces.md
```diff
@@ -204,63 +193,68 @@ def configure_website(website):
...
+              'endpoint.{relation_name}.changed.port')
+    def new_website(self):
```
galgalesh: To be clear, the `self` here is not specific to the hook/relation that triggered this handler, right? There is only one `HttpRequires` object for every endpoint using this interface, correct? We might want to make this explicit in the docs.

src/en/developer-layers-interfaces.md
```diff
@@ -141,41 +142,27 @@ We're now ready to implement the provider interface in `provides.py`.
...
+
+class HttpProvides(Endpoint):
+    def configure(self, port, hostname=None):
+        """
+        Configure the HTTP relation by providing a port and optional hostname.
+
+        If no hostname is provided, the unit's private-address is used.
+        """
+        for relation in self.relations:
+            relation.send['hostname'] = hostname or hookenv.unit_get('private_address')
```
galgalesh: It  might be best to make it explicit that `send` is **the only** way to share anything with the remote unit. All flags etc. are always local, never shared.

src/en/developer-layers-interfaces.md
galgalesh: Even better, why don't we call this `to_publish`? This describes the behavior even better. Each unit publishes relation data visible to all related units.

What do you think, @johnsca ?

evilnick:
I see the referenced juju-solutions branch is now merged. Is it okay to merge this now?

galgalesh:
Not yet, it's currently still in beta.

evilnick:
Okay, thanks for the update!

galgalesh:
This has been released to stable, you can merge this PR. :tada:

pmatulis:
@galgalesh @johnsca I backported to 2.3 but I am seeing conflicts when backporting to 2.2 (and probably 2.0). Since this is just one file can I safely "copy" developer-layers-interfaces.md to 2.2 and 2.0?

galgalesh:
That should be ok, yes.

pmatulis:
@galgalesh Ok, all done.

galgalesh:
\o/

PR #2208, opened 2017-10-04
[devel + 2.2] change to signed 2.2.4 windows exe link

### LINKED ISSUE #2121: Install commands should reference signed windows exe
We make a signed windows installer that we should probably point people to instead of the unsigned exe on

https://jujucharms.com/docs/stable/reference-install

The url difference is small, https://launchpad.net/juju/2.2/2.2.4/+download/juju-setup-2.2.4-signed.exe instead of https://launchpad.net/juju/2.2/2.2.4/+download/juju-setup-2.2.4.exe.

Actually curious if there is a reason to have an unsigned exe published on the milestone, or why we differentiate. We could drop the unsigned exe and push the signed one to https://launchpad.net/juju/2.2/2.2.4/+download/juju-setup-2.2.4.exe instead. That would mean the docs wouldn't have to change.

pmatulis:
Fixes #2121 

 - For 2.3, I will ask Core if they can try to have the signed exe ready for release time.
 - For 2.2, this PR can be backported to it.
 - For 2.1, the same file ([reference-install]( https://github.com/juju/docs/blob/2.1/src/en/reference-install.md)) needs to be updated. **I noticed that this file points to 2.2.1 for all non-Ubuntu platforms.** Is this normal?
 - For 2.0, an update is needed for [reference-releases](https://github.com/juju/docs/blob/2.0/src/en/reference-releases.md) to point to [juju-setup-2.0.4-signed.exe](https://launchpad.net/juju/2.0/2.0.4/+download/juju-setup-2.0.4-signed.exe).
 - For 1.25, no updates are needed (to [reference-releases](https://github.com/juju/docs/blob/1.25/src/en/reference-releases.md)).

### DIFF
--- a/src/en/reference-install.md
+++ b/src/en/reference-install.md
@@ -58,7 +58,7 @@ Alternatively, you can manually install Juju from the following archive:
 ### Windows
 
 A Windows installer is available for Juju and can be found here:
-: [juju-setup-2.2.4.exe](https://launchpad.net/juju/2.2/2.2.4/+download/juju-setup-2.2.4.exe)([md5](https://launchpad.net/juju/2.2/2.2.4/+download/juju-setup-2.2.4.exe/+md5))
+: [juju-setup-2.2.4-signed.exe](https://launchpad.net/juju/2.2/2.2.4/+download/juju-setup-2.2.4-signed.exe)([md5](https://launchpad.net/juju/2.2/2.2.4/+download/juju-setup-2.2.4-signed.exe/+md5))
 {: .instruction }
 
 ### Centos and other Linuxes


evilnick:
Hi. Sorry I was going to do this before but I was distracted by the fact that the link looks *terrible* in the docs - there seems to have been some CSS change which insists on making it look like the least likely thing you would ever want to click on. but yes, we should update the link

evilnick:
I also think we should go back to the scheme where the reference-releases page was identical in all 2.0 series. It seems odd that the 2.0 docs claim the latest version is 2.0.4...

evilnick:
So rather than backporting changes, you may just want to copy this page to all 2.x series

pmatulis:
How did this happen?
https://github.com/juju/docs/blob/2.1/src/en/reference-releases.md
https://github.com/juju/docs/blob/2.1/src/en/reference-install.md

Confirm that you want **only** an identical reference-releases.md across 2.x and reference-install.md can be removed in 2.1, 2.2, and devel?

evilnick:
@pmatulis Ah, I forgot that there were changes to those docs and the way they relate during the 'JAAS' amendments. Forget it for now, just fix the current one. I'll make an issue for unifying it

evilnick:
To clarify - install is the NEW page. We should really have install in 2.0 also

pmatulis:
@nskaggs We will simply **not** publish a link to a Windows installer unless it points to a signed one. The mantra now is "signed only".

pmatulis:
@evilnick If I understand you correctly, I just need to merge this PR to both master and 2.2. Re master, and the corresponding devel branch on the web site, I think it's fine to point to the 2.2 signed Windows installer. That's what we have at this time in the dev cycle.

You will then fix up (unify) the rest via another issue?

evilnick:
Yes, this needs to go to master and 2.2. I'm not going to fix 2.1 I don't think as it is not long for this world, but I will at some point fix 2.0

Unsigned .exes are fine for dev versions I guess - the point is we only want to add ONE link, not an unsigned one that we then update to a signed one at some unspecified point in the future

pmatulis:
I said "signed 2.2" in dev (because that's the latest signed we have at this time).

PR #2209, opened 2017-10-05
Overview of updating the series of a juju controller or workload.

hmlanigan:


hmlanigan:
How do you provide an html link from a 2.2 doc to this branch?  The doc included needs to link to the docs in PRs 2019 and 2045.

Swift review will be appreciated as we'd like to have this live for the 2.3 beta release in the next day or so.

pmatulis:
@hmlanigan Hi. Once the previous 2 PRs are merged to master, if they apply to 2.2 then they will be backported to 2.2 making them available for linking from 2.2 **and** master/2.3 pages. It's unclear whether this current PR is for 2.3 **only**. Regardless, even assuming this current PR is 2.3-only, you can still link to those pages in this PR as they will definitely, presuming no one wants you to change the page names, end up in master/2.3 (as mentioned). They will only work, though, once those PRs are merged. Just do this for now:

```
[app-update]: ./howto-applicationupdateseries.html
[mach-update]: ./howto-machineupdateseries.html

evilnick:
The best/only way to link to documents in a different branch is to use absolute links to where they are/will be published

Anyhow, there are all sorts of small bits to change here. In the interests of getting it done quickly I am going to merge this and then do a new PR with the fixes

hmlanigan:
@pmatulis, This doc could be for 2.1, 2.2 and moving forward.  Not sure I'd
want to go farther back?  The other two docs would work for 2.1 also.

PR #2221, opened 2017-10-06
Charms.reactive documentation has moved to readthedocs

galgalesh:
This PR fixes all the links.

*PS: This PR also removes whitespace before newline in some files since atom does that automatically during save*


galgalesh:
These fixes will need to be backported to all versions (even 1.25 has some pages on `charms.reactive`). How does does this work? Do I need to do something to make that happen?

evilnick:
@galgalesh they do , but no, it's fine, we have a tool for doing that

galgalesh:
Any update on getting this backported? The docs on pythonhosted are already a few releases behind, so we'd like to see this fixed asap.

PR #2236, opened 2017-10-11
Add lost token info for user registration

### LINKED ISSUE #2020: get new user registration string
There is no a way to re-generate a registration for new user. This helps with scenarios where a user lost provided token and controller admin need to provide a new registration to enable user to register/login.

User passwords can be reset as a consequence too.

For more details, bug https://bugs.launchpad.net/juju/+bug/1657187 with linked PRs.

evilnick:
Fixes #2020


evilnick:
apologies for some format fixes

src/en/users-creating.md
```diff
@@ -65,13 +65,34 @@ juju register
 MD4TA2ppbTAVExMxMC4xOTUuMTAuMTMwOjE3MDcwBCDENImZEDvYcWLU1USgaUJQXhWR98JNLWcbS0-MiW6NiQA=
 ```
 
+The user will then be prompted to enter and confirm a password, and also supply a unique name for the controller.
+
+### Lost tokens
+
+If a situation occurs where the registering token fails to work for some reason, or is
+mislaid, a new token can be generated by the admin user with the `change-user-password` command.
```
pmatulis: I know we use en-GB but I've never heard of 'mislaid' before. Maybe consider something more direct like 'lost'.

src/en/users-creating.md
```diff
@@ -20,10 +20,10 @@ controller.
...
 the following additional actions (in the context of the controller):
 
-- List controllers (`juju controllers`)
-- Show controller details (`juju show-controller`)
-- List enabled controller users (`juju users`)
-- Show details for controller users (`juju show-user`)
+ - List controllers (`juju controllers`)
+ - Show controller details (`juju show-controller`)
+ - List enabled controller users (`juju users`)
+ - Show details for controller users (`juju show-user`)
```
pmatulis: Does adding an extra space do something special?

evilnick: it is more proper, depending on which of the many versions of markdown syntax you subscribe too. It doesn't make any significant difference to us, except maybe it looks nicer?

src/en/users-creating.md
evilnick: hmm. ok. But mislay is also in all the US dictionaries . Maybe it is just Canada that doesn't use it ;)

PR #2237, opened 2017-10-11
remove review item from nav

pmatulis:
Recall that the review process has gone away but, for now, the review page remains. This PR is to simply remove a link to it from the menu.


evilnick:
cool

PR #2248, opened 2017-10-20
CMR concepts and scenarios

### LINKED ISSUE #2225: Finish practical portion of CMR docs
The original PR for CMR (#2219) was incomplete due to time constraints (2.3 beta 1 just *had* to be released with those docs). Only the conceptual side was done. This issue is to get the practical example scenarios published.

pmatulis:
- This is presented in an unorthodox manner.
   - One page has concepts and simplified command syntax.
   - Two pages present scenarios that give more practical usage.
   - Once stuff settles down I'm willing to refactor.
 - Another pass of the commands (and their outputs) will be necessary prior to 2.3 release. If the model or controller names are too long I can change them at this time.
 - The links to the scenarios are not obvious. Is this another style fail?
 - Another scenario should be added soon (TODO).

fixes #2225

### DIFF
--- a/src/en/charms-relations.md
+++ b/src/en/charms-relations.md
@@ -55,13 +55,13 @@ error: ambiguous relation: "mediawiki mysql" could refer to
 ```
 
 The solution is to be explicit when referring to an *endpoint*, where the
-latter has a format of `<application>:<application endpoint>`. In this case,
-the application endpoint is 'db' for both applications but because only the
-mediawiki endpoint is ambiguous (according to the error message), it is not
-necessary to mention the mysql endpoint:
+latter has a format of `<application>:<application endpoint>`. In this case, it
+is 'db' for both applications. However, it is not necessary to specify the
+mysql endpoint because only the mediawiki endpoint is ambiguous (according to
+the error message). Therefore, the command becomes:
 
 ```bash
-juju add-relation mediawiki:db mysql
+juju add-relation mysql mediawiki:db
 ```
 
 !!! Note:
--- a/src/en/charms-relations.md
+++ b/src/en/charms-relations.md
@@ -49,10 +49,21 @@ In some cases, there may be ambiguity about how the applications should connect.
 For example, in the case of specifying a database for the Mediawiki charm.
 
 ```bash
-juju add-relation mediawiki mysql
+juju add-relation mysql mediawiki 
 error: ambiguous relation: "mediawiki mysql" could refer to
   "mediawiki:db mysql:db"; "mediawiki:slave mysql:db"
 ```
+
+The solution is to specify the nature of the relation using the
+relation interface identifier. In this case, we want MySQL to provide the 
+backend database for mediawiki ('db' relation), so this is what we need to 
+enter:
+
+```bash
+juju add-relation mysql mediawiki:db
+```
+
+<!-- REMOVED FROM PR 2248, TO BE REVIEWED LATER
 
 The solution is to be explicit when referring to an *endpoint*, where the
 latter has a format of `<application>:<application endpoint>`. In this case, it
@@ -70,6 +81,8 @@ juju add-relation mysql mediawiki:db
     [Charm Store][charm-store] or by querying the Store with the
     [Charm Tools][charm-tools] (using a command like
     `charm show <application> charm-metadata`).
+
+-->
 
 We can check the output from `juju status` to make sure the correct relationship
 has been established:
--- a/src/en/charms-relations.md
+++ b/src/en/charms-relations.md
@@ -1,4 +1,5 @@
 Title: Managing relationships  
+TODO:  Critical: review required
 
 # Managing relationships
 
@@ -53,14 +54,21 @@ error: ambiguous relation: "mediawiki mysql" could refer to
   "mediawiki:db mysql:db"; "mediawiki:slave mysql:db"
 ```
 
-The solution is to specify the nature of the relation using the
-relation interface identifier. In this case, we want MySQL to provide the 
-backend database for mediawiki ('db' relation), so this is what we need to 
-enter:
+The solution is to be explicit with the ambiguous *interface*, where the latter
+has a format of `<application>:<application endpoint>`. In this case, the
+*endpoint* is 'db' for both but only the mediawiki interface is ambiguous, so
+this is what we can do:
 
 ```bash
 juju add-relation mediawiki:db mysql
 ```
+
+!!! Note:
+    An application endpoint can be discovered by looking at the metadata of the
+    corresponding charm. This can be done by examining the charm on the
+    [Charm Store][charm-store] or by querying the Store with the
+    [Charm Tools][charm-tools] (using a command like
+    `charm show <application> charm-metadata`).
 
 We can check the output from `juju status` to make sure the correct relationship
 has been established:
@@ -116,3 +124,5 @@ Relations can also work across models, even across multiple controllers. See
 <!-- LINKS -->
 
 [models-cmr]: ./models-cmr.html
+[charm-tools]: ./tools-charm-tools.html
+[charm-store]:  https://jujucharms.com
--- a/src/en/charms-relations.md
+++ b/src/en/charms-relations.md
@@ -54,10 +54,11 @@ error: ambiguous relation: "mediawiki mysql" could refer to
   "mediawiki:db mysql:db"; "mediawiki:slave mysql:db"
 ```
 
-The solution is to be explicit with the ambiguous *endpoint*, where the latter
-has a format of `<application>:<application endpoint>`. In this case, the
-application endpoint is 'db' for both but only the mediawiki interface is ambiguous, so
-this is what we can do:
+The solution is to be explicit when referring to an *endpoint*, where the
+latter has a format of `<application>:<application endpoint>`. In this case,
+the application endpoint is 'db' for both applications but because only the
+mediawiki endpoint is ambiguous (according to the error message), it is not
+necessary to mention the mysql endpoint:
 
 ```bash
 juju add-relation mediawiki:db mysql
--- a/src/en/charms-relations.md
+++ b/src/en/charms-relations.md
@@ -1,4 +1,4 @@
-Title: Managing relationships  
+Title: Managing relationships
 TODO:  Critical: review required
 
 # Managing relationships
@@ -50,13 +50,13 @@ For example, in the case of specifying a database for the Mediawiki charm.
 
 ```bash
 juju add-relation mediawiki mysql
-error: ambiguous relation: "mediawiki mysql" could refer to 
+error: ambiguous relation: "mediawiki mysql" could refer to
   "mediawiki:db mysql:db"; "mediawiki:slave mysql:db"
 ```
 
-The solution is to be explicit with the ambiguous *interface*, where the latter
+The solution is to be explicit with the ambiguous *endpoint*, where the latter
 has a format of `<application>:<application endpoint>`. In this case, the
-*endpoint* is 'db' for both but only the mediawiki interface is ambiguous, so
+application endpoint is 'db' for both but only the mediawiki interface is ambiguous, so
 this is what we can do:
 
 ```bash
@@ -110,7 +110,7 @@ juju remove-relation mediawiki mysql
 
 In cases where there is more than one relation between the two applications, it
 is necessary to specify the interface at least once:
-  
+
 ```bash
 juju remove-relation mediawiki mysql:db
 ```
--- a/src/en/models-cmr-scene-1.md
+++ b/src/en/models-cmr-scene-1.md
@@ -34,6 +34,9 @@ juju add-model cmr-model
 juju deploy mysql --constraints "cores=4 mem=16G root-disk=1T"
 ```
 
+!!! Note:
+    The created model is optional. The 'default' model could have been used.
+
 The output to `juju status` will eventually look similar to:
 
 ```no-highlight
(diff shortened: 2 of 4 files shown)


src/en/charms-relations.md
```diff
@@ -49,19 +50,27 @@ For example, in the case of specifying a database for the Mediawiki charm.
...
-enter:
+The solution is to be explicit when referring to an *endpoint*, where the
+latter has a format of `<application>:<application endpoint>`. In this case, it
+is 'db' for both applications. However, it is not necessary to specify the
+mysql endpoint because only the mediawiki endpoint is ambiguous (according to
+the error message). Therefore, the command becomes:
 
 ```bash
-juju add-relation mediawiki:db mysql
+juju add-relation mysql mediawiki:db
```
evilnick: why is the order reversed? I mean, it shouldn't matter, but we have always used <from> <to> order in the examples

src/en/charms-relations.md
```diff
@@ -49,19 +50,27 @@ For example, in the case of specifying a database for the Mediawiki charm.
...
-error: ambiguous relation: "mediawiki mysql" could refer to 
+error: ambiguous relation: "mediawiki mysql" could refer to
   "mediawiki:db mysql:db"; "mediawiki:slave mysql:db"
 ```
 
-The solution is to specify the nature of the relation using the
-relation interface identifier. In this case, we want MySQL to provide the 
-backend database for mediawiki ('db' relation), so this is what we need to 
-enter:
+The solution is to be explicit when referring to an *endpoint*, where the
```
evilnick: If we are going to start calling these objects 'endpoints', I think we should actually introduce them a bit more than that

src/en/models-cmr-scene-1.md
```diff
@@ -11,12 +15,12 @@ controller; used by the admin user, and consumed by multiple models.
```
evilnick: Maybe explain the cmr aspect of this in some way first. it is a bit confusing to go straight from this statement to creating a controller...

src/en/models-cmr-scene-1.md
```diff
@@ -27,65 +31,69 @@ servicing several remote applications.
 
 ```bash
 juju add-model cmr-model
-juju deploy mysql --constraints "mem=16G root-disk=1T"
+juju deploy mysql --constraints "cores=4 mem=16G root-disk=1T"
 ```
 
-The output to `juju status` should eventually look similar to:
+!!! Note:
```
evilnick: I don't think this note is very noteworthy

pmatulis: OK, but I wanted to make clear that a special CMR model is not required.

src/en/charms-relations.md
pmatulis: Right, I made this change minimal because it's not strictly a CMR thing, but it really needed changing. I didn't want to get caught in a rabbit hole. You want more of this here or do you want this done in another PR?

src/en/models-cmr-scene-1.md
pmatulis: OK, but it *is* an example scenario, and I *do* refer to the main page for background information.

src/en/models-cmr-scene-1.md
```diff
@@ -94,35 +102,42 @@ juju expose wordpress
 juju relate wordpress:db admin/cmr-model.mysql
```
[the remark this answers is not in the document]
pmatulis: Agreed.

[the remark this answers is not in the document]
pmatulis: Do we keep `juju status` everywhere? It's an alias for `juju show-status`.

src/en/charms-relations.md
evilnick: From the thing doing the giving, to the consumer. wordpress consumes mysql, as does mediawiki

src/en/models-cmr-scene-1.md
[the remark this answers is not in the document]
mitechie: What is an alias vs not isn't that obvious to the user and I think that defaulting to the shortest commands is reasonable.

src/en/models-cmr-scene-1.md
evilnick: I'm not actually sure that is helped by showing people the steps to create a controller and a model. We could just state that we have started with a model called whatever, and leave it to the reader to set up whatever model/controller they want.

src/en/charms-relations.md
evilnick: Maybe we should tack this on in a different pr when we have nailed down the CMR content

src/en/models-cmr-scene-2.md
```diff
@@ -244,104 +257,32 @@ ian:admin/default.hosted-mysql:
       access: read
 ```
 
-To find offers in a specified model:
+To find offers in model `cmr-model-1` on controller `lxd-cmr-1`:
 
 ```bash
-juju find-offers admin/default
-juju find-offers ian:admin/default
+juju find-offers lxd-cmr-1:admin/cmr-model-1
```
evilnick: is it always necessary, when targeting a model, to include the user name too? I mean, it seems to be from my experiments, but is that really true?

I guess what I am saying is that I would assume 

juju find-offers lxd-cmr-1:cmr-model-1 

would work. It doesn't. Am I missing something?

src/en/models-cmr-scene-2.md
```diff
@@ -244,104 +257,32 @@ ian:admin/default.hosted-mysql:
...
-
-### Relating to offers from behind a firewall
-
-Sometimes, the consuming application is deployed behind a firewall where NAT is
-used, such that traffic egresses via a different address/network to that on
-which the consuming application is hosted.
-
-In this case, the relate --via option is used to inform the offering side so
-that the correct firewall rules can be set up.
+This will work providing the offering side did not previously set up a
```
evilnick: I find this paragraph confusing. Does this mean that it will work if no whitelist has been set?

src/en/models-cmr.md
```diff
@@ -113,21 +124,23 @@ These are applied similarly to how standard model access is applied, via the
...
-## Relating to offers
+### Relating to offers
 
 If a user has consume access to an offer, they can deploy an application in
 their model and establish a relation to the offer by way of its URL.
 The controller part of the URL is optional if the other model resides in
 the same controller.
 
-`juju relate <application> <offer url>`
+`juju relate <application>[:<application endpoint>] <offer url>[:<offer endpoint>]`
```
evilnick: add-relation

I suppose in the sense that offers are external objects, doing things this way around *feels* correct. I wonder if it would then make sense to change every other add-relation command in the docs

src/en/models-cmr.md
```diff
@@ -28,12 +17,13 @@ clouds.
...
 model and share them with disparate models. One can imagine models dedicated to
-tasks such as service monitoring, block storage, and database backends.
+tasks such as service monitoring, block storage, and database backends. Another
+use case would be when you are simply using different cloud types.
 
 The commands related specifically to this subject are:
 
 [`consume`][commands-consume]
-: Adds a remote offer to a model.
+: Adds a remote offer to a model without relating to it.
```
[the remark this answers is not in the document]
evilnick: Okay, having read some more I think I understand this better now. 'consume' is like, take up the offer, but don't attach it to anything, save it for later? How do you attach to it later? How do you know it exists?

The comments about the description still stand

src/en/models-cmr.md
```diff
@@ -136,57 +149,45 @@ Note that the relations block in status shows any relevant status information
...
-
-It's possible to consume an offer without relating to it. This creates the
-proxy application in the consuming model, which can then be related to
-afterwards. Doing it this way enables an alias to be used for the offer, if
-there's a need to avoid conflicts with an existing application already deployed
-to the consuming model.
+An offer can be consumed without relating to it. This workflow sets up the
+proxy application in the consuming model and creates a user-defined alias for
+the offer. This latter is what's used to relate to. Having an offer alias can
+avoid a namespsce conflict with a pre-existing application.
```
evilnick: namespace

src/en/models-cmr-scene-2.md
pmatulis: I agree I can make it clearer. The answer to your question is yes, without a whitelist any request for a firewall rule will be accepted.

src/en/models-cmr.md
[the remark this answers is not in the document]
pmatulis: Suggestions welcome :smiley_cat:

src/en/models-cmr.md
pmatulis: I'm not sure how it is more correct by using `relate`. I would like to stay consistent by not using aliases. Maybe I'm missing your point.

src/en/models-cmr.md
pmatulis: OK

src/en/models-cmr.md
evilnick: I meant the order of the application / offer url. Normally we do `add-relation <provider> <consumer>` 

we should definitely use add-relation

src/en/models-cmr.md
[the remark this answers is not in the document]
evilnick: How about : "Accepts an offer from a different model without immediately adding a relation to it. The offered application will subsequently appear as an application in the current model (e.g. in juju status) and a relation can made as with any other native application."

Or something

src/en/models-cmr.md
pmatulis: Ohh

src/en/models-cmr.md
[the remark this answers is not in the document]
pmatulis: That's awfully long!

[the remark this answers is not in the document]
evilnick: Feel free to make it more concise ;)

src/en/models-cmr-scene-2.md
pmatulis: The username pertains to the offering side and I don't see how the consuming side can guess what it is.

pmatulis: Sorry, I misunderstood.
So what you say makes sense and I tried it and it works for me. :rabbit: 
```
juju find-offers lxd-cmr-1:cmr-model-1
Store      URL                      Access  Interfaces
lxd-cmr-1  admin/cmr-model-1.mysql  admin   mysql:db

evilnick: That is even more worrying! I will try again tomorrow

src/en/models-cmr.md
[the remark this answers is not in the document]
evilnick: Adds is a word I had difficulty with here - it could be read as the user is "adding" (creating) the offer rather than consuming it. Also, I don't see why the explanation needs to be one line?

src/en/charms-relations.md
pmatulis: So doesn't my new order follow your policy? You said the giver comes first. I've changed it so that mysql, the giver (doing the giving), comes first.

src/en/models-cmr-scene-1.md
pmatulis: After re-reading it looks OK. The very first sentence says the scenario pertains to CMR (link). Do you really want me to re-defined what CMR is?

src/en/models-cmr-scene-1.md
pmatulis: I feel that due to the various contexts (what controller, what model) it is important to be explicit.

src/en/models-cmr-scene-2.md
pmatulis: This is done, and with less words.

src/en/models-cmr.md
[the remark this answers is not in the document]
pmatulis: Because all the other commands have short descriptors. I'd like to avoid long lines here. I agree that it comes off as abrupt since these commands are listed from the outset, before the reader has a chance to digest anything. This is the same approach I took with [storage](https://github.com/juju/docs/blob/master/src/en/charms-storage.md). I asked for the team's opinion on that and the consensus was that it was fine.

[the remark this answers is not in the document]
pmatulis: What about "Links an offer to a model..."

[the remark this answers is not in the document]
pmatulis: It also doesn't help that the `consume` command is the very first one (due to the alphabetical order).

src/en/models-cmr-scene-1.md
evilnick: I just think the note is distracting. If it is felt to be important to mention you can use any model, you could just mention that when you create the ones above. It isn't a big deal, it just feels like a big interruption for the reader for something minor(is)

pmatulis: I removed the Note.

src/en/models-cmr.md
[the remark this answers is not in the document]
evilnick: What is wrong with "accepts"?

It is fine when the descriptions actually work.

src/en/models-cmr-scene-1.md
evilnick: No, I just think this doesn't start well - it mentions mediawiki briefly and then suddenly you are building a controller (or actually, bizarrely, adding credentials).
I think the whole thing would work a lot better if we actually set out what we were going to do, and lay out some requirements (e.g. In this example we have a controller called xxx and a model called yyy. Nobody cares that it is on aws. Then you can start the actual tutorial with something that matters, like adding mysql.

pmatulis: OK, I thought your comment was for scenario #2. It makes more sense now.

src/en/charms-relations.md
evilnick: Yes, sorry it does now

src/en/charms-relations.md
pmatulis: OK. I've left this text commented.

pmatulis:
I believe all your concerns have been addressed.

evilnick:
Okay, we can probably merge this and continue to refine it.

One thing I think we ought to address is the use (maybe overuse) of CMR as an acronym to define what is going on here. We may call this feature CMR internally, but I'm not sure it makes sense to the user.

PR #2255, opened 2017-10-26
reword and correct config language

### LINKED ISSUE #2244: https://jujucharms.com/docs/2.2/authors-charm-config doc is incorrect about default values
If the default value is left empty as described in this document, proof will complain about it:

"proof: I: config.yaml: option extra_sans has no default value"

Is proof wrong here or is the documentation?

pmatulis:
resolves #2244


src/en/authors-charm-config.md
```diff
@@ -2,25 +2,26 @@ Title: Creating config.yaml and configuring charms
 
 # Charm configuration
 
-The optional `config.yaml` file defines how a service running the charm can be
-configured by its user.
+The optional `config.yaml` file defines how the charm can be configured by the
+user.
```
anastasiamac: Do you think it's worthwhile here making a distinction between different users? Eg. here it'd be a "charm author" which is different to "a user that wishes to deploy this charm".

src/en/authors-charm-config.md
```diff
@@ -2,25 +2,26 @@ Title: Creating config.yaml and configuring charms
...
-  - `default` should contain a value of the appropriate type. If set as
-    `default:` with no trailing characters, it is treated as null (which is
-    always a valid value in this context).
+The following fields are available:
 
-Any option without these three fields will generate a Warning from the the
-[charm proof tool](tools-charm-tools.html#proof)
-indicating the option is not compliant with charm store policy. This policy
-allows older versions of juju to safely unset values.
+ - `type` : can be `string`, `int`, `float`, or `boolean`. The default type is
```
anastasiamac: I am not sure if we need to say "The default type is..".
I mean, providing the type is mandatory when defining an option. And when the type is provided, it cannot be anything but {string, int, float, boolean}. Default implies that if you just say 'type' without one of the 4 types, something will be substituted. Whereas I am pretty sure, we will err out. There is no 'default' as such, is there?

anastasiamac: I think it will be helpful to say here 'Type must be provided" or something to this effect.
And for 'description' and 'default' that they are optional.

src/en/authors-charm-config.md
```diff
@@ -2,25 +2,26 @@ Title: Creating config.yaml and configuring charms
...
+  `string`.
+ - `description` : contains an explanation of what the user might achieve by
+  altering the setting along with valid values.
+ - `default` : contains a default value for the option. This value must be of
+  the appropriate type and be coherent for the given charm. If such a default
+  is mandatory for the charm to run then a value must be supplied. Null
+  values are valid for the `string` type where (`default:` and `default: ''`
```
anastasiamac: Empty (or null values) are valid for any option type. In fact, if the default is not defined in charm's config.yaml, it is expected that the value for this particular option will be provided by the user that will be deploying this charm in application's config.yaml (if this option is necessary for a successful deployment of this charm or successful running of the application).

anastasiamac: When I say 'expected', I do not mean that Juju will check that the value for this option has been specified by the user that is deploying the charm. Juju may have troubles deploying this charm or running its application.
So, it may be worthwhile to add that our charm authors may wish to be explicit when documenting the charm - "these options have default - a, b, c... But these options need to have  values explicitly stated at deployment time - x, y, z..." [because no default values are stated for these options in charm's config.yaml]

src/en/authors-charm-config.md
pmatulis: I chose "by the user" carefully. This, to me, and hopefully the reader, indicates the end user. I had assumed the charm author knew what config.yaml is for. However, after looking at other pages and where this page fits in, I realized this may not be the case.

This page is not found in the menu. Actually, it is quite far down the rabbit hole:

https://jujucharms.com/docs/2.2/authors-intro
https://jujucharms.com/docs/2.2/authors-charm-components
https://jujucharms.com/docs/2.2/authors-charm-config

I'll reword. But it's subtle because this file is available to the end user if they choose to download and inspect the charm, for the exact purpose of discovering how it can be configured, and this is the file that shows that to them, the end user.

src/en/authors-charm-config.md
pmatulis: Agreed.

src/en/authors-charm-config.md
pmatulis: Again, more subtlety. Juju itself may not require certain values in order to *attempt* to deploy but the charm may require them.

pmatulis: How does the end user provide an option value at deploy time? Can it only be done through a separate yaml file (`juju deploy --config myconfig.yaml mysql`)?

jameinel:
you can also specify --config key=value, pairs one by one.

pmatulis:
@jameinel Interesting. This is not mentioned in `juju help deploy` nor anywhere in our docs.

anastasiamac:
@pmatulis,

Thank you for adding clarity to this page \o/
Yes, please open a bug for help doc on the command to clarify --config input.

As for making the distinction between different users based on the function/operation they perform, like 'charm author' vs 'a user that deploys a charm', it's fairly standard. It brings in extra context for the reader :)  Of course, ideally we'd find a shorter name for the latter user, although 'charm deployer' is the best that I could come with and it feels rough...

PR #2278, opened 2017-11-06
Fix typo in juju-coexist install example

szeestraten-ntnu:
Fixes a typo in the juju-coexist install example.


pmatulis:
Thank you. :scorpion:

pmatulis:
I'm wondering whether we should not tack on `sudo` here.

szeestraten-ntnu:
It might be for a different issue, but the document might also need another look to determine if it's still relevant and/or if it should be updated for installing via snaps.

PR #2282, opened 2017-11-07
Tweak relation id to include endpoint name prefix

wallyworld:
relation ids may optionally include the endpoint name as a prefix.
Mention both formats in the doc here.

https://bugs.launchpad.net/juju/+bug/1728553


pmatulis:
resolves #2283

PR #2285, opened 2017-11-07
remove container constraint

### LINKED ISSUE #1369: Constraints Pages Updates
Just a list of notes on further cleanups on the constraints page:
- It’s cpu-cores, not core(s). We should update the doc to be consistent.
- ECU is no longer valid; they use vCPU now. Can this be updated? Can we talk about what it might mean for other providers?
- Is it possible to use the container constraint in a substrate besides GCE?
- Arch in Manual Provider is misspelled: “arhc”
- https://jujucharms.com/docs/2.0/reference-constraints isn’t the same as https://jujucharms.com/docs/devel/reference-constraints. The former seems to be out of date? Why are they different?

@anastasiamac adding you to confirm and help with the questions.

pmatulis:
resolves #1369 

Call from Juju team to remove this constraint.

### DIFF
--- a/src/en/reference-constraints.md
+++ b/src/en/reference-constraints.md
@@ -43,12 +43,6 @@ multiple constraints delimited by a space.
     blank to indicate any architecture is acceptable, or one of `amd64`,
     `arm`, `i386`, `arm64`, or `ppc64`.
 
-- container
-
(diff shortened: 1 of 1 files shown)


evilnick:
Hurrah!

PR #2291, opened 2017-11-07
Removed outdated authors-hook-debug* pages

galgalesh:
- `authors-hook-debug` is completely replaced by `developer-debugging`
- `authors-hook-debug-dhx` is completely replaced by `developer-debug-dhx`.

I removed the two pages and pointed all links to the updated pages. I also went through AskUbuntu to point all links to the new pages.

*Ps: sorry for all the noise in the commit message, Atom automatically fixes whitespace to conform to Unix guidelines (no whitespace before newline, one newline at the end of a file)*


pmatulis:
I have a feeling this does not apply to 1.25. If that's the case, then we'll lose image `authors-hook-debug-1.png` that's being referenced from 1.25:

```
src/en/authors-hook-debug.md:  ![Image showing tmux debug hooks session](./media/authors-hook-debug-1.png)
```

If this is the case, the image rename in this PR should really be the creation of a new image.

It's not obvious, but we use a central image directory in `master` that is used for all releases.

galgalesh:
@pmatulis 

This PR is for 2.x only.

Thanks, I didn't know about the central image directory, I re-added the image.

PR #2305, opened 2017-11-17
Explain primary-network, external-network and datastore options

galgalesh:
While trying to get Juju working on a VMWare cluster and digging through the VMWare provider code, I found three crucial vmware-specific config options:

- **`primary-network`**: The primary network that VMs will be connected to. If
  this is not specified, the first accessible network will be used.
- **`external-network`**: The name of an additional "external" network to which
  the VM should be connected. The IP from this network will be used as the
  `public-address` of the VM's.
- **`datastore`**: Datastore is the name of the datastore in which to create
  the VM. If this is not specified, the first accessible datastore will be used.

This PR adds an explanation for those options to the docs.

Applies to Juju 2.x

### DIFF
--- a/src/en/help-vmware.md
+++ b/src/en/help-vmware.md
@@ -81,7 +81,7 @@ juju bootstrap myvscloud myvscontroller
 ```
 
 There are three VMware-specific options you can use to specify the network and
-datatore to use.
+datastore to use.
 
 - **`primary-network`**: The primary network that VMs will be connected to. If
   this is not specified, the first accessible network will be used.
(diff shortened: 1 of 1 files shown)


galgalesh:
@degville ping?

src/en/help-vmware.md
```diff
@@ -70,18 +70,48 @@ juju add-credential myvscloud
...
 
 With credentials added, you can now start using Juju with your vSphere cloud:
 
 ```bash
 juju bootstrap myvscloud myvscontroller
 ```
 
-!!! Note: 
+There are three VMware-specific options you can use to specify the network and
+datatore to use.
```
degville: datastore

src/en/help-vmware.md
```diff
@@ -70,18 +70,48 @@ juju add-credential myvscloud
...
+You can specify these options during bootstrap:
+
+```bash
+juju bootstrap myvscloud myvscontroller --config primary-network=PRIMARY_NET --config external-network=EXTERNAL_NET --config datastore=NFSSTORE
+```
+
+!!! Note:
+    When you specify these options in the bootstrap command, they will only
+    apply to the `controller` and `default` model. Use
+    [`model-defaults`](https://jujucharms.com/docs/2.2/models-config) if you
```
degville: Could you update this link to refer to `models-config` rather than the full URL (to avoid the specific version link)?

galgalesh:
Fixed!

PR #2306, opened 2017-11-21
fan networking

### LINKED ISSUE #2226: Docs needed: LXD fan networking
From the 2.3 beta1 Release Notes

*****

### FAN networking in containers (initial support)

A new "container-networking-method" model config attribute is introduced with 3
possible values: "local", "fan", "provider".
* local = use local bridge lxdbr0
* provider = containers get their IP address from the cloud via DHCP
* fan = use FAN

The default is to use "provider" if supported. Otherwise, if FAN is configured
use that, else "local".
On AWS, FAN works out of the box. For other clouds, a new fan-config model
option needs to be used, eg

    juju model-config fan-config="<underlay1>=<overlay1> <underlay2>=<overlay2>

pmatulis:
Resolves #2226 :pig: 

For the table on models-config.md, most of the noise is due to columns being resized. The only factual change is the addition of options `fan-config` and `container-networking-method`.


src/en/charms-fan-aws-vpc.md
```diff
@@ -0,0 +1,65 @@
+Title: Creating an AWS VPC
+
+# Creating an AWS VPC
+
```
wupeka: It's worth noting that, as of today, a default VPC will be created automatically with users' account.

src/en/charms-fan-aws-vpc.md
```diff
@@ -0,0 +1,65 @@
...
+    In the resulting dialog box, enter a name for the subnet (e.g.
+    'vpc-juju-subnet'), select the VPC you created earlier, an Availability
+    Zone (optional), and an IPv4 CIDR block that resides within the initial
+    address space (e.g. '192.168.1.0/24').
+
+    Select the subnet's checkbox and hit `Subnet Actions`. There, choose
+    `Modify auto-assign IP settings` and then select `Enable auto-assign public
+    IPv4 address`.
+
```
wupeka: For complete HA - create one subnet for each Availability Zone.

src/en/charms-fan.md
```diff
@@ -0,0 +1,202 @@
...
+that was hitherto not possible. The typical use case is the seamless
+interaction between deployed applications running within LXD containers on
+separate Juju machines.
+
+## FAN overview
+
+The FAN is a mapping between a smaller IPv4 address space (e.g. a /16 network)
+and a larger one (e.g. a /8 network) where **subnets** from the larger one (the
+*underlay* network) are assigned to **addresses** on the smaller one (the
+*overlay* network). Connectivity between containers on the larger network is
```
wupeka: Underlay is the smaller (/16) one, overlay is the larger one (/8). I always mess them up, https://wiki.ubuntu.com/FanNetworking has a nice image to explain.

src/en/charms-fan.md
```diff
@@ -0,0 +1,202 @@
...
+following values:
+
+ - local : standard LXD; addressing based on the LXD bridge (e.g. lxdbr0)
+ - provider : addressing based on host bridge; works only with providers with
+   built-in container addressing support (e.g. MAAS with LXD)
+ - fan : FAN networking; works with any provider, in principle
+
+Once FAN is enabled, by setting the above option to 'fan', all that is needed
+is to map the underlay network to the overlay network. The `fan-config` model
+option is used for this and has the following format:
```
wupeka: fan-config has to be set prior to setting container-networking-method to fan.

src/en/charms-fan.md
```diff
@@ -0,0 +1,202 @@
...
+
+  `<underlay-network>=<overlay-network>`
```
wupeka: It can be a space-separated list of multiple mappings: "<underlay-network1>=<overlay-network1> <underlay-network2>=<overlay-network2>"

src/en/charms-fan.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+In this example, the underlay network is 10.0.0.0/16 and the overlay network is
+252.0.0.0/8.
+
+See [Configuring models][models-config] for how to set model options.
+
+## Cloud provider requirements
+
+Juju autoconfigures FAN networking for both the AWS and GCE clouds. All that
```
wupeka: on AWS only in a VPC, there still is legacy mode in which we won't autoconfigure fan.

src/en/charms-fan.md
pmatulis: Me too. Thanks.

src/en/charms-fan-aws-vpc.md
pmatulis: OK

src/en/charms-fan-aws-vpc.md
pmatulis: I mentioned it on the main page but I can repeat here I suppose.

src/en/charms-fan.md
pmatulis: Ohh.

src/en/charms-fan.md
pmatulis: OK

src/en/charms-fan.md
pmatulis: I do mention later that if the user has an old AWS account the VPC may be deficient. I may reword.

src/en/charms-fan-aws-vpc.md
pmatulis: Actually, I don't quite get this comment. HA also requires manipulation of the applications. Unless I'm missing something, I'd rather not get into that.

src/en/charms-fan-aws-vpc.md
```diff
@@ -0,0 +1,69 @@
...
+# Creating an AWS VPC
+
+This is in connection with [Juju and FAN networking][charms-fan] where an AWS
+user may wish to create an additional VPC (virtual private network).
+
+Note that only older AWS accounts will actually require a new VPC in order for
+Juju to autoconfigure the FAN. Recent AWS accounts are furnished with a VPC
+which will work out-of-the-box.
+
+The procedure covered here refers to the AWS web interface. Presumably it is
```
evilnick: I would omit the part about doing it with the API. Just saying presumably is a little un-authoratative.

src/en/charms-fan.md
```diff
@@ -0,0 +1,207 @@
+Title: Juju and FAN networking
+TODO:  bug tracking: https://bugs.launchpad.net/juju/+bug/1733354 (remove constraints Note if no longer required)
+       hardcoded: default AWS instance type of 'm3.medium'
+       consider a third example: manually configure the FAN with Azure
+
+# Juju and FAN networking
+
```
evilnick: This could maybe use a few words of what we mean by 'FAN'. Is it an acronym? Is it, indeed, FAN and not Fan, as it appears elsewhere

src/en/charms-fan-aws-vpc.md
pmatulis: I'll reword and include a [link](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVpc.html).

src/en/charms-fan.md
pmatulis: I'll change to Fan.

Issue #206, opened 2014-10-30
Offline-mode documentation


niedbalski:
Hello,

The only publicly available reference I can locate for doing off-line Juju deployments is https://juju.ubuntu.com/docs/howto-offline-charms.html. 

However, there are a few other things we also should be covering on this document:

1) How to synchronize the juju tools locally.
2) Setting up a package mirroring using apt-proxy for packages update/install
3) Configure squid or http/https proxy
4) How to deal with charms that requires internet access for fetching arbitrary resources.
5) Step by step example of an off-line deployment.

kwmonroe:
This issue is super old, but still relevant.  In fact, we've seen lots more people in `#juju` asking about proxy settings for juju since 2.0 was released (eg, #1676).  I'd like to propose something like the following for an offline/restrictive juju environment doc:

---
Juju works in restricted network environments and honors various `proxy` settings (see `*-proxy` options [here](https://jujucharms.com/docs/2.0/models-config#list-of-model-keys).  There are 2 places to configure proxies when bootstrapping juju:

- The bootstrap unit itself may need proxy information to retrieve various packages.  This is configured with `--config` options during bootstrap.

- Models may also need this proxy information if charms deployed in those models require external network access (for example, access to pypi to install requisite pip packages).  This is configured with `--model-default` options during bootstrap.

As an example, consider an environment where all network traffic must go through an `http://squid.internal:3128` proxy.  We plan to deploy charms locally with lxd, so we need to ensure all inter-container traffic is *not* proxied.  This can be achieved as follows:

```bash
## no proxy for localhost, our eth0 ip address, and our lxd subnet
$ export no_proxy=`echo localhost 10.245.67.130 10.44.139.{1..255} | sed 's/ /,/g'`
$ export http_proxy=http://squid.internal:3128
$ export https_proxy=http://squid.internal:3128
```

And now, bootstrap juju with appropriate proxy configuration:

```bash
$ juju bootstrap \
--config http-proxy=$http_proxy \
--config https-proxy=$https_proxy \
--config no-proxy=$no_proxy \
--model-default http-proxy=$http_proxy \
--model-default https-proxy=$https_proxy \
--model-default no-proxy=$no_proxy \
localhost
```
---

I would make a formal PR for the above, but I'm not sure where it would fit best.  Perhaps as a section in the `models-config` page?  Maybe on its own page? Wadda you think @mbruzek  @evilnick @niedbalski?

evilnick:
[...]

PR #2308, opened 2017-11-25
rewrite offline charms from 1.25 doc and rationalize existing pieces

pmatulis:
affects #206 

 - **Most** of this can be reused to update the [1.25 doc](https://github.com/juju/docs/blob/1.25/src/en/juju-offline-charms.md) upon which this PR was based. An equivalent page was non-existent in 2.x. Used some disparate pieces from charms-deploying.md.
 - **Some** of this (opening section) may be more appropriate in an eventual general offline document, which will link to this new one, as well as other upcoming new ones.


pmatulis:
Re 'charm names' IKWYM. Maybe we should put them in monospaced (backticks).

PR #2309, opened 2017-11-27
Fix a litle typo on 'About-juju' document

obedmr:
Signed-off-by: Obed N Munoz <obed.n.munoz@gmail.com>


obedmr:
I found this https://en.wiktionary.org/wiki/organisation and this may be the reason why it was written like this.

PR #2316, opened 2017-12-05
add recycled machines

### LINKED ISSUE #2315: Deploying bundles can now target existing machines
> When deploying bundles, machines specified in the bundle are added to the model as new machines. In order to use the existing machines in the model rather than create new machines, the --use-existing-machines flag can be used. To specify a particular machine, the --bundle-machine flag can be used to specify a specific machine for  a particular machine in the bundle. These flags can be combined with the --bundle-machine flag taking precidence. For example, if there was a bundle that specified machines 1, 2, and 3, and the model had machines 1, 2, 3 and 4, the following deployment of the bundle would use machines 1 and 2 in the model for machines 1 and 2 in the bundle and use machine 4 in the model for the bundle machine 3.
    juju deploy some-bundle --use-existing-machines --bundle-machine 3=4

pmatulis:
resolves #2315

### DIFF
--- a/src/en/charms-bundles.md
+++ b/src/en/charms-bundles.md
@@ -380,6 +380,24 @@ mysql/1      waiting   allocating  1                               waiting for m
 wordpress/0  waiting   allocating  1                               waiting for machine
 ```
 
+### Recycling machines
+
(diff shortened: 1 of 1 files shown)


evilnick:
:cucumber:

PR #2320, opened 2017-12-05
add overlay bundle

### LINKED ISSUE #2314: Bundle deploy flag --bundle-config replaced with --overlay
> The bundle config files used to allow overwriting configuration values of applications that had already been defined in the base bundle, but did nothing else.
The overlay files are now full bundles in their own right. The overlay bundle can specify new applications, change values like the previous bundle-config files, and also specify the removal of an application in the base bundle. An application can be removed from the bundle definition by defining the application name in the application section, but no values specified. Removing an application also removes all the relations for that application. If a machines section is specified in an overlay bundle it replaces the machines section of the base bundle. No merging of machine information is attempted. Multiple overlay bundles can be specified and they are processed in the order they appear on the command line.

pmatulis:
resolves #2314

### DIFF
--- a/src/en/charms-bundles.md
+++ b/src/en/charms-bundles.md
@@ -243,36 +243,28 @@ applications:
             cert: include-base64://my-cert.crt
 ```
 
-Furthermore, the `--bundle-config` option can be used when you want to use a
-standard bundle but keep model-specific configuration in a separate file.  Any
-values specified for an application in the file override those values defined
-in the bundle, with the exception of the map type values, where the maps are
-merged with preference given to the bundle-config file. The file can make use
-of the `include-file://` and `include-base64://` directives mentioned above for
-local bundles (for options and annotation values). A file location can be
-expressed with an absolute or relative (to the bundle-config file) path. For
-example:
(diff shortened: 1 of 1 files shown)


src/en/charms-bundles.md
```diff
@@ -243,36 +243,28 @@ applications:
...
-Furthermore, the `--bundle-config` option can be used when you want to use a
-standard bundle but keep model-specific configuration in a separate file.  Any
-values specified for an application in the file override those values defined
-in the bundle, with the exception of the map type values, where the maps are
-merged with preference given to the bundle-config file. The file can make use
-of the `include-file://` and `include-base64://` directives mentioned above for
-local bundles (for options and annotation values). A file location can be
-expressed with an absolute or relative (to the bundle-config file) path. For
-example:
+## Overlay bundles
```
evilnick: Hmm, okay. I think at some point this needs a proper worked example - e.g. a bundle, an example overlay, and what you would end up with. Maybe you could add that as a TODO

pmatulis: Done.

PR #2394, opened 2018-01-23
Review concepts page

### LINKED ISSUE #2389: Review needed: juju-concepts.md
As a follow up to #2368, this issue is about conducting an actual review of the textual content of [juju-concepts.md](https://github.com/juju/docs/blob/master/src/en/juju-concepts.md).

pmatulis:
resolves #2389

### DIFF
--- a/src/en/juju-concepts.md
+++ b/src/en/juju-concepts.md
@@ -8,9 +8,9 @@ To Juju, a *cloud* (or backing cloud) is a resource which provides machines
 (instances), and possibly storage, in order for application units to be
 deployed upon them. This includes public clouds such as Amazon Web Services,
 Google Compute Engine, and Microsoft Azure as well as private OpenStack-based
-clouds. Juju can also make use of environment which are not clouds per se, but
-which Juju can treat as a cloud. [MAAS][maas] and [LXD][lxd] fit into this
-latter category.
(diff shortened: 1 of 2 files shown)


src/en/juju-concepts.md
```diff
@@ -1,31 +1,66 @@
...
+The Juju *controller* is the initial cloud instance which is created in order
+for Juju to gain access to a cloud. It is created by having the Juju client
+contact the cloud's API. The controller is a central management node for the
+chosen cloud, taking care of all operations requested by the Juju client.
+Multiple clouds (and thus controllers) are possible and each one may contain
+multiple models and users.
+
+For more information see [Controllers][controllers].
+
+## Client
```
evilnick: I don't think this is the right place for client or agent - either put them both at the top or both at the bottom

pmatulis: I put the client immediately after controller because the essence of the client lies in its connection to the controller.
Likewise, I wanted to put the agent as close as possible to the controller as it also connects to the controller. I put it after model because that concept brought in the idea of a machine (where the agent runs).

pmatulis:
@evilnick Can you approve this please? I put the client and the agent at the bottom as requested.

PR #2395, opened 2018-01-23
add troubleshooting entry to failed juju upgrade

### LINKED ISSUE #2328: how to wipe / reinstall juju once everything goes south
Q:  is there anyway to completly wipe a juju controller/env?   I made the mistake of trying to upgrade juju version - which corrupted the env.   while trying to recover/remove/clean I just released the bootstrap node and uninstalled juju apt-get remove juju.    then tried to start from scratch - install juju,etc.   the problem is there are still remnants of the old system.   I just want to wipe and start from scratch.   juju kill-controller, destroy-controller all fail.

Docs found this on the web also doe not work  "sudo apt-get purge --auto-remove juju"

what I found that works is to:

juju unregister <controller-name>
juju add-cloud  .... maas, etc using the same name as unregister. this will overwrite the prev conf
juju bootstrap <controller-name>  works

pmatulis:
resolves #2328

### DIFF
--- a/src/en/troubleshooting-upgrade.md
+++ b/src/en/troubleshooting-upgrade.md
@@ -94,3 +94,21 @@ juju upgrade-juju --ignore-agent-versions
 !!! Note:
     The flag `--ignore-agent-versions` is only available starting with Juju
     2.2.6.
+
+
+## Case #4 - Recovering from an upgrade failure
+
+If an attempted upgrade results in failure it may prove difficult to return to
+a working setup. The first remediation plan is to remove the controller with
+the `juju destroy-controller` or `juju kill-controller` commands. If this is
+insufficient one way forward is to unregister the controller, add the
+respective cloud, and then create a new controller. Doing this will not remove
+the original controller so there will remnants of it in the database. This
+method should therefore be used as a last resort. Here are the relevant
+commands:
+
+```bash
+juju unregister <controller>
+juju add-cloud --replace <cloud>
+juju bootstrap <cloud>
+```
--- a/src/en/troubleshooting-upgrade.md
+++ b/src/en/troubleshooting-upgrade.md
@@ -96,19 +96,16 @@ juju upgrade-juju --ignore-agent-versions
     2.2.6.
 
 
-## Case #4 - Recovering from an upgrade failure
+## Case #4 - Dealing with an upgrade failure
 
 If an attempted upgrade results in failure it may prove difficult to return to
-a working setup. The first remediation plan is to remove the controller with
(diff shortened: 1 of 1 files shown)


src/en/troubleshooting-upgrade.md
```diff
@@ -94,3 +94,21 @@ juju upgrade-juju --ignore-agent-versions
 !!! Note:
     The flag `--ignore-agent-versions` is only available starting with Juju
     2.2.6.
+
+
+## Case #4 - Recovering from an upgrade failure
+
+If an attempted upgrade results in failure it may prove difficult to return to
+a working setup. The first remediation plan is to remove the controller with
```
evilnick: I think this probably warrants some further detail - i.e. this might get Juju working again, but will lose anything in that cloud for sure. Also 'remediation plan'  doesn't sound very 'plain English'

src/en/troubleshooting-upgrade.md
```diff
@@ -94,3 +94,21 @@ juju upgrade-juju --ignore-agent-versions
...
+the `juju destroy-controller` or `juju kill-controller` commands. If this is
+insufficient one way forward is to unregister the controller, add the
+respective cloud, and then create a new controller. Doing this will not remove
+the original controller so there will remnants of it in the database. This
```
evilnick: will be

src/en/troubleshooting-upgrade.md
```diff
@@ -94,3 +94,21 @@ juju upgrade-juju --ignore-agent-versions
...
     2.2.6.
+
+
+## Case #4 - Recovering from an upgrade failure
+
+If an attempted upgrade results in failure it may prove difficult to return to
+a working setup. The first remediation plan is to remove the controller with
+the `juju destroy-controller` or `juju kill-controller` commands. If this is
+insufficient one way forward is to unregister the controller, add the
+respective cloud, and then create a new controller. Doing this will not remove
```
evilnick: why do you need to add the cloud again? cloud details are stored separate from controllers

src/en/troubleshooting-upgrade.md
pmatulis: Right, I can underline that the original controller is not actually being recovered.

OK, for 'remediation plan'.

src/en/troubleshooting-upgrade.md
pmatulis: To overwrite any settings for that cloud set by the borked controller.

src/en/troubleshooting-upgrade.md
pmatulis: :ok:

PR #2444, opened 2018-02-13
audit logging

### LINKED ISSUE #2435: documentation needed for audit logging
Audit logging has been rewritten - we used to have some code to do it but it wasn't really suitable, and it wasn't documented as far as I can tell. 

(I'm not sure how much detail to put into this, I haven't raised a docs bug before.)

The audit log records CLI commands by users that change things in the controller or models, and the API calls that are generated by those commands (as well as any errors that come back from the API).

It's driven by these controller config options (from [controller/config.go](https://github.com/juju/juju/blob/develop/controller/config.go#L34)):
- `auditing-enabled`
- `audit-log-capture-args`
- `audit-log-exclude-methods`
- `audit-log-max-backups`
- `audit-log-max-size`

It writes json log entries to `/var/log/juju/audit.log`, with file rotation controlled by the `max-size` and `-backups` options.

The log contains [records](https://github.com/juju/juju/blob/develop/core/auditlog/auditlog.go#L95) which are either:
- a [Conversation](https://github.com/juju/juju/blob/develop/core/auditlog/auditlog.go#L24), which corresponds to a top-level CLI command, 
- a [Request](https://github.com/juju/juju/blob/develop/core/auditlog/auditlog.go#L47) (a call to an API method)
- a [ResponseErrors](https://github.com/juju/juju/blob/develop/core/auditlog/auditlog.go#L70), which has details of any errors from calling the method.

I'm babbageclunk on irc as well if you need more details.

pmatulis:
resolves #2435

### DIFF
--- a/src/en/controllers-config.md
+++ b/src/en/controllers-config.md
@@ -95,7 +95,15 @@ is the special value of 'ReadOnlyMethods'. As the name suggests, this
 represents all read-only events. Hence, by default, only events involving a
 change (a write) will appear in the audit log.
 
-An example value for this key is: `ReadOnlyMethods,????`.
+An example value for this key is: `[ReadOnlyMethods,Pinger.Ping]` where a log
+message intended for removal (by including 'Pinger.Ping' as a value) contains
+the following text:
+
+```no-highlight
+"request-id":4428,"when":"2018-02-12T20:03:45Z","facade":"Pinger","method":"Ping","version":1}}
+```
+
+Only those Conversations containing *all* excluded methods will be omitted.
 
 Click the triangle below to reveal a listing of API methods denoted by the
 key value of 'ReadOnlyMethods'. 
--- a/src/en/controllers-config.md
+++ b/src/en/controllers-config.md
@@ -1,7 +1,8 @@
 Title: General configuration options
-TODO: Check accuracy of key table
-      error: table's default value keys do not show up with controller-config (e.g. bootstrap-)
+TODO: Check accuracy of key table (https://github.com/juju/juju/blob/ec89c99e51fa83cd1a2cb5e5f24e73d5b096de20/controller/config.go#L29)
+      error: table's default value keys do not show up with controller-config (e.g. bootstrap-). See above note.
       "dynamically set by Juju" could use some explaination
+      ReadOnlyMethods updated from https://github.com/juju/juju/blob/2.3/apiserver/observer/auditfilter.go#L130
 
 
 # Configuring controllers
@@ -35,7 +36,8 @@ create and configure a cloud named 'lxd':
 juju bootstrap --config bootstrap-timeout=700 lxd
 ```
 
-Once a controller is created, all its settings become immutable.
+In general, once a controller is created, all its settings become immutable. An
+exception to this rule are the five keys related to audit logging.
 
 !!! Note:
     The `--config` option may also be used to configure the 'default' model.
@@ -47,10 +49,14 @@ Once a controller is created, all its settings become immutable.
 
 This table lists all the controller keys which may be assigned a value.
 
-| Key                        | Type   | Default  | Valid values             | Purpose |
-|:---------------------------|--------|----------|--------------------------|:---------|
-api-port                     | integer | 17070   |                          | The port to use for connecting to the API
-auditing-enabled             | bool   | false    | false/true               | Sets whether the controller will record auditing information
+| Key                        | Type    | Default  | Valid values             | Purpose |
+|:---------------------------|---------|----------|--------------------------|:---------|
+api-port                     | integer | 17070    |                          | The port to use for connecting to the API
+auditing-enabled             | bool    | false    | false/true               | Sets whether audit logging is enabled.
+audit-log-capture-args       | bool    | false    | false/true               | Sets whether the audit log will contain the arguments passed to API methods.
+audit-log-exclude-methods    | string  | ReadOnlyMethods |                   | What information to exclude from the audit log. See [additional info][excluding-information-from-the-audit-log].
+audit-log-max-backups        | integer | 10       |                          | The maximum number of backup audit log files to keep.
+audit-log-max-size           | integer | 300      |                          | The maximum size for an audit log file (units: MiB).
 autocert-dns-name            | string |          |                          | Sets the DNS name of the controller. If a client connects to this name, an official certificate will be automatically requested. Connecting to any other host name will use the usual self-generated certificate.
 autocert-url                 | string |          |                          | Sets the URL used to obtain official TLS certificates when a client connects to the API. By default, certificates are obtained from LetsEncrypt. A good value for testing is "https://acme-staging.api.letsencrypt.org/directory".
 allow-model-access           | bool   |          | false/true               | Sets whether the controller will allow users to connect to models they have been authorized for even when they don't have any access rights to the controller itself.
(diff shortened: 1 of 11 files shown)


src/en/troubleshooting-logs-audit.md
```diff
@@ -0,0 +1,28 @@
+Title: Audit logging
+
+# Audit logging
+
+Juju audit logging provides a chronological account of all events. These logs
```
babbageclunk: It captures all commands run by users (or scripts talking to the API) - it doesn't contain API calls by Juju agents. It might be clearer to say "all user commands" here?

pmatulis: :ok:

src/en/controllers-config.md
```diff
@@ -68,8 +78,91 @@ mongo-memory-profile         | string | low      | low/default              | Se
...
+    Conversation contains several read-only events and a single write event
+    then all these events will appear in the log.
+
+Click the triangle below to reveal a listing of API methods designated by the
+key value of 'ReadOnlyMethods'. 
+
+^# ReadOnlyMethods 
+
+  ```
+  Action.Actions
```
evilnick: Why is this list in random order?

src/en/troubleshooting-logs-audit.md
```diff
@@ -0,0 +1,28 @@
+Title: Audit logging
+
+# Audit logging
+
+Juju audit logging provides a chronological account of all events by capturing
+invoked user commands. These logs reside on the controller involved in the
+transmission of commands affecting the Juju client, Juju machines, and the
+controller itself.
+
+See [Juju log][logs] documentation to learn about standard Juju logging.
```
evilnick: I wouldn't put this here. In fact, it would be best to omit all together and include all this stuff on the logging page

src/en/troubleshooting-logs-audit.md
```diff
@@ -0,0 +1,28 @@
...
+
+The audit log filename is `/var/log/juju/audit.log` and contains records which
```
evilnick: Would be useful to outline how to access/read this log.

src/en/controllers-config.md
```diff
@@ -68,8 +78,91 @@ mongo-memory-profile         | string | low      | low/default              | Se
...
+this topic.
+
+Excluding information from the audit log is done via the
+`audit-log-exclude-methods` key above, which refers to API calls/methods. The
+recommended approach for configuring the filter is to view the log and make a
+list of those calls deemed undesirable. There is no definitive API call list
+available in this documentation.
+
+The default value of key `audit-log-exclude-methods` is the special value of
+'ReadOnlyMethods'. As the name suggests, this represents all read-only events.
```
evilnick: This doesn't explain why we would care about that. What other types of events are there?

src/en/controllers-config.md
```diff
@@ -68,8 +78,91 @@ mongo-memory-profile         | string | low      | low/default              | Se
...
+### Excluding information from the audit log
+
+See [Audit logging][troubleshooting-logs-audit] for background information on
+this topic.
+
+Excluding information from the audit log is done via the
+`audit-log-exclude-methods` key above, which refers to API calls/methods. The
+recommended approach for configuring the filter is to view the log and make a
+list of those calls deemed undesirable. There is no definitive API call list
+available in this documentation.
```
evilnick: is there one anywhere? This just seems lame.

src/en/controllers-config.md
```diff
@@ -68,8 +78,91 @@ mongo-memory-profile         | string | low      | low/default              | Se
...
+
+The default value of key `audit-log-exclude-methods` is the special value of
+'ReadOnlyMethods'. As the name suggests, this represents all read-only events.
+
+An example value for this key is: `[Pinger.Ping]` where a log message intended
```
evilnick: This explanation seems completely back to front. We just told people they should examine the current log to see wht things to exclude, so start with the example log message, then explain how to remove it

src/en/controllers-config.md
```diff
@@ -68,8 +78,91 @@ mongo-memory-profile         | string | low      | low/default              | Se
...
+  Storage.ListVolumes
+  Storage.ListFilesystems
+  Subnets.ListSubnets
+  ```
 
 <!-- LINKS -->
 
 [controllers-creating]: ./controllers-creating.html "Creating a controller"
 [models-config]: ./models-config.html "Configuring models"
+[anchor__excluding-information-from-the-audit-log]: #excluding-information-from-the-audit-log
```
evilnick: I am a big anti-fan of this new pattern you are making up. If you want to differentiate an internal link, why not just use '#'?

src/en/troubleshooting-logs-audit.md
```diff
@@ -0,0 +1,28 @@
+Title: Audit logging
+
+# Audit logging
```
evilnick: I don't understand why this has been made a tiny page on its own and not part of the logging page (which it even tells you to read)

src/en/controllers-config.md
```diff
@@ -68,8 +78,91 @@ mongo-memory-profile         | string | low      | low/default              | Se
...
+An example value for this key is: `[Pinger.Ping]` where a log message intended
+for removal (by including 'Pinger.Ping' as a value) contains the following
+text:
+
+```no-highlight
+"request-id":4428,"when":"2018-02-12T20:03:45Z","facade":"Pinger","method":"Ping","version":1}}
+```
+
+!!! Important:
+    Only those Conversations whose methods have *all* been excluded will be
```
evilnick: Sadly, you have defined 'Conversation' on a different page

src/en/controllers-config.md
pmatulis: It is the order in the source code. This makes it easy to update.

src/en/controllers-config.md
pmatulis: No list available. It might appear more lame if we don't say anything.

src/en/troubleshooting-logs-audit.md
pmatulis: I didn't do that because the regular log page is long. I also already refer readers to a separate page for remote logging.

src/en/controllers-config.md
pmatulis: It has to be defined somewhere. I can repeat maybe.

src/en/controllers-config.md
pmatulis: :ok:

src/en/controllers-config.md
pmatulis: :ok:

src/en/troubleshooting-logs-audit.md
pmatulis: :ok:

src/en/controllers-config.md
pmatulis: Writeable events. I would have to trawl the log to get individual thingies (methods?). Suggestions on how to express this?

src/en/troubleshooting-logs-audit.md
pmatulis: Done. I also moved remote logging.

pmatulis:
@evilnick Sorry, but I encountered some branch troubles with this PR and needed to submit a separate one. Please see #2483. I also moved the remote logging info to the central logs page.

PR #2044, opened 2017-09-01
review, corrections, image & output updates

pmatulis:
Command output and images still need updating in master. Issue raised #2043.

PR #2446, opened 2018-02-14
fix add manual dialog

### LINKED ISSUE #2445: manual add-cloud documentation is incorrect
Original: https://github.com/CanonicalLtd/jujucharms.com/issues/519

The fold-out for "Manual" clouds is a duplicate of "MAAS":

https://jujucharms.com/docs/stable/clouds#specifying-additional-clouds

For juju-2.3.2 at least, the exchange should look like this:

$ juju version
2.3.2-highsierra-amd64

$ juju add-cloud
Cloud Types
  maas
  manual
  openstack
  oracle
  vsphere

Select cloud type: manual

Enter a name for your manual cloud: mycloud

Enter the controller's hostname or IP address: <sudo_user>@<ip>

Cloud "mycloud" successfully added
You may bootstrap with 'juju bootstrap mycloud'
You'll notice I put in <sudo_user>@<ip> for the hostname. It is important that the user know that Juju needs to be able to ssh over to the machine with an account that has sudo privileges. I'm not sure the best way to present this requirement in this doc; feel free to change <sudo_user>@<ip> in whatever way makes this more clear.

pmatulis:
resolves #2445 

A wording refresh at the bottom of the Manual entry for add-cloud. I intend to use this wording in the other entries during an upcoming review of the clouds.md page.

An erroneous admonishment was removed from the credentials.md page (source: PR #2044).

### DIFF
--- a/src/en/clouds.md
+++ b/src/en/clouds.md
@@ -1,6 +1,8 @@
 Title: Clouds
 TODO:  Needs to explain available auth types for clouds
-       Review required, check command outputs
+       Critical: Review required
+       Bug tracking: https://bugs.launchpad.net/juju/+bug/1749302
+       Bug tracking: https://bugs.launchpad.net/juju/+bug/1749583
   
 # Clouds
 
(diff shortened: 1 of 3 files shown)


src/en/clouds.md
```diff
@@ -161,29 +164,36 @@ expand the relevant section). You can also generate a YAML file
...
+      Select cloud type: manual
+      
+      Enter a name for your manual cloud: mycloud
+      
+      Enter the controller's hostname or IP address: noah@10.143.211.93
+      
+      Cloud "mycloud" successfully added
+      You may bootstrap with 'juju bootstrap mycloud'
+
+   You must now add a credential for this cloud prior to creating a controller
```
kwmonroe: You mentioned adding language like this for other clouds, but I don't think it needs to be said here for a manual cloud.

pmatulis: Gah, I knew that and forgot! :+1:

PR #2447, opened 2018-02-15
generic credentials management and update azure credentials example

### LINKED ISSUE #2441: update-credential documentation is unclear
As was seen by the experience in https://bugs.launchpad.net/bugs/1735402, we need to make sure that our online documentation provide clear guidance as to what needs to happen when cloud credentials expire, say  on a long living controller.

Clarity around update-credential and its limitations is a start...

pmatulis:
resolves #2441 

This was tricky.

### DIFF
--- a/src/en/credentials.md
+++ b/src/en/credentials.md
@@ -251,11 +251,12 @@ The update is a two-step process. First change the credentials locally with the
 `add-credential` command (in conjunction with the `--replace` option) and then
 upload those credentials to the controller.
 
-Below, we log in with the correct Juju username (e.g. 'admin'), change the
-contents of the credential called 'joe', and then update them on a Google cloud
-controller:
+Below, we explicitly log in with the correct Juju username ('admin'), change
+the contents of the credential called 'joe', and then update them on a Google
+cloud controller:
 
 ```bash
+juju login -u admin
 juju add-credential --replace joe
 juju update-credential google joe
 ```
--- a/src/en/credentials.md
+++ b/src/en/credentials.md
@@ -211,13 +211,20 @@ You can set the default credential for a cloud:
 juju set-default-credential aws carol
 ```
 
-If only one credential name exists, it will become the effective default
-credential.
+Notes:
 
-### Replacing local credentials
+ - This affects operations that require a newly-input credential (e.g.
+   `juju add-model`). In particular, it does not change what is currently in
+   use (on a controller).
+ - If only one credential name exists, it will become the effective default
+   credential.
 
-To replace an existing credential locally, edit or create a file, such as
-our 'mycreds.yaml' example above, and run:
+### Updating local credentials
+
+To update an existing credential locally use the `add-credential` command with
+the `--replace` option.
+
+Here we decided to use the file 'mycreds.yaml' from a previous example:
 
 ```bash
 juju add-credential aws -f mycreds.yaml --replace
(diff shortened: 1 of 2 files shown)


src/en/credentials.md
```diff
@@ -198,14 +203,21 @@ juju credentials --format yaml --show-secrets
 
 The YAML output will be similar to our 'mycreds.yaml' sample above.
 
-You can set the default credential:
+### Setting default credentials
+
+You can set the default credential for a cloud:
```
anastasiamac: This only works for locally stored credentials. I think we should be here about this distinction here too.

src/en/credentials.md
```diff
@@ -198,14 +203,21 @@ juju credentials --format yaml --show-secrets
...
 ```
 
-To replace or update an existing credential, edit or create a file, such as
-our 'mycreds.yaml' example above, then run:
+If only one credential name exists, it will become the effective default
+credential.
+
+### Replacing local credentials
+
+To replace an existing credential locally, edit or create a file, such as
```
anastasiamac: To replace or update?...

src/en/credentials.md
```diff
@@ -214,12 +226,90 @@ juju add-credential aws -f mycreds.yaml --replace
...
+juju add-credential --replace joe
+juju update-credential google joe
+```
+
+!!! Warning:
+    It is not possible to update the credentials if the initial credential name
+    is unknown. This restriction will be removed in an upcoming release of
+    Juju.
+
+####  Updating credentials with a different Juju user
```
anastasiamac: Thats' not right. This effectively is talking about "replacing' a model credential on the controller, i.e. changing 'cloud-name_userA_credential-name' with 'cloud-name_userB_credential-name'.

This is not possible. You must be the same the user.

There have been cases in the field where credential owner has left the company. These scenarios, until my rework lands, cannot be dealt with without db surgery.

src/en/credentials.md
```diff
@@ -214,12 +226,90 @@ juju add-credential aws -f mycreds.yaml --replace
...
+    old-credential-name:
+      auth-type: service-principal-secret
+      application-id: foo1
+      application-password: foo2
+      subscription-id: bar
+```
+
+### Removing credentials
+
 If a credential is no longer required, it can be removed:
```
anastasiamac: This should clarify - 'locally'. 'remove-credential' command only operates on a client- not on the controller-stored credentials.

src/en/credentials.md
pmatulis: :ok:

src/en/credentials.md
pmatulis: Let's use "update".

src/en/credentials.md
pmatulis: I understand, but the instructions are meant to express what a normal person wants to do. These steps *do* "update the credentials on the controller" by way of "using a different Juju user".

src/en/credentials.md
pmatulis: :ok:

src/en/credentials.md
```diff
@@ -214,12 +226,90 @@ juju add-credential aws -f mycreds.yaml --replace
...
+to use a different one. This implies adding a new credential name, copying over
+any authentication material into the old credential name, and finally updating
+the credentials. Below we demonstrate this for the Azure cloud:
+
+Add a new temporary credential name (like 'new-credential-name') and gather all
+credential sets (new and old):
+
+```bash
+juju add-credential azure
+juju credentials azure --format yaml --show-secrets > azure-creds.yaml
```
evilnick: I'm not sure I really understand the use case or the procedure here

pmatulis: Evidently it's not clear, and I'll work on that, but these two commands generate a new credential set and dumps it (and the old credential set) into a file. I summarized this as "Add a new temporary credential name (like 'new-credential-name') and gather all credential sets (new and old)".

src/en/credentials.md
```diff
@@ -214,12 +226,90 @@ juju add-credential aws -f mycreds.yaml --replace
...
+
+ - same cloud name
+ - same Juju username (logged in)
+ - same credential name
+
+The update is a two-step process. First change the credentials locally with the
+`add-credential` command (in conjunction with the `--replace` option) and then
+upload those credentials to the controller.
+
+Below, we log in with the correct Juju username (e.g. 'admin'), change the
```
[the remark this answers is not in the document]
pmatulis: :ok:

PR #2464, opened 2018-02-23
Fix list

galgalesh:
List was not displayed correctly because of the lack of space between paragraph and list.


galgalesh:
see: https://jujucharms.com/docs/2.3/models-config#firewall-mode

evilnick:
Cool, thanks!

PR #2525, opened 2018-03-16
Review and refactor charms-deploying.md

### LINKED ISSUE #2520: Review charms-deploying.md
As per TODO notes.

pmatulis:
resolves #2520 

:information_source: I didn't move the commented section (in charms-deploying.md) to a troubleshooting page immediately because of a recently merged PR. I didn't merge the latter into this branch as I understand this messes up the backporting tool.

### DIFF
--- a/src/en/charms-deploying-advanced.md
+++ b/src/en/charms-deploying-advanced.md
@@ -1,0 +1,259 @@
+Title: Deploying applications - advanced
+
+# Deploying applications - advanced
+
+This page is dedicated to more advanced topics related to deploying
+applications with Juju. The main page is
+[Deploying applications][charms-deploying].
+
+Topics covered here are:
+
+ - Multi-series charms
+ - Deploying to specific machines
+ - Deploying to spaces
+
+## Multi-series charms
+
+Charms can be created that support more than one release of a given operating
+system distro, such as the multiple Ubuntu releases shown below. It is not
+possible to create a charm to support multiple distros, such as one charm for
+both Ubuntu and CentOS. Supported series are added to the charm metadata like
+this:
+
+```
+name: mycharm
+summary: "Great software"
+description: It works
+maintainer: Some One <some.one@example.com>
+categories:
+   - databases
+series:
+   - trusty
+   - xenial
+provides:
+   db:
+     interface: pgsql
+requires:
+   syslog:
+     interface: syslog
+```
+
+The default series for the charm is the first one listed. So, in this example,
+to deploy `mycharm` on `trusty`, all you need is:
+
+```bash
+juju deploy mycharm
+```
+
+You can specify a different series using the `--series` flag:
+
+```bash
+juju deploy mycharm --series xenial
+```
+
+You can force the charm to deploy using an unsupported series using the
+`--force` flag:
+
+```bash
+juju deploy mycharm --series yakkety --force
+```
+
+Here is a more complete example showing a new machine being added that uses
+a different series than is supported by our `mycharm` example and then forcing
+the charm to install:
+
+```bash
+juju add-machine --series yakkety
+Machine 1 added.
+juju deploy mycharm --to 1 --series yakkety --force
+```
+
+It may be required to use `--force-series` when upgrading charms. For example,
+in a case where an application is initially deployed using a charm that
+supports `precise` and `trusty`. If a new version of the charm is released that
+only supports `trusty` and `xenial` then it will be allowed to upgrade
+applications deployed on `precise`, but only using `--force-series`, like this:
+
+```bash
+juju upgrade-charm mycharm --force-series
+```
+
+## Deploying to specific machines
+
+It is possible to specify which machine or container an application is to be
+deployed to. One notable reason is to reduce costs when using a public cloud;
+applications can be consolidated instead of dedicating a machine per
+application unit.
+
+Below, the `--constraints` option is used to create an LXD controller with
+enough memory for other applications to run. The `--to` option is used to
+specify a machine:
+
+```bash
+juju bootstrap --constraints="mem=4G" lxd lxd-controller
+juju deploy mysql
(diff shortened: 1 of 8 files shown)


src/en/charms-deploying-advanced.md
```diff
@@ -0,0 +1,289 @@
...
+
+The first example deploys to a specific AWS zone while the second example
+deploys to a named machine in MAAS.
+
+### add-unit --to
+
+The `add-unit` command also supports the `--to` option, including placement
+directives.
+
+<!-- THE EXAMPLE DOES NOT WORK AND THE PURPOSE OF THE TEXT IS NOT CLEAR.
```
evilnick: I don't see the point of purposefully adding something we aren't happy with and then exposing it in a comment.

The purpose of add-unit  --to is for situaions e.g. to mitigate for failure, where multiple nodes are running many services. So in what way does it not work?

src/en/charms-deploying.md
```diff
@@ -1,42 +1,37 @@
...
-through the use of charms (the magic bits of code that make things just work).
-These charms can exist in the [Charm Store](https://jujucharms.com/store) or on
-the file system (previously downloaded from the store or written locally).
+The fundamental purpose of Juju is to deploy and manage software applications
+in a way that is fast and easy. All this is done with the help of *charms*,
+which are bits of code that contain all the necessary intelligence to do these
+things. Charms can exist online (in the [Charm Store][charm-store]) or on your
+local filesystem (previously downloaded from the store or written locally).
 
 Charms use the concept of *series* analogous as to how Juju does with Ubuntu
```
evilnick: In this new context, I think this paragraph needs rewriting too.

src/en/charms-deploying.md
```diff
@@ -90,69 +82,10 @@ See [Upgrading applications][charms-upgrading] for how charm upgrades work.
...
-supports `precise` and `trusty`. If a new version of the charm is released that
-only supports `trusty` and `xenial` then it will be allowed to upgrade
-applications deployed on `precise`, but only using `--force-series`, like this:
-
-```bash
-juju upgrade-charm mycharm --force-series
-```
+Charms can be written to support more than one release of a given operating
+system distro, such as the multiple Ubuntu releases shown below. See the
+documentation on [Multi-series charms][deploying-multi-series-charms] to learn
```
evilnick: On a user level, I don't think anybody cares, apart from the upgrading case

src/en/charms-deploying-advanced.md
pmatulis:
```
juju deploy --constraints="mem=4G" openstack-dashboard
Located charm "cs:openstack-dashboard-258".
Deploying charm "cs:openstack-dashboard-258".

juju add-unit --to 1 rabbitmq-server
ERROR application "rabbitmq-server" not found (not found)
```

src/en/charms-deploying.md
pmatulis: :ok:

src/en/charms-deploying.md
pmatulis: Kindly elaborate on what you would like changed.

src/en/charms-deploying-advanced.md
evilnick: well, I imagine that's because openstack-dashboard doesn't include rabbit-mq. It doesn't mean that add-unit --to is broken, just that this example doesn't work because the openstack charms have changed

src/en/charms-deploying.md
evilnick: I think it is more about 'Deploying a charm with a specific series'

Some charms support more than one series. It is also possible to force a charm to deploy to a different series blah blah etc.

src/en/charms-deploying-advanced.md
pmatulis: I didn't say that `add-unit` was broken. I said the example was broken. And because it was, I couldn't understand the surrounding text. Now I think I get it. The purpose is to show a charm (a bundle?) that installs multiple applications (e.g. openstack-dashboard and rabbitmq-server) distributed across multiple machines, and to place an additional application unit (e.g. rabbitmq-server) on one of those machines where it currently does not exist. It's fairly convoluted because the user needs to know that the original application installs multiple sub-applications and which machine does not already contain the application-to-be-added. I personally don't think it's a great example. It has more the potential to confuse than the potential to clarify. If you have another legitimate example I can include it otherwise I think we can just omit one altogether.

evilnick: no, it's a rubbish example. But we don't really need to invent some scenario with real charms where it makes sense, we just tell people why they may want to do it, then show them how. Use nginx and postgres or something

src/en/charms-deploying.md
pmatulis: :ok:

pmatulis:
@evilnick Please review

src/en/charms-upgrading.md
```diff
@@ -56,6 +56,15 @@ so long as the following conditions are met:
 
 The new charm may add new relations and configuration settings.
 
+## Forced upgrades
+
+A charm upgrade may require the use of the `--force-series` option.
+
+Consider the case where an application is initially deployed using a charm that
+supports Trusty and Precise. If a new version of the charm is released that
+only supports Trusty and Xenial then applications deployed on Precise will only
```
evilnick: Several things trouble me about this paragraph:
 Surely, "Precise and Trusty". Why wouldn't you put them in chronological order?
i don't understand 'applications on precise will only be allowed by...'
What series, in this scenario, was the application deployed on?
What series are you trying to move it to? 
It isn't very clear or easy to follow

pmatulis: Yes, I changed the order of Trusty and Precise due to a brain fart.
Re your other comments, I neglected a couple of words. It should read better now.

PR #2530, opened 2018-03-20
LP #1753593: AWS st1 and sc1 volumes

AdrianGPrado:
## Please provide the following details to expedite Pull Request review:

Launch Pad[1753593](https://bugs.launchpad.net/juju/+bug/1753593)

----

## Description of change

AWS has changed its volume instances types.
- New `st1` magnetic volume type, with: big data, data warehouses, log processing... as use cases.
- New `sc1` magnetic volume type, with: cold data storage, few scans per day... as use cases.
- `magentic` type marked as depreciated, volume type ssd will be the new default option.

### DIFF
--- a/src/en/charms-storage.md
+++ b/src/en/charms-storage.md
@@ -130,11 +130,11 @@ Here is sample output for a newly-added AWS model:
 
 ```no-highlight
 Name     Provider  Attrs
-ebs      ebs       
+ebs      ebs
 ebs-ssd  ebs       volume-type=ssd
-loop     loop      
-rootfs   rootfs    
-tmpfs    tmpfs     
+loop     loop
+rootfs   rootfs
+tmpfs    tmpfs
 ```
 
 !!! Note:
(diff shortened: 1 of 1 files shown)


pmatulis:
@AdrianGPrado Unless I'm missing something, Juju now supports (after 2 years of being offered by AWS) the `st1` and `sc1` magnetic volume types. In addition, Juju is changing its default volume type from `standard` (magnetic) to `gp2` (ssd). I don't think we should use the term "deprecated" since the `standard` type is still available. I also don't see the need to include a link to the 2-yr old "update" from AWS. Unless you want to add any more commentary, I am willing to merge this now and I will make some changes in a later PR.

pmatulis:
@nskaggs Can you confirm that the ssd-as-default change only affects Juju 2.3? It does seem that 2.2 docs should be updated regarding the availability of `st1` and `sc1` volume types though. Or would those only be options for 2.3 as well?

PR #2543, opened 2018-03-21
Add documentation-builder to 1.25 docs

anthonydillon:
## Done
- Added a build script
- Created the navigation yaml
- Moved the media to the src directory
- Updated links to the images in the docs

## QA
- Run `./build-docs.sh`
- Then `cd build`
- Run a server, for example: `python -m SimpleHTTPServer 8000`
- Go to http://localhost:8000/en/config-aws.html
- Look around and check it all works


evilnick:
suggestion - don't move all the media files! just create a symlink

      cd src
      ln -s ../htmldocs/media media

this removes the need to break current jujucharms build, cuts the diff considerably

src/en/authors-charm-components.md
```diff
@@ -9,7 +9,7 @@ metadata, configuration data, and hooks with some extra support files.
 
 A charm requires only a single file in order to be considered valid by juju:
 
- - `metadata.yaml` [describes the charm](./authors-charm-metadata.html) and the
+ - `metadata.yaml` [describes the charm](../authors-charm-metadata.html) and the
```
evilnick: I don't think this is the path we want, it should be just
      [describes the charm](authors-charm-metadata.md)
or
     [describes the charm](./authors-charm-metadata.md)

src/en/authors-charm-metadata.md
```diff
@@ -22,7 +22,7 @@ Here's a valid metadata file:
     description: |
       MongoDB is a high-performance, open source, schema-free document- oriented
       data store that's easy to deploy, manage and use. It's network accessible,
-      written in C++ and offers the following features:  
+      written in C++ and offers the following features:
```
evilnick: I think covered in a different review, we don't want to strip all the trailing spaces - some of them have meaning

tools/mdbuild.py
```diff
@@ -143,7 +143,7 @@ def post_process(self):
             ('%%CONTENT%%', self.parsed),
             ('%%DOCNAV%%', doc_nav),
             ('src="media/', 'src="../media/'),
-            ('src="./media/', 'src="../media/'),
+            ('src="../media/', 'src="../media/'),
```
evilnick: don't change this

src/en/authors-charm-components.md
anthonydillon: Good point, I'll revert the html links and leave the image links.

src/en/authors-charm-metadata.md
anthonydillon: What are they used for? I have checked both versions and they look the same.

tools/mdbuild.py
anthonydillon: :+1:

anthonydillon:
I copied the images for now as we want to remove the htmldocs directory once we go live with the new system. I didn't want use to forgot but it is more maintenance so will symlink for now. Just need to remember when removing the htmldocs directory to move the media directory.

src/en/authors-charm-metadata.md
evilnick: two trailing spaces insert a `<br>`. It is often a neater way to do it that to include html in the source or to mess the format up with hard carraige returns (which can cause other problems in markdown). They are pretty much essential in lists, as its the only way to force a linebreak without ending the list. So, although many of them are just cruft, some are there on purpose. It's just easier to disable any trailing space deletion, or people get [upset about it](https://github.com/Microsoft/vscode/issues/1679)

evilnick:
don't worry, we won't forget about them :smiley:

anthonydillon:
I'll re-do the content so it retains whitepspacing. Will try and get to that tonight.

anthonydillon:
@evilnick Ok think I have reverted all the whitespace changes.

tools/mdbuild.py
```diff
@@ -143,7 +143,7 @@ def post_process(self):
             ('%%CONTENT%%', self.parsed),
             ('%%DOCNAV%%', doc_nav),
             ('src="media/', 'src="../media/'),
-            ('src="./media/', 'src="../media/'),
+            ('src="./media/', 'src="./media/'),
```
evilnick: :laughing: now you've changed the other side of this! Leave poor mdbuild.py alone!

evilnick:
This looks promising! i will give it a proper run through in the morning

anthonydillon:
Woops, my bad. reverted `mdbuild.py`. Cheers @evilnick

PR #2562, opened 2018-03-28
update release notes for 2.3.5

pmatulis:
I reformatted the auxiliary sections by including hyperlinks. Some slight rewording was also made in those sections and these notes will therefore differ a little from those posted on the [mailing lists](https://lists.ubuntu.com/archives/juju-dev/2018-March/006682.html). Note that the actual bug titles do not correspond to the (mailing list) posted notes. I corrected them in these online notes.


degville:
:+1: looks good. I'd end each item with a full stop, make Openstack OpenStack and backtick the snap command, but it may not be worth it.

pmatulis:
@degville The bug title spells it like that ("Openstack") and I felt this should be preserved.

PR #2563, opened 2018-03-28
update commands file and implement some temporary fixes

### LINKED ISSUE #2553: Regenerate commands.md
There is at least one command (`juju ssh` as per https://github.com/juju/juju/pull/8261) that needs updating in the commands.md file, probably more.

pmatulis:
resolves #2553 

## Problems

My recent change to the opening paragraph failed as a formatted (backticked) command ended up getting linewrapped (and a carriage return was thus injected; see existing [bug](https://github.com/canonical-webteam/documentation-builder/issues/101)). I will send in another PR to can-doc-utils that mirrors the current temporary fix.

Temporarily fixed: `import-filesystem` had its usage line linewrapped, causing breakage

### Examples formatting

There is an everpresent problem with example section formatting:

Temporarily fixed: ssh, scp

But I decided that a more indepth, and permanent, solution should wait for another time. A few instances of malformatted examples sections:

add-relation
add-storage
add-unit

There are even command prompts.

### DIFF
--- a/src/en/commands.md
+++ b/src/en/commands.md
@@ -1,10 +1,14 @@
-Title: Juju commands and usage
+Title:Juju commands and usage
 
 # Juju Command reference
 
-You can get a list of Juju commands by entering `juju help commands` on the
-command line. These commands are also listed below, along with usage and
-examples. Click on the expander triangle to see the details of a command. 
+You can get a list of all Juju commands by invoking `juju help commands` in a
+terminal.
+
+To drill down into each command use `juju help <command name>`.
+
+This same information is also provided below. Click on the
+triangle alongside a command to view that command's entry.
 
 ^# actions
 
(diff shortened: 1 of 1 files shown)


src/en/commands.md
```diff
@@ -1226,6 +1239,8 @@ examples. Click on the expander triangle to see the details of a command.
 
           # Sets the budget for the current model to 10.
           juju budget 10
+          # Moves the budget for the current model to wallet 'personal' and sets the limit to 10.
```
evilnick: we should fix this upstream, this is a shitty summary

src/en/commands.md
```diff
@@ -8433,9 +8486,12 @@ examples. Click on the expander triangle to see the details of a command.
 
    **Examples:**
 
-          juju sla essential              # set the support level to essential
-          juju sla standard --budget 1000 # set the support level to essential with a maximum budget of $1000
-          juju sla                        # display the current support level for the model.
+          # set the support level to essential
+          juju sla essential
```
evilnick: more poorly formatted examples

evilnick:
The problem with the examples is simply down to the way there are about 4 different formats in use, in the juju help, and the conversion script has to guess what is going on. It is easier/better for everyone to fix them upstream

PR #2566, opened 2018-03-29
add admin vs owner notes

### LINKED ISSUE #2531: Controller admins can upgrade any model
As per https://github.com/juju/juju/pull/7424

Previously, apparently, controller admins could not upgrade models they did not create, only model creators (model owners?) could. Now, controller admins also have this right.

pmatulis:
resolves #2531

### DIFF
--- a/src/en/models-upgrade.md
+++ b/src/en/models-upgrade.md
@@ -44,6 +44,8 @@ run on each machine Juju creates, including controllers.
 
 Overview:
 
+- A controller admin user can upgrade any model within their controller and a
+  model owner can only upgrade their own model.
 - Upgrades must be applied to the controller model first.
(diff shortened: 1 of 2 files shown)


src/en/models-upgrade.md
```diff
@@ -44,6 +44,8 @@ run on each machine Juju creates, including controllers.
 
 Overview:
 
+- A controller admin user can upgrade any model within their controller and a
```
evilnick: I would make these two separate things

 - The admin user of a controller can upgrade any model within that controller
 - The creator of a model can upgrade that model

pmatulis: Ok, but I preserved the user terminology used elsewhere. Also, it really is "model owner". A user can be made an owner without having created the model.

PR #2571, opened 2018-03-29
Update 'juju switch' to cover various usage shortcuts

### LINKED ISSUE #2537: Extend and enhance `switch` command
Since 2.0, it was possible to:

`juju switch foo` -> switches to the last used model in controller foo if it exists, otherwise model foo in the current controller
`juju switch :foo` -> switches to model foo in the current controller
`juju switch foo:bar` -> switches to model bar in controller foo

This is not well explained and so the current docs can be enhanced.

Secondarily, a recent PR introduced:

`juju switch foo:` -> switches to the last used model in controller foo

This is the extension to the docs (as per https://github.com/juju/juju/pull/7556).

pmatulis:
resolves #2537

### DIFF
--- a/src/en/models.md
+++ b/src/en/models.md
@@ -101,8 +101,6 @@ Common model management tasks are summarized below.
    
    To change to a model:
    
-   <!-- WARNING: EXTRA SPACES AT EOL -->
-   
    `juju switch foo`  
    Selects the last used model in controller 'foo' (if the latter exists),
    otherwise model 'foo' in the current controller.
--- a/src/en/models.md
+++ b/src/en/models.md
@@ -92,13 +92,30 @@ Common model management tasks are summarized below.
    
    Use the `juju switch` command to go from one model to another:
    
-   `juju switch [<controller>|<model>|<controller>:<model>]`
+   `juju switch [<controller or model>|<:model>|<controller>:<model>|<controller:>]`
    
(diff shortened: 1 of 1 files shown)


src/en/models.md
```diff
@@ -92,13 +92,30 @@ Common model management tasks are summarized below.
...
+   `juju switch [<controller or model>|<:model>|<controller>:<model>|<controller:>]`
    
    Running the command with no arguments will return the currently active 
    controller and model:
      
          juju switch
    
+   To change to a model:
+   
+   <!-- WARNING: EXTRA SPACES AT EOL -->
```
evilnick: we don't need this, we just need to shoot anyone who strips spaces

src/en/models.md
```diff
@@ -92,13 +92,30 @@ Common model management tasks are summarized below.
...
+   
+   `juju switch foo`  
+   Selects the last used model in controller 'foo' (if the latter exists),
```
evilnick: does this mean if you have a controller foo, and a model bar:foo then:

juju switch bar
juju switch foo

will take you to back to the foo container?

pmatulis: If controller foo exists, the command `juju switch foo` will, irrespective of previous commands ("where you are"), always take you to the last used model in controller foo.

evilnick: how confusing

PR #2589, opened 2018-03-30
add bootstrap example with cloud-specific feature

### LINKED ISSUE #2532: Explain `show-cloud --include-config`
This option is useful enough to mention. It allows the user to see what cloud-specific configuration options are available. So far it looks like only the 'aws' and 'openstack' clouds are affected. It is meant to be run prior to the creation of a controller (`juju show-cloud --include-config aws`). This brings up the current problem of not knowing where/how an option is used: as a controller config option, as a model (controller or non-controller) option, or as a hidden option (e.g. `bootstrap-timeout`). 

As per https://github.com/juju/juju/pull/7449

pmatulis:
resolves #2532

### DIFF
--- a/src/en/clouds.md
+++ b/src/en/clouds.md
@@ -88,6 +88,12 @@ If you want more detail about a particular cloud, use:
 juju show-cloud azure
 ```
 
+To learn of any special features a cloud may support the `--include-config`
+option can be used with `show-cloud`. These can then be passed to either of the
+`bootstrap` or the `add-model` commands. See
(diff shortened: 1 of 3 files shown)


src/en/clouds.md
```diff
@@ -88,6 +88,12 @@ If you want more detail about a particular cloud, use:
 juju show-cloud azure
 ```
 
+To learn of any special features a cloud may support the `--include-config`
```
evilnick: I don't really like this here, as it seems to be making something more complicated than it should be, but if it doesn't go here, I am not sure how anyone would know it exists.

PR #2613, opened 2018-04-17
Add ProxyFix as a middleware to set the WSGI scheme ffor X-Forwarded-Proto

anthonydillon:
The issue is that images are redirected within the app via Flask. Flask is not aware its served over HTTPS so redirects are unsecure. Resulting in an orange lock in the browser. 

I am not sure of a good way to test this locally.

Please review @nottrobin @WillMoggridge

### DIFF
--- a/app.py
+++ b/app.py
@@ -4,6 +4,7 @@ import re
 
 # External packages
 import flask
+from werkzeug.contrib.fixers import ProxyFix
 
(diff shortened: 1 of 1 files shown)


evilnick:
well, it looks sensible to me, but yeah, I'm not sure how we can adequately test

PR #2636, opened 2018-04-18
Add legacy versions of the docs

anthonydillon:
## Done
Added the legacy versions of the docs to the build folder and stopped it from being removed when running `./run clean`.

## QA
- Pull this branch run
- Run `./run`
- Open http://0.0.0.0:8205/2.3/en/about-juju
- Check it looks good
- Check the versions in the menu work
- Change the URL to http://0.0.0.0:8205/1.18/en/about-juju
- See that it works and looks like the existing docs
- Run `./run clean`
- Run `./run` and check it all works still


evilnick:
I can't get your version to build:
https://paste.ubuntu.com/p/Y8GG7j2Gjn/

anthonydillon:
@nottrobin any idea?

anthonydillon:
@evilnick Oh this looks like you need to update the versions in your fork.

evilnick:
but it isn't my fork is it, it's yours. Anyhow, I have checked out all the remote branches that we publish - 1.25, 2.2, 2.3

And besides that, from the pastebin it looks exactly like documentation builder failing to parse something

anthonydillon:
I have updated all my branches, I thought it would pull from the origin but you could be right.

evilnick:
Okay, I got this to run by checking out all the required bracnhes, adding juju/docs.git as an additional remote and manually pulling all the changes from there.

Unfortunately, I think this will be an issue every time we want to review a branch under the new system This does cause serious problems - not to do with this PR, but in general.

Anyhow, the styling on the old docs is a little different.

olddocs

Also the images don't load - I think they expect a 1.25/media folder, for example. I think another symlink would be fine for now

nottrobin:
So. The problem is that for `./run` to work, you need to have all version branches properly created in your local git repository. So to fix it, you need to create all these version branches from the main repository remote.

On my system, the main remote is called "origin". If it's different for you (e.g. "upstream"), replace "origin" in the command below.

``` bash
for remote in `git branch -r | grep -v /HEAD | grep origin/`; do git checkout --track $remote ; done
```

Now check out this PR again, and `./run` should work for you.

anthonydillon:
@evilnick I have fixed the images. The issue with the fonts is a known one as Google fonts doesn't have the correct font for 300. I don't think this should be a blocker and something that I can work on later.

PR #2644, opened 2018-04-20
Add CAAS

pmatulis:


src/en/caas.md
```diff
@@ -0,0 +1,145 @@
+Title: CAAS and Juju
+TODO:  Once 2.4 is officially released remove support Note and update status output
+
+# CAAS and Juju
+
+CAAS is *Containers as a Service*, a cloud service that allows users to upload,
```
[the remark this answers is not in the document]
pmatulis: The problem is that the acronym CAAS is already taken. It really is a type of cloud service that other vendors are providing, such as Amazon's Elastic Container Service (ECS) and Google's Kubernetes Engine. So I find it difficult to use the term "CAAS" as something unique to Juju. This is definitely worthy of discussion.

src/en/caas.md
```diff
@@ -0,0 +1,145 @@
...
+
+First off, a Kubernetes cluster will be required. Essentially, you will use it
+as you would any other cloud that Juju interacts with: the cluster becomes the
+backing cloud.
+
+To summarise, the steps for using Kubernetes with Juju are:
+
+ 1. Obtain a cluster
+ 1. Add the cluster to Juju
+ 1. Deploy CAAS-specific charms
```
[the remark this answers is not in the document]
pmatulis: :ok:

src/en/caas.md
```diff
@@ -0,0 +1,145 @@
...
+kubernetes-master:cni                flannel:cni                          kubernetes-cni    subordinate  
+kubernetes-master:kube-api-endpoint  kubernetes-worker:kube-api-endpoint  http              regular      
+kubernetes-master:kube-control       kubernetes-worker:kube-control       kube-control      regular      
+kubernetes-worker:cni                flannel:cni 			  kubernetes-cni    subordinate
+```
+
+### Add the cluster to Juju
+
+We will need some information about the cluster in order to add it to Juju.
+This is found within the main Kubernetes configuration file.
```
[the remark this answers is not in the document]
pmatulis: I explain all this just a few lines down.

src/en/caas.md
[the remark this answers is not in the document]
evilnick: I'm not sure whether what @wallyworld is suggesting is that we don't call it 'CAAS'?

That would probably be better than labelling it 'CAAS' but meaning something different

src/en/caas.md
```diff
@@ -0,0 +1,145 @@
+Title: CAAS and Juju
+TODO:  Once 2.4 is officially released remove support Note and update status output
+
+# CAAS and Juju
+
+CAAS is *Containers as a Service*, a cloud service that allows users to upload,
+organize, run, scale, manage, and stop containers. Combining this with Juju
+opens up a new vista of possibilities for Juju users. Currently, Juju supports
```
evilnick: I'm not sure about a 'vista of possibilities', let's just stick with practical benefits

pmatulis:
@wallyworld I've put the configuration keys into a table. I'm unsure about the defaults. I kept them as-is from the gdoc but can they be made better? For instance, is there a difference between `[]` and `""`?

src/en/caas.md
```diff
@@ -0,0 +1,192 @@
+Title: CAAS and Juju
```
wallyworld: Generally, CAAS should be "Kubernetes" as per Mark's latest thinking on terminology.

src/en/caas.md
```diff
@@ -0,0 +1,192 @@
...
+TODO:  Once 2.4 is officially released remove support Note and update status output
+       Should eventually link to CAAS-charm developer documentation
+       Update when storage becomes a Juju drivable aspect.
+       Add architectural overview/diagram once CAAS becomes stable.
+       Consider manually adding a cluster (third-party installs) via `add-cloud` and `add-credential`
+
+# CAAS and Juju
+
+CAAS is *Containers as a Service*, a cloud service that allows users to upload,
```
wallyworld: Given the Kubernetes focus, we probably just want to say something like that Juju supports managing workloads on Kubernetes in a similar way to how it manages workloads on VM based clouds etc etc

src/en/caas.md
```diff
@@ -0,0 +1,192 @@
...
+managing containerised applications at scale (see the
+[Kubernetes documentation][upstream-kubernetes-docs] for more information). It
+most commonly employs Docker as its container technology.
+
+## CAAS-specific workflow
+
+Here we discuss how a building and working with a CAAS environment may differ
+from a standard Juju workflow.
+
+The only CAAS-specific Juju command is `add-k8s`, which acts as the `add-cloud`
```
wallyworld: add-k8s does 2 things:
- add-cloud
- add-credential
It creates a cloud definition for the k8s cluster and imports a credential and CA cert from the k8s config so Juju can talk to the cluster. The add-k8s command is currently undergoing a little refinement as per https://bugs.launchpad.net/bugs/1768837

src/en/caas.md
```diff
@@ -0,0 +1,192 @@
...
+
+First off, a Kubernetes cluster will be required. Essentially, you will use it
+as you would any other cloud that Juju interacts with: the cluster becomes the
+backing cloud.
+
+To summarise, the steps for using Kubernetes with Juju are:
+
+ 1. Obtain a cluster
+ 1. Add the cluster to Juju
+ 1. Create a controller (and optionally a model)
```
wallyworld: So this is where some of the add-k8s refinement is happening. Currently you need a controller before add-k8s is run since the command tells the controller about the cloud as well as storing the info locally. Cory's latest comment in the bug has prompted a slight rethink there.

src/en/caas.md
```diff
@@ -0,0 +1,192 @@
...
+```bash
+juju add-k8s k8cloud
+```
+
+## Configuration
+
+Juju CAAS applications support application specific configuration. This allows
+k8s configuration to be used to control how Juju deploys the application on
+Kubernetes. The following are supported (these names are the Juju configuration
+attribute names; the k8s meaning should be obvious):
```
wallyworld: the k8s meaning is described in the relevant k8s service and ingress resource documentation

src/en/caas.md
```diff
@@ -0,0 +1,192 @@
...
+kubernetes-service-target-port			| string  | <container port> |		    |
+kubernetes-service-loadbalancer-ip		| string  | ""		     |		    |
+kubernetes-service-loadbalancer-sourceranges	| string  | []		     |		    |
+kubernetes-service-externalname			| string  | ""		     |		    |
+kubernetes-ingress-class			| string  | nginx	     |		    |
+kubernetes-ingress-ssl-redirect			| boolean | false	     |		    |
+kubernetes-ingress-ssl-passthrough		| boolean | false	     |		    |
+kubernetes-ingress-allow-http			| boolean | false	     |		    |
+
+There are three other configuration attributes which are not k8s-specific:
```
wallyworld: two

pmatulis:
As per https://github.com/juju/juju/pull/8905, CAAS storage is always removed with the unit.

pmatulis:
The main lingering question is the lack of production charms.

pmatulis:
[Kubernetes](https://trello.com/c/kBqAb58w/646-kubernetes)

pmatulis:
@wallyworld I'm done here.

mitechie:
- The lack of production charms

Talked to the JAAS folks today and all is deployed. There's a patch to the charm command that was up today that fixes an issue submitting k8s charms to the store. Once that gets built we can test out adding/deploying k8s charms from the public production store.

PR #2670, opened 2018-05-01
Review constraints

### LINKED ISSUE #2460: Review reference-constraints.md
As per TODO notes.

### LINKED ISSUE #2669: Review charms-constraints.md
Current constraints documentation is terribly belaboured, confusing, and repetitive. A re-write and refactor  is in order.

involves #2460

pmatulis:
resolves #2669 
resolves #2460


src/en/charms-constraints.md
```diff
@@ -1,187 +1,233 @@
...
+[Reference: Juju constraints][reference-constraints] page.
 
-For more granularity, it is also possible to add a machine with specific 
-constraints (`juju add-machine`) and then specify that machine when deploying 
-applications ([see the documentation on `juju deploy`](./charms-deploying.html)).
+The idealized use case is that of stipulating a constraint when deploying an
+application and the backing cloud providing a machine with those exact
+resources. In the majority of cases, however, default constraints may have been
+set (at various levels) and the cloud is unable to supply those exact
+resources.
```
evilnick: I don't think it is good to split up the 'ideal' case and the explanation of what happens, which is somewhat below here.

src/en/charms-constraints.md
```diff
@@ -1,187 +1,233 @@
...
-    
-  - **spaces** : Target a particular network space, or avoid one. See 
-  [Network spaces][network-spaces].
-  
-  - **arch** : Short for 'architecture', indicates the processor type an
-  application must run on. One of amd64, arm, i386, arm64, or ppc64el.
-  
-With these you can make sure an application has the resources it needs to run 
-properly.
+## Clouds and constraints
```
evilnick: I don't think this should go in a separate section, this is common to ALL clouds/non-clouds and really, I think these things should be in the 'noteworthy' list.

src/en/charms-constraints.md
```diff
@@ -1,187 +1,233 @@
...
-```
+## Constraint scopes, defaults, and precedence
 
-To further ensure that it also has at least 2 CPU cores:
-  
-```bash
-juju deploy mariadb --constraints "mem=4G cores=2"
-```
+Constraints can be applied to various levels or scopes. Defaults can be set on
+some of them, and in the case of overlapping configurations a precedence is
```
evilnick: this seems vague at this point. Also, the only things you can't apply 'defaults' to are things which make no sense - controller - there is only one. Single-machine, it depends what you mean by that. If you mean a machine generated by 'add-machine', there is a default for that. The only way I can thing you can make a single machine that doesn't inherit constraints if they are set is if you manually add it to a cloud. Am I missing something?

The point to my comment is, this information would be a lot easier to digest if there were less of it, so leaving out a section on 'defaults' and an additional list would be a bonus.

src/en/charms-constraints.md
```diff
@@ -1,187 +1,233 @@
...
-    When setting more than one constraint you will need to utilize quotes.
+### Scopes
+
+On a per-controller basis, the following constraint scopes exist:
+
+ - Controller machine
+ - All models
+ - Single model
+ - Single application
+ - All units of an application
```
evilnick: diagram would probably help here

src/en/charms-constraints.md
```diff
@@ -1,187 +1,233 @@
...
 
-Constraints allow you to choose the hardware (or virtual hardware)
-to which your applications will be deployed, e.g. by specifying the amount of 
-RAM you want them to have. This is particularly useful for making sure that the
-application is deployed somewhere it can actually run efficiently, or that it is
-connected to the right network. Constraints may be set for models and
-applications, with application constraints taking precedence. Changes to
-constraints do not affect any units which have already been placed on machines.
+A *constraint* is a user-defined minimum hardware specification for a machine
+that is spawned by Juju. There are a total of nine constraints, with the most
```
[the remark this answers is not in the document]
pmatulis: :ok:

src/en/charms-constraints.md
pmatulis: :ok:

src/en/charms-constraints.md
pmatulis: But the characteristics are short bullet points and after bringing down the 'idealised case' there are now three paragraphs. How do I combine these with the bullet points?

src/en/charms-constraints.md
pmatulis: AFAIK, you cannot specify a default for the addition of machines. You can set one for a model, though, which affects `add-machine`. This is analogous to setting a default for all models, which affects `add-model`, but there is a definite default that can be set for 'all-models' and a definite default that can be set for 'a model'.

As you know, I'm all for less text but default settings are a central part of constraints. The user should be told about them upfront.

src/en/charms-constraints.md
pmatulis: I'll make a TODO.

src/en/charms-constraints.md
```diff
@@ -1,187 +1,262 @@
...
-is the simplest and most frequent case for using constraints. So, to deploy the
-'mariadb' application to a machine with 4 gigabytes of memory or more:
-  
-```bash
-juju deploy mariadb --constraints mem=4G
-```
+When using the localhost cloud, constraints are ineffectual due the nature of
+this cloud's underlying technology (LXD), where each machine will, by default,
+have access to **all** of the LXD host's resources. Here, an exact hardware
+specification can be requested, but is done at the LXD level. An example will
```
evilnick: ...at the LXD level (see example below).

I think is better

src/en/reference-constraints.md
```diff
@@ -1,157 +1,136 @@
...
-constraints that would conflict with other clouds and cannot be used
-in combination. Use this list to help you understand the differing needs.
+Constraints cannot be applied towards a backing cloud in an agnostic way. That
+is, a particular cloud type may support some constraints but not others. Also,
+even if two clouds support a constraint, sometimes the constraint **value** may
+work with one cloud but not with the other. All this is the natural consequence
+of Juju striving to support widely differing cloud types. The list below
+addresses the situation.
+
+<!-- EXPLANATION REQUIRED
```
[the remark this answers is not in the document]
pmatulis: :ok:

src/en/reference-constraints.md
```diff
@@ -1,157 +1,136 @@
...
+Sometimes, different clouds also dictate constraints that would conflict with
+other clouds and cannot be used in combination.
+-->
 
-###Azure Provider:
+### Azure:
 - Unsupported: [cpu-power, tags, virt-type]
 - Valid values: arch=[amd64]; instance-type=[defined on the cloud]
```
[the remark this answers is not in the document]
pmatulis: I'll add a TODO.

src/en/reference-constraints.md
```diff
@@ -1,157 +1,136 @@
...
+    space, or negative (prefixed with "^"), listing something the space does
+    not have.
 
-- virt-type
+    Example: spaces=storage,db,^logging,^public (meaning, select machines
+    connected to the storage and db spaces, but NOT to logging or public
+    spaces).
 
-    Specifies the type of virtualization to be used, such as `kvm`.
+    **Note:** EC2 and MAAS are the only providers that support the spaces
```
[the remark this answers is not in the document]
pmatulis: :ok:

src/en/charms-constraints.md
pmatulis: :ok:

PR #2699, opened 2018-05-24
Update commands page for 2.3

pmatulis:
Due to 2.3.8 being released an update to the commands reference page was deemed necessary.


pmatulis:
Note that I overwrote most of the manual changes I had made to the last update.

PR #2711, opened 2018-05-25
Fix for shabby jinja2

evilnick:
@pmatulis, you'll just have to take my word for it that this actually does work. I thought it didn't but I was fooled by testing it with ./run. Unless you want to spend half an hour scratching your head, I suggest you just give me some cake and push the button


evilnick:
thanks!

PR #2723, opened 2018-05-28
Review models upgrade (take II)

pmatulis:


src/en/client.md
```diff
@@ -85,8 +82,33 @@ tar -xzf juju-yymmdd-hhmmss.tar.gz
...
+
+```bash
+sudo apt update
+sudo apt install juju
+```
+
+If you have installed Juju via the snap package:
+
+```bash
+sudo snap refresh juju
```
degville: my only comment is that it may be worth mentioning snaps update automatically.

pmatulis: Good idea.

Issue #2724, opened 2018-05-29
default redirects aren't working


evilnick:
https://docs.jujucharms.com/devel/ currently redirects to https://docs.jujucharms.com/devel/en/ and then gives a 500

**What should happen?**
https://docs.jujucharms.com/devel/en/ should redirect to https://docs.jujucharms.com/devel/en/getting-started

**ALSO:** 
https://docs.jujucharms.com/stable/ redirects to https://docs.jujucharms.com/2.3/en/ and then gives a 500

**What should happen**

anything like `https://docs.jujucharms.com/<version>/en/` should point to `.../<version>/en/getting-started`

evilnick:
@pmatulis @anthonydillon @WillMoggridge

anthonydillon:
@evilnick Redirects added here: https://github.com/juju/docs/pull/2727

PR #2727, opened 2018-05-29
Add default redirects from devel and stable

anthonydillon:
Added redirects from:
- `/devel/en/` to `/devel/en/getting-started`
- `/stable/` to `/2.3/en/getting-started`
- `/stable/en/` to `/2.3/en/getting-started`

Fixes https://github.com/juju/docs/issues/2724


pmatulis:
Where does `/devel/` go?

anthonydillon:
Good question, `devel` goes to `/devel/en/getting-started`.

PR #2737, opened 2018-05-31
Review of metadata page 

### LINKED ISSUE #2736: metadata reference scrambled
On https://docs.jujucharms.com/2.3/en/authors-charm-metadata

"A metadata file defines the charm's relations, and whether it's designed for deployment as a subordinate service." is out of place. It is talking about relations, but is under storage. This might just be a missing header, since there is no actual relation section.

~~"Endpoint names are strings and must not match existing relation names from the Provides, Requires, or Peers metadata sections. The values beside each endpoint name must be left out (i.e. "foo": <anything> is invalid)." would also belong under the missing relation section, but is found under Extra bindings. Looks like someone started an update with the newer Endpoint terminology, or a mismerge.~~ (use of endpoint confused me, as it is a term we are now using for relations and normally had just used 'binds' instead of 'bind endpoints')

"Other available fields are:" doesn't belong under the Extra Bindings section, but instead an other/misc/stuff-that-doesn't-fit-elsewhere section.

pmatulis:
resolves #2736

### DIFF
--- a/src/en/authors-charm-metadata.md
+++ b/src/en/authors-charm-metadata.md
@@ -14,10 +14,11 @@ Every charm should have these fields declared:
 
   - `name`: The charm name, which is used to form the charm URL.
     - The following criteria are applied:
-          - Contains only characters `a-z`, `0-9`, and `-`
+          - Contains only characters `a-z`, `0-9`, and `-` (hyphen)
           - Must start with `a-z`
-          - Must not end with a `-`
-          - May only end with digits if the digits are _not_ directly preceded by a space.
+          - Must not end with a `-` (hyphen)
+	  - May only end with digits if the digits are _not_ directly preceded
+	    by a space.
     - Simple examples: 'foo' and 'foo-bar-baz'.
   - `summary`: A one-line description of the charm.
   - `description`: A longer description of the charm and its features. It will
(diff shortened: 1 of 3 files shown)


src/en/authors-charm-metadata.md
```diff
@@ -3,22 +3,27 @@ Title: Charm metadata
...
 
+Every charm should have these fields declared:
 
-Here's a valid metadata file:
+  - `name`: The charm name, which is used to form the charm URL.
+    - The following criteria are applied:
+          - Contains only characters `a-z`, `0-9`, and `-`
+          - Must start with `a-z`
+          - Must not end with a `-`
+          - May only end with digits if the digits are _not_ directly preceded by a space.
```
degville: to me, those hyphens are little difficult to spot at the end of the line. Maybe put `(hyphen)` or similar after them?

src/en/authors-charm-metadata.md
```diff
@@ -99,21 +111,11 @@ storage:
...
-    will participate in.
-  - if the charm is subordinate, it must contain at least one `requires`
-    relation with container scope.
+See developer [Storage][developer-storage] documentation for more information.
 
-## Resources
+## Resources field
 
-`resources` allows you to add blobs that your charm can utilize.
+The `resources` field allows is used to add blobs that your charm can utilize.
```
degville: typo, ~~allows~~, also *utilise*?

src/en/authors-charm-metadata.md
pmatulis: :+1:

src/en/authors-charm-metadata.md
pmatulis: :+1:

PR #2755, opened 2018-06-06
Review charms-scaling.md

### LINKED ISSUE #2526: Review charms-scaling.md
As per TODO notes.

pmatulis:
resolves #2526

### DIFF
--- a/src/en/charms-destroy.md
+++ b/src/en/charms-destroy.md
@@ -42,18 +42,22 @@ It is possible to remove individual units instead of the entire application
 (i.e. all the units):
 
 ```bash
-juju remove-unit mediawiki/1
+juju remove-unit postgresql/2
 ```
+
+In the case that the removed unit is the only one running on the corresponding
+machine, unless any of the following is true for that machine, it will also be
+removed:
+
+ - it was created with `juju add-machine`
+ - it is not being used as the only controller
+ - it is not hosting Juju-managed containers (KVM guests or LXD containers) 
 
 To remove multiple units:
 
 ```bash
-juju remove-unit mediawiki/1 mediawiki/2 mediawiki/3 mysql/2
+juju remove-unit mediawiki/1 mediawiki/3 mediawiki/5 mysql/2
 ```
-
-In the case that these are the only units running on a machine, unless that 
-machine was created manually with `juju add machine`, the machine will also be 
-removed.
 
 The `--destroy-storage` option is available for this command as it is for the
 `remove-application` command above.
--- a/src/en/charms-scaling.md
+++ b/src/en/charms-scaling.md
@@ -1,70 +1,68 @@
 Title: Scaling applications
-TODO:  The Scaling down section should reference charms-destroy.html (remove-unit)
-       Critical: review required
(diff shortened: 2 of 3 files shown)


src/en/charms-destroy.md
```diff
@@ -42,19 +42,23 @@ It is possible to remove individual units instead of the entire application
 (i.e. all the units):
 
 ```bash
-juju remove-unit mediawiki/1
+juju remove-unit postgresql/2
 ```
 
+In the case that the removed unit is the only one running on the corresponding
+machine, unless any of the following is true for that machine, it will also be
+removed:
```
degville: I get a little confused half-way through. Maybe something more like "the corresponding machine will be removed unless" ...

src/en/charms-scaling.md
```diff
@@ -1,88 +1,68 @@
...
 
 # Scaling applications
 
-One of the killer features of computing in the cloud is that it (should)
-seamlessly allow you to scale up or down your applications to meet your needs
-and whims. Juju not only makes it simple to deploy applications, but crucially
-makes it easy to manage them too. It won't anticipate you getting slashdotted
-or on the front page of hacker news (yet), but it does mean that when you do
-you can reliably scale your applications to meet the demand.
+The capability of a service to adjust its resource footprint to a level
```
degville: It makes me a little sad that the hyperbole in the first couple of sentences has gone, but I can see why  you've cut it for a dryer explanation.

src/en/charms-scaling.md
```diff
@@ -1,88 +1,68 @@
...
+appropriate for fulfilling client demands placed upon it is known as
+*scalability*. Scaling *vertically* affects the resources of existing machines
+(memory, CPU, disk space) whereas scaling *horizontally* involves the number of
+units available.
 
+Scaling with Juju amounts to horizontal scaling where "scaling up" adds units
```
degville: I think this is a little unclear after a sentence where unit scaling was an example of horizontal, although I know what you mean. Maybe something more like 'Juju abstracts horizontal scaling, where adding and removing units becomes scaling up and down within the Juju environment', or something much better.

src/en/charms-scaling.md
```diff
@@ -163,49 +139,46 @@ assigned an ID of '3'.
...
 
-For example, the following scales down the MediaWiki application by removing a
-specific unit:
+Note that because this is the only unit running on the underlying machine, the
+machine will also be removed.
 
-```bash
-juju remove-unit mediawiki/1
-```
+A machine can be manually removed unless any of the following is true for that
```
degville: I find it difficult to follow these positive and negative assertions, *can be manually removed*, *unless any of the following*, *is true*. As above, could this be untwisted?

src/en/charms-destroy.md
pmatulis: :ok:

src/en/charms-scaling.md
pmatulis: This sentence was more of a reinforcement of the previous sentence so I just removed it.

src/en/charms-scaling.md
pmatulis: :disappointed:

PR #2778, opened 2018-06-14
Change d-b required version to 1.5.3

pmatulis:
This is to make ./run work.

@WillMoggridge 

Does this just go to master?


pmatulis:
I'm pretty sure this is superseded by #2781.

evilnick:
yes, I think we need 1.6 anyhow. It would be cool if someone published the snap

PR #2806, opened 2018-06-27
Clarifying cloud pages

### LINKED ISSUE #2772: Straighten out adding clouds manually, especially MAAS
Page [clouds-maas-manual.md](https://docs.jujucharms.com/2.3/en/clouds-maas-manual) is too MAAS-centric. It should read that it is providing just one example of manually adding a cloud. It should also be reconciled somehow with section [Manually specifying additional clouds](https://docs.jujucharms.com/2.3/en/clouds#manually-specifying-additional-clouds) on the clouds.md page. Page clouds-maas.md also has a section named [Manually adding/defining a MAAS cloud](https://docs.jujucharms.com/2.3/en/clouds-maas#manually-defining-maas-clouds) that points to the manual page. The other cloud pages do not do this and makes MAAS seem very special.

In the end, how we present clouds and how we refer to adding them manually requires attention.

pmatulis:
resolves #2772 

Drive-by attempt to rectify clouds-lxd-resources.md not being built (see https://github.com/canonical-webteam/documentation-builder/issues/122).


src/en/clouds-manual.md
```diff
@@ -1,7 +1,7 @@
-Title: Adding a manual cloud
+Title: Using the Manual cloud with Juju
```
evilnick: I don't know why you changed this to 'the' manual cloud. It makes it sound like some sort of entity. We don't say 'the MAAS cloud'. The original phrasing wasn't great either. Maybe 'Creating a manual cloud to use with Juju' ?

src/en/clouds.md
```diff
@@ -311,81 +320,136 @@ expand the relevant section). You can also generate a YAML file.
...
+juju add-cloud devmaas maas-clouds.yaml
+juju add-cloud prodmaas maas-clouds.yaml
+```
+
+You must now add a credential for this cloud prior to creating a controller
+(`juju bootstrap`). See the [Credentials][credentials] page for details. The
+[Using MAAS with Juju][clouds-maas-add-credentials] page also includes
+MAAS-specific guidance on this.
+
+#### Manually adding an OpenStack cloud
```
evilnick: This section should very definitely include a link to the how to for setting up YOUR OWN openstack cloud. Perhaps in the form of a note at the beginning (These instructions assume an Openstack cloud which already has relevant Ubuntu images available. For help in configuring your own Openstack cloud to work with Juju, see...) or something

src/en/clouds-manual.md
pmatulis: Another terminology fail. We are left with the generic term of "manual" to work with. We don't say "the MAAS cloud" because MAAS is a proper noun used as an adjective, but "manual" is not a proper noun. So "a manual cloud" is devoid of meaning when used on its own and can even be construed as a cloud one has built manually (say with OpenStack). Also, this type of cloud has been flagged as "special" so using the term in a generic way is contradictory. By saying "the Manual cloud" we get around all this. One could argue that "*a* Manual cloud" could suffice but some may see it as a typographical error.

src/en/clouds.md
pmatulis: Agreed.

evilnick: did you do this?

pmatulis: Not yet.

pmatulis:
I'm done. BTW, this works now: https://docs.staging.jujucharms.com/2.4/en/clouds-lxd-resources

PR #2809, opened 2018-06-29
[2.4] Update Release Notes

pmatulis:
resolves #2625

PR #2811, opened 2018-06-29
Add What' s new 2.4

evilnick:


pmatulis:
You probably should add the new four new proxy options. My [PR 2809](#2809) vulgarises it.

PR #2828, opened 2018-07-02
Rewrite cahrms.reactive debugging instructions

evilnick:
Add several more bold bits saying this needs to be run within 'debug-hooks'

### DIFF
--- a/src/en/developer-debugging.md
+++ b/src/en/developer-debugging.md
@@ -170,21 +170,43 @@ for more information on the automatic retry feature and how to disable this
 behaviour. Note that option `--no-retry` can be used to prevent re-execution of
 failed hooks on the unit.
 
-## Debugging reactive Charms
+## Debugging reactive charms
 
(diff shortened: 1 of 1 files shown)


src/en/developer-debugging.md
```diff
@@ -170,21 +170,43 @@ for more information on the automatic retry feature and how to disable this
...
+
+!!! Tip: 
+    If essential hooks such as `install` and `config-changed` are not in the
+    `hooks` directory of your Charm, you forgot to include `layer:basic` in
+    your `layer.yaml` file.*
 
 ### charms.reactive command
 
-charms.reactive has a commandline tool that can help you debugging Charms. You can use it to view, set and clear flags in a debug-hooks session. *Charms built with older versions of the reactive framework might require you to use `state` instead of `flag`.*
+There is a commandline charms.reactive tool that can help with debugging charms.
```
degville: My only suggestion is that, as you're touching this doc, it may be worth adding a space to commandline. It's written like this only in one other place in the Juju dev docs (`reference-environment-variables.md`).

evilnick: Fair enough

PR #2781, opened 2018-06-18
Update documentation-builder to 1.6.0

Caleb-Ellis:
## Done

- Update documentation-builder to 1.6.0
- Update metadata.yaml with favicon, logo url, and navigation styles

## QA

- Pull code
- Either change branch name to `devel`, or add whatever the name of the branch is to versions.txt
- Check the new styling has been applied

PR #2840, opened 2018-07-05
Revert "Update documentation-builder to 1.6.0"

evilnick:
Reverts juju/docs#2781


evilnick:
I believe this may be what is causing the build failures. works locally, but not on jenkins

PR #2841, opened 2018-07-05
Revert "Revert "Update documentation-builder to 1.6.0""

evilnick:
Reverts juju/docs#2840


evilnick:
this doesn't seem to be it

PR #2854, opened 2018-07-05
Add new requirement (controller to provider)

### LINKED ISSUE #2592: Cloud Provider not Mentioned as Required by Controller (offline guide)
In the table here https://jujucharms.com/docs/2.3/charms-offline-config#internet-based-resources cloud provider should be listed as required by a juju controller.

For example this is how a squid proxy log looks like if http_proxy is set during bootstrap:

1522799633.515     16 10.10.101.75 TCP_MISS/200 606 GET http://10.10.101.2/MAAS/api/2.0/version/ - HIER_DIRECT/10.10.101.2 application/json

This means that the controller actually uses HTTP proxy to access the cloud provider (MAAS in this case)

pmatulis:
resolves #2592

### DIFF
--- a/src/en/charms-offline-config.md
+++ b/src/en/charms-offline-config.md
@@ -1,5 +1,4 @@
 Title: Configuring Juju for offline usage
-TODO:  for localhost cloud and controller being on the same host as the client, it's possible to use LXD remotely. worth mentioning?
 
 # Configuring Juju for offline usage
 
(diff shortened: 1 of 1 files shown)


src/en/charms-offline-config.md
```diff
@@ -50,7 +49,10 @@ and the appropriate Juju configuration changes (covered
    The **client** requires access to the backing cloud in order to create a
    controller. Most public clouds have a RESTful API that operates over TCP
    port 443. A special case is the localhost cloud, in which the client talks
-   to the local LXD daemon.
+   to the local LXD daemon.  
+
+    Depending on the provider chosen and the network design the **controller**
```
degville: there may be a wayward space before *Depending*...

pmatulis: It's intended. It's how to get that paragraph indented with the upper paragraph.

PR #2863, opened 2018-07-06
Improve UX clouds and credentials

### LINKED ISSUE #2797: Improve UX for moving from clouds to credentials
There is a definite need for improvement here - nothing too serious but users can be guided better. Minor corrections are also needed. LXD certificates can be explained better.

pmatulis:
resolves #2797

My information regarding variables and files supported by `autoload-credentials` was taken from the CLI help.


pmatulis:
@degville 
@evilnick 

:hear_no_evil: :see_no_evil: :speak_no_evil:
This PR is scheduled for a merge-without-review (MWR).

PR #2864, opened 2018-07-07
Move juju availability zone variable

### LINKED ISSUE #2595: JUJU_AVAILABILITY_ZONE variable is not just for developers
As per https://bugs.launchpad.net/juju/+bug/1759737, JUJU_AVAILABILITY_ZONE is a valid environment variable and is not limited to development use only.

Please update https://jujucharms.com/docs/stable/reference-environment-variables.
At the very list, it should not be in the list at the bottom of the page.

pmatulis:
resolves #2595


src/en/reference-environment-variables.md
```diff
@@ -1,4 +1,4 @@
-Title: Juju environment variables  
+Title: Juju environment variables | Reference
```
evilnick: why? By which I mean No

PR #2933, opened 2018-07-30
LXD constraints

### LINKED ISSUE #2833: Add LXD constraints
There are now LXD constraints. These act differently than traditional constraints in that they impose maximums instead of minimums. This is due to the way LXD works under the hood. This will require careful consideration on how to intermingle these two types of constraints.

As per https://github.com/juju/juju/pull/8891

pmatulis:
resolves #2833

### DIFF
--- a/src/en/charms-constraints.md
+++ b/src/en/charms-constraints.md
@@ -13,6 +13,8 @@ definitive constraint resource is found on the
 
 Several noteworthy constraint characteristics:
 
+ - Whenever a new machine is spawned (with commands `bootstrap`, `deploy`,
+   `add-unit`, or `add-machine`) a constraint can be specified.
  - Some constraints are only supported by certain clouds.
  - Changes to constraint defaults do not affect existing machines.
  - Multiple constraints are logically AND'd (i.e. the machine must satisfy all
@@ -23,7 +25,7 @@ Several noteworthy constraint characteristics:
 The idealized use case is that of stipulating a constraint when deploying an
 application and the backing cloud providing a machine with those exact
 resources. In the majority of cases, however, default constraints may have been
-set (at various levels) and the cloud is unable to supply those exact
+set (at various levels) and the cloud may be unable to supply those exact
 resources.
 
 When the backing cloud is unable to precisely satisfy a constraint, the
@@ -31,24 +33,22 @@ resulting system's resources will exceed the constraint-defined minimum.
 However, if the cloud cannot satisfy a constraint at all then an error will be
 emitted and a machine will not be provisioned.
 
-### Constraints and the LXD cloud
+### Constraints and LXD containers
 
-For the 'lxd' cloud type (`v.2.4.1`) things are different. With this cloud,
-constraints are interpreted as resource maximums as opposed to minimums. In the
-absence of constraints, a machine (container) will, by default, have access to
-**all** of the underlying system's (LXD host) resources.
+Constraints can be applied to LXD containers (`v.2.4.1`) either when they're
+running directly upon a LXD cloud type or when hosted on a Juju machine
+(residing on any cloud type). **However, with containers, constraints are
+interpreted as resource maximums as opposed to minimums.**
 
-This cloud type also honours instance constraints where instance type names
-from either [AWS][aws-types-kirkland], [Azure][azure-types-kirkland], or
-[GCE][gce-types-kirkland] can be used (e.g. AWS type 't2.micro' maps to 1 CPU
-and 1 GiB of memory). When used in combination with specific CPU/MEM
-constraints the latter values will override corresponding instance type values.
+In the absence of constraints, a container will, by default, have access to
+**all** of the underlying system's resources.
 
-!!! Warning:
-    Due to the different meaning of constraints in the context of a LXD
-    cloud, care must be taken when using commands in which the cloud type is
-    not referenced.
-    
+LXD constraints also honour instance type names from either
+[AWS][aws-types-kirkland], [Azure][azure-types-kirkland], or
+[GCE][gce-types-kirkland] (e.g. AWS type 't2.micro' maps to 1 CPU and 1 GiB of
+memory). When used in combination with specific CPU/MEM constraints the latter
+values will override corresponding instance type values.
+
 ## Constraint scopes, defaults, and precedence
 
 Constraints can be applied to various levels or scopes. Defaults can be set on
--- a/src/en/charms-constraints.md
+++ b/src/en/charms-constraints.md
@@ -196,6 +196,13 @@ GiB of memory:
 juju deploy postgresql --constraints "instance-type=c5.large mem=3.5G"
 ```
 
+To deploy Zookeeper to a new LXD container (on a new machine) limited by 5 GiB
+of memory and 2 CPUs:
+
+```bash
+juju deploy zookeeper --constraints "mem=5G cores=2" --to lxd
+```
+
 An application's current constraints are displayed with the `get-constraints`
 command:
  
--- a/src/en/charms-constraints.md
+++ b/src/en/charms-constraints.md
@@ -38,11 +38,11 @@ constraints are interpreted as resource maximums as opposed to minimums. In the
 absence of constraints, a machine (container) will, by default, have access to
 **all** of the underlying system's (LXD host) resources.
 
-This cloud type also honours instance constraints where 
-[AWS instance type][aws-instance-types-kirkland] names are used (e.g.
-'t2.micro' maps to 1 CPU and 1 GiB of memory). When used in combination with
-specific CPU/MEM constraints the latter values will override corresponding
-instance type values.
+This cloud type also honours instance constraints where instance type names
+from either [AWS][aws-types-kirkland], [Azure][azure-types-kirkland], or
+[GCE][gce-types-kirkland] can be used (e.g. AWS type 't2.micro' maps to 1 CPU
+and 1 GiB of memory). When used in combination with specific CPU/MEM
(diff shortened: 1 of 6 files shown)


pmatulis:
[Docs - test and write on LXD constraints and instance types (2.4)](https://trello.com/c/lmHeX5pt/411-docs-test-and-write-on-lxd-constraints-and-instance-types-24)

src/en/charms-constraints.md
```diff
@@ -31,12 +31,24 @@ resulting system's resources will exceed the constraint-defined minimum.
...
-below).
+### Constraints and the LXD cloud
 
+For the 'lxd' cloud type (`v.2.4.1`) things are different. With this cloud,
+constraints are interpreted as resource maximums as opposed to minimums. In the
+absence of constraints, a machine (container) will, by default, have access to
+**all** of the underlying system's (LXD host) resources.
+
+This cloud type also honours instance constraints where 
+[AWS instance type][aws-instance-types-kirkland] names are used (e.g.
```
manadart: Instance types from Azure and GCE are recognised as well.
The full list is here: https://github.com/dustinkirkland/instance-type/tree/master/yaml

src/en/reference-constraints.md
```diff
@@ -102,7 +102,7 @@ addresses the situation.
 - Valid values: instance-type=[defined on the cloud]
 
 ### LXD:
-- Unsupported: [cpu-cores, cpu-power, instance-type, tags, virt-type]
+- Unsupported: [cpu-cores, cpu-power, tags, virt-type]
```
manadart: You will get a warning with _cpu-cores_, as it is a deprecated alias of _cores_; but technically it is supported.

src/en/charms-constraints.md
mitechie: Does LXD pull this as a dep? Can we link to lxd vs a former employee's GH?

src/en/charms-constraints.md
```diff
@@ -31,12 +31,24 @@ resulting system's resources will exceed the constraint-defined minimum.
...
+'t2.micro' maps to 1 CPU and 1 GiB of memory). When used in combination with
+specific CPU/MEM constraints the latter values will override corresponding
+instance type values.
+
+!!! Warning:
+    Due to the different meaning of constraints in the context of a LXD
+    cloud, care must be taken when using commands in which the cloud type is
+    not called out.
```
mitechie: cloud type is not called out? What are you warning of here?

src/en/charms-constraints.md
```diff
@@ -114,22 +126,13 @@ See [Creating a controller][controllers-creating] for more guidance.
...
 
-For the localhost cloud, the following invocation will achieve a similar goal
-to the previous command (assuming that the LXD containers are using the
-'default' LXD profile):
+For the LXD cloud, the following invocation will place a **limit** of 2GiB of
+memory for each machine:
 
 ```bash
-lxc profile set default limits.memory 4GB
+juju bootstrap --constraints mem=2G localhost
```
mitechie: doesn't this need to be a --boostrap constraint? If we're just demonstrating a constraint on a workload let's show a deploy vs a bootstrap.

src/en/charms-constraints.md
```diff
@@ -139,6 +142,13 @@ that model, with the `set-model-constraints` command:
 juju set-model-constraints mem=4G
 ```
 
+For the LXD cloud, all new machines in the current model will be limited
+to an instance type of 'c5.large' (2 CPU and 4 GiB):
+
+```bash
+juju set-model-constraints instance-type=c5.xlarge
```
mitechie: lxd users have the instance type available to them but it's a 1% use case in our experience. Let's demo the hard cpu/etc settings in examples unless talking about the instance-type feature.

src/en/charms-constraints.md
```diff
@@ -179,6 +189,15 @@ model or application level):
 juju deploy apache2 --constraints "mem=4G cores=" 
 ```
 
+For the LXD cloud, we deploy PostgreSQL using a combination of an instance type
+and a specific CPU constraint. Instance 'c5.large' maps to 2 CPU and 4 GiB but
+the specific memory constraint of 3.5 GiB yields a machine with 2 CPUs and 3.5
+GiB of memory:
+
+```bash
+juju deploy postgresql --constraints "mem=3.5G instance-type=c5.large"
```
mitechie: since we're talking about things overriding I think the example would be more readable/grok-able to put the memory constraint after the instance-type so it'll read left->right as we speak of how it sorts out in your prose above.

src/en/charms-constraints.md
```diff
@@ -235,28 +254,19 @@ spaces, and not connected to either the 'storage' or 'dmz' spaces.
 
 See the [Network spaces][network-spaces] page for details on spaces.
 
-To get exactly two CPUs for a machine in a localhost cloud:
+To get a maximum of two CPUs for a machine in a LXD cloud:
```
mitechie: To create a machine with two CPUs for a machine in a LXD cloud:

src/en/charms-constraints.md
pmatulis: The fact that the given constraint in these example commands do entirely different things:

`juju deploy <charm> --constraints mem=4G`

and

`juju deploy <charm> --constraints mem=4G`

So one command is applied towards a LXD type cloud and one is applied towards a non-LXD type cloud. That is, the cloud is not "called out" (referenced).

src/en/charms-constraints.md
pmatulis: But a workload constraint is **not** being demonstrated here.

src/en/charms-constraints.md
mitechie: Right but since there's nothing in the command "called out". I can see saying something about "which cloud you're using" or something. But called out feels like I'm looking for a cloud type and it's not there.

src/en/charms-constraints.md
pmatulis: I don't follow. This **is** showing the instance-type feature.

src/en/reference-constraints.md
mitechie: let's go ahead and make sure to update to just cores as we go then please.

src/en/charms-constraints.md
pmatulis: OK

src/en/charms-constraints.md
pmatulis: How about

"For a LXD cloud, to create a machine with two CPUs:"

Repeating "a machine" is awkward.

src/en/charms-constraints.md
pmatulis: I'll reword.

src/en/charms-constraints.md
mitechie: This seems to be highlighting using set-model-constraints and not explaining how instance types is a valid constraint key to be used in all clouds.

src/en/charms-constraints.md
mitechie: fair enough on the double machine. I think we can let go of the "maximum" and calling out lxd apart somewhere through here. It's worth a note but I think it'd be worth letting go of it and treating them as common vocabulary through the doc.

src/en/charms-constraints.md
pmatulis: You compelled me to check `--constraints` behaviour. Turns out it does *not* set constraints for all models ('controller', 'default', and *all future models*), just 'controller' and 'default'. So I'll need to reword this entire section. IMO, `--constraints` should just be retired from the Juju code base as all it does more than `bootstrap-constraints` is add constraints to the 'default' model.

src/en/charms-constraints.md
pmatulis: I'm merely showing how to set the instance-type constraint for a LXD model. That's all.

Also, the instance-type constraint is not supported by all clouds.

src/en/reference-constraints.md
pmatulis: The charms-bundles.md page contained 'cpu-cores' and was updated.

pmatulis:
I'm done and will merge this tomorrow barring any more comments.

szeestraten:
A small comment from the side after chatting with @manadart in #juju earlier today. I feel charm-constraints.md could use some clarification and examples on how constraints work for lxd containers deployed in non lxd clouds such as aws or maas. Currently it reads more like this only is a thing for the lxd cloud. I think that's an important use case for folks running multiple lxd containers on production clouds.

pmatulis:
@szeestraten Thanks. That's a great suggestion. Lemme ruminate.

szeestraten:
@pmatulis, looks good. Perhaps an example of colocating services with some constrained lxd containers on a single machine would be a handy addition to the `charms-bundles.md` as well?

I was thinking something like this following the same style as the current examples, but please change as you see fit.

```yaml
services:
  wordpress:
    charm: "cs:trusty/wordpress-5"
    num_units: 1
    constraints:
      cores=1
      mem=1024
    annotations:
      "gui-x": "339.5"
      "gui-y": "-171"
    to:
    - lxd:0
  mysql:
    charm: "cs:trusty/mysql-57"
    num_units: 1
    constraints:
      cores=2
      mem=2048
    annotations:
      "gui-x": "79.5"
      "gui-y": "-142"
    to:
    - lxd:0
relations:
  - - "wordpress:db"
    - "mysql:db"
machines:
  "0":
    series: xenial
    constraints: "arch=amd64 cores=4 mem=4096"
```

pmatulis:
@szeestraten Sweet. :lollipop: I should be able to get that in pretty quickly.

pmatulis:
I'm done and will merge this in an hour.

PR #2958, opened 2018-08-08
Update the local base template and build from that

anthonydillon:
## Done
Copied the template from documentation-builder to the local base.tpl. Hardcoded some aspects of the template for Juju. Made the build command use the local template instead of the documentation builder one.

## QA
- Run `./run`
- Open http://0.0.0.0:8205/ 
- Check that the favicon works (the root issue)
- Check that the docs still look amaz-balls


nottrobin:
LGTM, works perfectly :+1:

PR #2973, opened 2018-08-22
Adding some more content to the charm promulgation page

hyperbolic2346:
Trying to add more content to this page to address questions and problems I had in my first attempt at promulgation.

### DIFF
--- a/src/en/charm-promulgation.md
+++ b/src/en/charm-promulgation.md
@@ -2,6 +2,20 @@ Title: Charm promulgation
 
 # Charm promulgation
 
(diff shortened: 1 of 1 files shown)


pmatulis:
This page is not for providing instruction on how to promulgate a charm or what promulgation means. A charm author who has finished their charm is here:

https://docs.jujucharms.com/stable/en/developer-getting-started#next-steps

Granted that the edited page should state better what its purpose is and provide a link to the above.

evilnick:
the linked page from the  linked page you mention on promulgation is wrong anyhow... Perhaps these should be combined in some meaningful way that people can follow

PR #3001, opened 2018-09-06
GCE credentials and review

### LINKED ISSUE #2968: GCE credential instructions are out of date and misses some things
The page at https://docs.jujucharms.com/2.4/en/help-google is now slightly out of date with the website. 

It also now needs to advise enabling the IAM API, which is required for some charms to work properly.

Rather than just regenerate all the steps for the latest iteration of the website, I suggest it might be better/easier/more durable to recommend using the  [cloud sdk](https://cloud.google.com/sdk/) tools from google to do this. I suspect there will be more longevity in the command-line tools than the website design, and it would be easier to test if it required updating.

pmatulis:
resolves #2968 

This has turned into a beast. Possibly a sub-page. But maybe it's fine.

I removed some sample command outputs to keep the page from getting too long.

Due to the immense change I chose to write this on a new page for now (clouds-gce.md).

One important piece that needs clarification is how to tell GCP what project to use by default. With `gcloud` I see a certain project in the "default configuration" but tests show that a different project ends up getting used on GCP.

### DIFF
--- a/src/en/clouds-gce.md
+++ b/src/en/clouds-gce.md
@@ -30,21 +30,25 @@ Although Juju knows how GCE works, there are a few tasks you must perform in
 order to integrate your account with Juju. We give an overview of the steps
 here:
 
- - Using the CLI tool
+ - Using the CLI tools
  - Assigning permissions
+ - Managing service accounts
  - Gathering credential information
 
 !!! Note:
-    The Google Cloud Platform Console (web UI) can also be used to complete the
-    above steps.
+    The Google [Cloud Platform Console][google-cpc] (web UI) can also be used
+    to complete the above steps.
 
-### Using the CLI tool
+### Using the CLI tools
 
-We show how to use the [Cloud SDK tools][gcloud-docs] from Google to manage
-your GCP (Google Cloud Platform) account. In particular, they will allow you to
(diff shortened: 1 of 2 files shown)


src/en/clouds-gce.md
```diff
@@ -0,0 +1,333 @@
...
+## Adding credentials
+
+The [Cloud credentials][credentials] page offers a full treatment of credential
+management.
+
+In order to access Google GCE, you will need to add credentials to Juju. This
+can be done in one of three ways (as shown below).
+
+Alternately, you can use your credentials with [Juju as a Service][jaas], where
+charms can be deployed using a web GUI.
```
evilnick: I think JAAS is more than just a GUI, maybe mention you get a free controller too

pmatulis:
[Rewrite Google cloud credentials and review page](https://trello.com/c/oyDb7o7O/543-rewrite-google-cloud-credentials-and-review-page)

pmatulis:
I figured out the problem stated earlier. I'll need to make some adjustments accordingly.

src/en/clouds-gce.md
pmatulis: :ok:

PR #3026, opened 2018-09-11
Limit getting-started redirect to 1.25

WillMoggridge:
Prevent the `/` and `/index` from redirecting to `/get-started`, except for 1.25 docs which don't have an index home page.
This removes the getting started page from the `/stable` redirects as well

http://localhost:8205/2.4/en/ should not redirect

http://localhost:8205/1.25/en/ should redirect to http://localhost:8205/2.4/en/getting-started

### DIFF
--- a/redirects.map
+++ b/redirects.map
@@ -1,21 +1,24 @@
-~^/(|en/?)?$ /2.4/en/getting-started;
-~^/devel/en/?$ /devel/en/getting-started;
+~^/(|en/?)?$ /2.4/en/;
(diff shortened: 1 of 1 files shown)


pmatulis:
Thanks.

pmatulis:
@WillMoggridge 

> http://localhost:8205/2.4/en/ should not redirect

Confirmed.

> http://localhost:8205/1.25/en/ should redirect to http://localhost:8205/2.4/en/getting-started

This ends up going to:

http://localhost:8205/1.25/en/getting-started

PR #3102, opened 2018-10-10
Share charms with everyone before promulgating

nobuto-m:
Users may want to share a charm with everyone for testing, then
promulgate it as the recommended charm. Fix the stream of the document
and clarify that promulgation is not just to publish, but to publish the
charm as the recommended one in the charm store.


pmatulis:
Thanks Nobuto.

PR #3119, opened 2018-10-22
Review credentials page and add command `set-credential`

### LINKED ISSUE #3117: [2.5] Document `set-credential` command

pmatulis:
resolves #3117

### DIFF
--- a/src/en/credentials.md
+++ b/src/en/credentials.md
@@ -13,57 +13,51 @@ We use the term *credentials* to describe the material necessary to do this
 represented by a *credential name* that is used to refer to those credentials
 in subsequent commands.
 
-Juju selects a credential according to how many credentials are defined. If you
-have only one credential, or if a credential is labelled 'default', then this
-is the credential that will be used by Juju. When multiple credentials are
-defined, with no default, a credential name must be specified at the model
-level.
+When credentials are added for a given cloud they become available to use on
+that cloud's controller and models. There are therefore two categories of
+credentials: those that are available (local) and those that have been sent to
+a controller (remote) due to requiring a credential for the first time.
 
+An active credential (one that has been used to bring about a change in Juju)
+is always associated with a cloud, a Juju user, and a model.
+ 
 ## Adding credentials
 
 Juju supports three methods for adding credentials:
 
- - Accepting credentials provided interactively by the user on the command line
- - Scanning for existing credentials via environment variables and/or "rc"
+ - accepting credentials provided interactively by the user on the command line
+ - scanning for existing credentials via environment variables and/or "rc"
    files (only supported by certain providers)
- - Reading a user-provided [YAML-formatted][yaml] file
+ - reading a user-provided [YAML-formatted][yaml] file
   
 !!! Note:
-    LXD deployments are a special case. Accessed from a Juju admin user, they
-    do not require credentials. Accessed from a non-admin user, a *certificate
-    credential* is needed. See
+    Local LXD deployments are a special case. Accessed from a Juju admin user,
+    they do not require credentials. Accessed from a non-admin user, a
+    *certificate credential* is needed. See
     [Additional LXD resources][clouds-lxd-resources-non-admin-creds] for
     details. 
 
 ### Adding credentials interactively
 
-You can add credentials interactively in this way:
+You can add credentials interactively. To do so with the AWS cloud:
 
 ```bash
-juju add-credential <cloud>
+juju add-credential aws
 ```
 
-You will be asked for credential information based on the chosen cloud. Here
-we're adding credentials for cloud 'aws':
+You will be asked for credential information based on the chosen cloud. For our
+AWS cloud the resulting interactive session would look like:
 
 ```no-highlight
 Enter credential name: carol
 Using auth-type "access-key".
 Enter access-key: *******
 Enter secret-key: *******
-Credentials added for cloud aws.
 ```
 
-If you eventually set multiple credential names for the same cloud you will
-need to set one as the default:
-
-```bash
-juju set-default-credential <cloud> <credential-name>
-```
-
-The default credential will be used when creating a controller with the
-`bootstrap` command. Otherwise, a credential can be specified with the
-`--credential` option with both the `bootstrap` and `add-model` commands.
+If you end up adding multiple credentials for the same cloud you will need to
+set one as the default. See below section
+[Setting default credentials][#setting-default-credentials].
 
 ### Adding credentials from environment variables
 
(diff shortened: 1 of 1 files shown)


pmatulis:
[Review credentials page and add `set-credential` command](https://trello.com/c/S6pD5gqG/725-review-credentials-page-and-add-set-credential-command)

src/en/credentials.md
```diff
@@ -13,57 +13,51 @@ We use the term *credentials* to describe the material necessary to do this
...
 
-Juju selects a credential according to how many credentials are defined. If you
-have only one credential, or if a credential is labelled 'default', then this
-is the credential that will be used by Juju. When multiple credentials are
-defined, with no default, a credential name must be specified at the model
-level.
-
+When credentials are added for a given cloud they become available to use on
+that cloud's controller and models. There are therefore two categories of
+credentials: those that are available (local) and those that have been sent to
```
anastasiamac: 'sent' or 'uploaded'?

src/en/credentials.md
```diff
@@ -13,57 +13,51 @@ We use the term *credentials* to describe the material necessary to do this
...
-need to set one as the default:
-
-```bash
-juju set-default-credential <cloud> <credential-name>
 ```
 
-The default credential will be used when creating a controller with the
-`bootstrap` command. Otherwise, a credential can be specified with the
-`--credential` option with both the `bootstrap` and `add-model` commands.
+If you end up adding multiple credentials for the same cloud you will need to
```
anastasiamac: This is a bit trickier although very close to the truth...

The rule for selecting a credential is:
1. If there is only one, obviously use that one;
2. if there is more than one, select first one (I am *pretty* sure that this is not happening now but some older versions of Juju might still do that... so dunno if you want to test/mention it);
3. If there is more than one credential than a default one needs to be specified and juju will always use that default one unless the credential name is used explicitly, for example in 'add-model --credential'

src/en/credentials.md
```diff
@@ -213,20 +207,48 @@ juju add-credential azure -f mycreds.yaml
     
 ## Managing credentials
 
-There are several management tasks that can be done related to credentials.
+Here we cover the management of credentials.
 
-### Listing credentials
+ - [Setting default credentials][#setting-default-credentials]
```
anastasiamac: At this stage, I think that this command only operates on local credentials. It's in the works to ensure that users can set default on the remote ones too..

src/en/credentials.md
```diff
@@ -213,20 +207,48 @@ juju add-credential azure -f mycreds.yaml
...
+juju set-default-credential aws carol
+```
 
-You can display what credentials are available by running the command:
+If only one credential exists for a cloud, it becomes the effective default
+credential for that cloud.
+
+Setting a default affects operations that require a credential such as creating
+a controller (`bootstrap`) and adding a model (`add-model`). It does not change
+what is currently in use (on a controller). With both these commands a
```
anastasiamac: Thank you! The lack of these being stated explicit was/is the cause of a lot of pain for users.

src/en/credentials.md
```diff
@@ -239,13 +261,13 @@ aws     bob*, carol
 google  wayne
 ```
 
-The asterisk '*' denotes the default credential, which will be used for the
-named cloud unless another is specified.
+The asterisk denotes the default credential, which will be used for the named
```
anastasiamac: Not too sure what 'named' here means... It really is 'the default for that cloud' I think.

src/en/credentials.md
```diff
@@ -262,9 +284,10 @@ local-credentials:
 Notice how the output says 'local-credentials', meaning they are stored on
 the local Juju client.
 
-#### In use
+### Listing remote credentials
 
-To see what credentials are in use by a model (here the 'default' model):
+To see what credentials are in use by a model (here the 'default' model) the
```
anastasiamac: Plural? not singular? 'what credential is in use'? A model can only have one credential...

src/en/credentials.md
```diff
@@ -341,8 +348,9 @@ Here we decided to use the file 'mycreds.yaml' from a previous example:
 juju add-credential aws -f mycreds.yaml --replace
 ```
 
-This will overwrite existing credential information, so make sure all current
-credentials are contained in the file, not just the new or changed one.
+This will overwrite **all** existing credentials for that cloud, so make sure
```
anastasiamac: I am not sure that it will overwrite *all* credentials. I am pretty sure it only updates the one?...

src/en/credentials.md
```diff
@@ -417,18 +425,31 @@ Credentials:
...
+The `remove-credential` command is used to remove a local credential (i.e. not
+cached on a controller):
 
 ```bash
 juju remove-credential aws bob
 ```
 
+### Setting a remote credential for a model
+
+To set what remote credential should be used for a model the `set-credential`
```
anastasiamac: The command is named 'set' but the functionality that it does is really more along the lines of 'change' or 'replace'.

src/en/credentials.md
```diff
@@ -417,18 +425,31 @@ Credentials:
...
+command (`v.2.5.0`) is available. For instance, to have remote credential 'bob'
+be used for model 'trinity' (for cloud 'aws'):
+
+```bash
+juju set-credential -m trinity aws bob
+```
+
+!!! Note:
```
anastasiamac: Maybe worthwhile mentioning that the user has to be either a controller admin or a model admin in order to be able to replace model's credential?

src/en/credentials.md
pmatulis: uploaded :)

src/en/credentials.md
pmatulis: I only tested this with the beta.

src/en/credentials.md
pmatulis: Will reword.

src/en/credentials.md
pmatulis: Nicely seen. Will reword. But this intrigues me. If a model is shared with another user isn't it reasonable to expect that user to want to add their own backing cloud authentication material (e.g. their AWS account) to use it? Why the limitation of one credential per model?

src/en/credentials.md
pmatulis: Ohh nice. Thank you.

src/en/credentials.md
pmatulis: Will reword. BTW, isn't the functionality of `update-credential` contained within this new command?

src/en/credentials.md
pmatulis: Alright.

src/en/credentials.md
anastasiamac: No, 'set-credential' will upload a credential from local store (client) if it does not exist remotely on the controller. It will not update the content of remotely stored credential, only 'update-credential' command can do that.

src/en/credentials.md
anastasiamac: Credential per model. This means that all communications with the cloud that this model does is linked to one credential, and potentially, is billed to one account.

If a user wants to use their own credential, they can either create another model specifying --credential; or now use 'set-credential'.

Eventually a credential can be shared between users. But this is another story and a bit further in the future.

src/en/credentials.md
```diff
@@ -13,57 +13,51 @@ We use the term *credentials* to describe the material necessary to do this
...
    files (only supported by certain providers)
- - Reading a user-provided [YAML-formatted][yaml] file
+ - reading a user-provided [YAML-formatted][yaml] file
   
 !!! Note:
-    LXD deployments are a special case. Accessed from a Juju admin user, they
-    do not require credentials. Accessed from a non-admin user, a *certificate
-    credential* is needed. See
+    Local LXD deployments are a special case. Accessed from a Juju admin user,
+    they do not require credentials. Accessed from a non-admin user, a
```
SimonRichardson: This isn't true is it, even local LXD deployments require credentials, the "localhost" credentials are stored inside the users local `credentials.yaml`. In fact they have an expiry upon them (ten years from initial creation).

pmatulis: Right. Here I'm letting the user know they don't need to *add* a credential. I'll reword a bit.

hmlanigan: Agreed, but the user isn't required to setup those credentials, unlike remote LXD and other clouds.

SimonRichardson: No, but they should be maybe aware there is an expiry?

pmatulis: I added the expiry.

PR #3140, opened 2018-10-31
SLA and budgeting documentation for Juju experts.

cmars:


cmars:
@buggtb Please let me know what you think. Anything you'd like to change or add? Thanks!

buggtb:
lovely job @cmars !

src/en/experts-sla.md
```diff
@@ -0,0 +1,162 @@
...
+[Juju experts](https://jujucharms.com/experts) may designate a Juju
+model as a managed, supported solution by setting an SLA.
+
+There are three support levels that may be set on a model:
+
+- Essential
+- Standard
+- Advanced
+
+## Canonical support
```
pmatulis: This section will probably need to be rewritten (I can do that). I've asked around and I have more information on how UA fits in.

As is, this text suggests that Canonical UA will triage support requests but I believe they should go directly to the partner right? If needed, that partner can escalate to UA for both Juju and non-Juju issues (but probably not for issues related to the build itself).

It also seems to suggest that those customers have access to the UA knowledgebase. How can that be?

Why do you treat the UA essential plan and the standard & advanced plans as different? They all seem to offer the same thing in this context.

cmars: This PR actually documents features that have been quietly available in Juju 2.x for private beta testing, but up until now, were not publicly documented. We're adding now in order to better support Juju experts who will be launching some managed solutions very soon.

> This section will probably need to be rewritten (I can do that). I've asked around and I have more information on how UA fits in.

Much appreciated!

> 
> As is, this text suggests that Canonical UA will triage support requests but I believe they should go directly to the partner right? If needed, that partner can escalate to UA for both Juju and non-Juju issues (but probably not for issues related to the build itself).

Apologies, I was mistaken on this point, I can update the wording.

> 
> It also seems to suggest that those customers have access to the UA knowledgebase. How can that be?
> 
> Why do you treat the UA essential plan and the standard & advanced plans as different? They all seem to offer the same thing in this context.

I'd like to improve this by linking to https://www.ubuntu.com/support/plans-and-pricing#server for an authoritative description of UA, rather than duplicate it here (where it might become stale/outdated/inaccurate).

cmars: @pmatulis Updated, PTAL, thanks!

pmatulis: @cmars I used `juju sla` and there is a login that happens. I think we need some background on this aspect. The Charm Store (or JAAS?) is used as an authentication server that ties into a contract that the authenticated user may have with a partner? Is there a different way to log in with Juju?

cmars: @pmatulis That is correct. The user who runs the `juju sla` command will be prompted to authenticate with Ubuntu SSO. Currently, the experts are running this command on behalf of their users, but we want to show how it all works. It's the same SSO login flow that users are prompted with in order to push charms to the charm store or agree to terms.

pmatulis:
@cmars I'm going to merge this now. I will follow it up with a PR on my side. I'll ping you for a final review.

PR #3151, opened 2018-11-02
Add floatingip option

lolwww:
Adding description for external-network and floating-ip config options during controller bootstrap on top of openstack


pmatulis:
Thanks. :lollipop:

PR #3162, opened 2018-11-06
Add SLA (Experts) support

pmatulis:


src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
-Title: Managed solutions under SLA
+Title: Managed solutions
```
mitechie: Should this be Juju Experts as that's the branding of the offering?

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
 
-# Managed solutions under SLA
+# Managed solutions
 
-[Juju experts](https://jujucharms.com/experts) may designate a Juju
-model as a managed, supported solution by setting an SLA.
+A Juju Expert is a partner backed by professional support at Canonical Ltd.
+Such a partner deploys and supports an infrastructure based on Juju. We call
```
mitechie: (remove an)

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
+this a *managed solution*.
```
mitechie: ok, I see the managed solution bit.

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
 
-There are three support levels that may be set on a model:
+Any support requests beyond the build, workload, and operation of a particular
+solution can be escalated to Canonical by means of their Ubuntu Advantage
```
mitechie: so this is written to the end user who might work with an expert. Does the expert do the escalation or the end user?

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
+support framework. Prime examples of such requests are for issues relating to
+Juju itself, Ubuntu-specific software, or the operating system kernel. The
+escalated support case will be treated by Canonical at the same support level
+currently ascribed to the partner. Hence, support level names used by both a
+Juju Expert and by Canonical are the same:
```
mitechie: I'd just state that there's a common set of support levels and not get into the customer vs the canonical one. Especially since they're common as part of the expert program it's like telling the user "they could be different words for this, but we've made sure both parties use the same words"

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
+information on Canonical support levels.
 
-Juju experts will provide details on the solution-specific support provided
-under each SLA, with budget estimates.
+The Juju Expert will provide details such as support levels and costs on a
+per-solution basis but ultimately an SLA will be conferred upon a specific Juju
+model (see [Models][models]).
 
-## Setting an SLA on a model
+Consult the external [Juju Experts][juju-experts] page to find a managed
```
mitechie: I'd remove "external" here.

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
-billed for more than this amount. If metered costs exceed this amount, the
-budget limit will need to be increased in order to continue to access support.
+## Requirements
 
-Set an Essential SLA on the current model _marketdata_, with a monthly budget
-limit of $100 USD:
+The following requirements need to be met before setting an SLA for the first
+time:
+
+ 1. The Juju operator will need an [Ubuntu SSO][ubuntu-sso] account.
```
mitechie: but is the operator the expert or the end user?

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
+Support costs will not be billed for more than the budget limit. If metered
+costs exceed this amount, the limit will need to be increased in order for
+support to remain active.
+
+For example, to set an Essential SLA on model 'default', with a monthly budget
+limit of USD 100:
 
 ```bash
-juju sla essential --budget 100
+juju sla -m default essential --budget 100
```
mitechie: can we use a different model name vs default for the example. It's almost certainly not "default" in any real case and might confuse folks.

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
 ```
 
+```no-highlight
+Model                   SLA                                                        Message
+lxd:admin/default       essential       the new sla will become effective on 2018-11-05
```
mitechie: Same here, I'd change lxd to something more intersting. Maybe use an expert company name or something.

src/en/experts-sla.md
```diff
@@ -1,149 +1,257 @@
...
+for (e.g. budget limit exceeded, problem with payment method). The 'Message'
+column will be updated with appropriate details.
+
+Determining factors for actual metering costs (e.g. fixed per-hour rate,
+per-unit hour, etc.) are negotiated between you and your Expert.
+
+## Creating a support case
+
+When it comes time to request help from your Expert you can file a support case
+using the same URL for requesting a credit limit increase:
```
mitechie: I'd not tie the url to the credit limit. Just show it as the correct url for someone looking to file an issue.

src/en/experts-sla.md
pmatulis: The Expert. I will specify.

src/en/experts-sla.md
pmatulis: Hmm.

src/en/experts-sla.md
pmatulis: This is fairly innovative and non-intuitive and I wanted to dispel any confusion by making it transparent.

src/en/experts-sla.md
pmatulis: :ok:

src/en/experts-sla.md
pmatulis: The term "Juju operator" is what the documentation uses to refer to the person who operates Juju. It is distinguished from a "Juju user" (`add-user`). I can just say "you" if that is clearer. We do not use the term "end user" in the documentation as it is too vague.

src/en/experts-sla.md
pmatulis: This was deliberate and is often used throughout the documentation. The base documentation should not be using exotic names when introducing new topics. Tutorials and blogs and documented use cases are better for that.

src/en/experts-sla.md
pmatulis: See above.

src/en/experts-sla.md
pmatulis: OK, but I thought it would be simpler on the user to know that there is just one URL and not a bunch to keep track of.

pmatulis:
> Much of it reads that the expert is managing the Juju for you

@mitechie 
Where do you see that?

src/en/experts-sla.md
pmatulis: I don't think so. The software allows for "managed solutions". It does not manage the Juju Experts program.

src/en/experts-sla.md
pmatulis: :white_check_mark:

pmatulis:
[add SLA (Experts) support](https://trello.com/c/SstnWzUE/757-add-sla-experts-support)

src/en/experts-sla.md
mitechie: Hmm? You support infrastructure, deploy infrastructure, manage infrastructure. I don't really think there's a case where you prefix it with an.

src/en/experts-sla.md
mitechie: So if this is the expert why put it here in end user documentation? They're not going to do the escalating and it's in the Experts area of wrapping things for the end user. This seems like sharing implementation details. We can just say we support the experts and you can trust them.

src/en/experts-sla.md
mitechie: Sorry, maybe my question is more is that if the Expert is the Juju operator I'd not put it here. Again, it's more details for end users they don't need to know or worry about. Just that they have a way to get expertise and Canonical has their back. They're trustworthy and we're attaching our name/trust to them.

src/en/experts-sla.md
mitechie: k

src/en/experts-sla.md
mitechie: Right, but if they come looking for a URL for X they're looking for X and anything not X is just noise to have to filter through.

src/en/experts-sla.md
pmatulis: We don't know. They can very well be looking for a solution in both camps. It's better to be transparent so the customer will be in a better position to take a decision.

src/en/experts-sla.md
pmatulis: The Juju operator is the customer. Everything in this document is for the customer, except the bit about escalation being done by the Expert.

src/en/experts-sla.md
pmatulis: It's not a big issue for me. I'll remove the extra words.

src/en/experts-sla.md
pmatulis: :ok:

src/en/experts-sla.md
pmatulis: Knowing that the Expert can escalate to Canonical, but only for certain things, is a huge part of the deal. More transparency.

mitechie:
I spoke with Casey and talked through the concerns and he sees my confusion between the two segments. He expressed there's a bit of a desire for the Experts to have something that shows potential customers "here's what we're doing for you". 

I think if we can take this page and basically split it in half we'll be ok. The top part focuses on what's an Expert, how they can help you solve problems, etc. The second half is very much "what the expert is doing" and so should impress an end user but not be step by step for the user as they're not going to be doing those steps at this point. 

With this I'd like to use headings and for us to look at the two parts in their own unique views. The intended audience of the page is meant to be the end user, but we should not hint/suggest end users run the commands the experts are doing for them.

pmatulis:
@cmars 
@mitechie 

I'm good to go once my remaining question is answered:

Would the Expert also open support cases for the customer? It seems so since the original authentication (to set SLA) via SSO was done by the Expert.

PR #3222, opened 2018-12-06
Add `diff-bundle` command

pmatulis:


pmatulis:
[Document `diff-bundle` command](https://trello.com/c/UX9WAReg/882-document-diff-bundle-command)

PR #3224, opened 2018-12-06
Add run-action on leader feature; quick format review

### LINKED ISSUE #3223: [2.5] Run action on leader
As per https://github.com/juju/juju/pull/9440.

pmatulis:
resolves #3223

### DIFF
--- a/src/en/actions.md
+++ b/src/en/actions.md
@@ -1,13 +1,14 @@
-Title: Juju actions
+Title: Working with actions
 
-# Juju Actions
+# Working with actions
 
 Juju charms can describe *actions* that users can take on deployed applications.
 
 Actions are scripts that can be triggered on a unit via the command line.
-Parameters for an action are passed as a map, either defined in a YAML file
-or given through the UI, and are validated against the schema defined in
-actions.yaml. See [Actions for the charm author][charmactions] for more information.
+Parameters for an action are passed as a map, either defined in a YAML file or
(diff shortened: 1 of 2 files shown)


src/en/actions.md
```diff
@@ -70,25 +70,27 @@ add-repo:
...
+
+### `run-action`
 
-### `juju run-action`
+Triggers an action.
 
-Trigger an action. This command takes a unit (or multiple units) as an argument
-and returns an ID for the action. The ID can be used with `juju
-show-action-output <ID>` or `juju show-action-status <ID>`.
+This command takes a unit (or multiple units) as an argument and returns an ID
```
mitechie: I feel like the /leader should be noted here as basically it's another format for specifying "which unit". While we talking about taking a unit/multiple having the leader option seems to fit here.

src/en/actions.md
```diff
@@ -102,10 +104,17 @@ application). So if there were two git units you can also do:
 juju run-action git/0 git/1 add-repo repo=myproject
 ```
 
-When running short-lived actions from the command line, it is more convenient to
-add the `--wait` option to this command. This causes the Juju client to wait for
-the action to run, and then return the results and other information in YAML
-format.
+To run an action on an application leader append the string "/leader" to the
```
mitechie: ah, I guess it's in this section. I read the diff a bit wrong. Thanks for the update.

PR #3355, opened 2019-02-05
Document distinction between remove, destroy, kill

### LINKED ISSUE #3353: Document the meaning and difference between remove-, destroy-, and kill- prefixes
These three prefixes have very precise and useful meanings in Juju, but the distinction is not immediately obvious from an outside perspective.   They should be called out explicitly on the Concepts and Terms page.

We might also want to call out list- vs show- and the fact that there are common aliases for the list- commands.

johnsca:
Resolves #3353

### DIFF
--- a/src/en/juju-concepts.md
+++ b/src/en/juju-concepts.md
@@ -107,6 +107,33 @@ the agents have been omitted:
 
 ![units][img__units]
 
+## Remove vs Destroy vs Kill
+
+Several commands in Juju have a prefix of `remove-`, `destroy-`, or `kill-`.
+Each prefix has a specific meaning in Juju, which increase in extent or
+severity:
+
+* `remove-` commands, such as `remove-application`, `remove-relation`, or
+  `remove-unit`, affect a single logical entity **within** a model. They can
+  be destructive, but only to at most a single application. Some of these
+  commands can be disabled via `juju disable-command remove-object`.
+
+* `destroy-model` and `destroy-controller` are destructive commands that
+  affect the entire model or even the entire controller. Generally, these
+  will refuse to operate on models which have persistent storage in them or
+  controllers which have active models in them. This can be overridden with
+  the `--destroy-storage` or `--destroy-all-models` flags, respectively. These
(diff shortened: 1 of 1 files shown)


pmatulis:
This is good information but I will be putting most into a tutorial on "Basic client usage". I don't feel the concepts page is the right place for command types.

johnsca:
I thought the Concepts and Terms page was a good place because the terms "remove", "destroy", and "kill" seemed like they had specific meaning, but maybe it's more about a grouping of commands.  I'd like someone from the Juju team to chime in on this.  @mitechie @wallyworld ?

mitechie:
I agree with Peter in that this is the one section that's very CLI specific. This is about explaining the entities/nouns that we use in Juju and I think in a CLI usage might be a way to explain some of the different in these verbs/commands. It might be a good transition for a CLI or UX concepts/etc that goes into the expectations of the UX and how we have these nouns and common actions on them like list, add, remove/destroy, etc.

johnsca:
Would it make sense, then to remove the CLI specific bits and document the more general meaning of these three terms, or are they intrinsically tied to the CLI commands that use those terms?  If the latter, then I can go ahead and close this PR in favor of the tutorial.

pmatulis:
The stuff I've been working on is here:

https://github.com/juju/docs/pull/3356/files#diff-db859208243b3ee33f7ebd8020b23be1

johnsca:
@pmatulis Can you provide feedback on why you closed this?  I think the reworked wording addressed the concerns about it being CLI focused while still addressing the need to make clear how the otherwise confusing terminology is used.

pmatulis:
@johnsca Yeah, it was closed by mistake when I merged my other PR. I like what you did with this. I'm still a little concerned though. The concepts page deals in "objects" whereas you are explaining actions. But I'm at a loss right now where to put it. Lemme think about it.

pmatulis:
@johnsca I found a place for this. The [Removing things](https://docs.jujucharms.com/2.5/en/charms-destroy) page. I will merge this here and I'll move it over afterwards.

PR #3356, opened 2019-02-05
Merge lxd tutorial with lxd clouds page, spawn basic client usage tutorial

### LINKED ISSUE #3321: Review LXD tutorial
This needs an update (versions in output, etc) as well as rewording the introduction and moving to the lxd snap. Rename file from `tut-lxd.md` to `tutorial-lxd.md`. Adjust inter-linking accordingly. Put in HTML redirects (`redirects.yaml`). Point to this page from the bottom of a few pages. Consider a Tutorials page in the menu where all tutorials can be linked from.

Now I'm wondering how this tutorial will be much different from the LXD clouds page.

pmatulis:
resolves #3355 
resolves #3321


src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
+Title: Basic client usage - tutorial
+TODO:  Warning: Ubuntu release versions hardcoded
+       Include links to planned tutorials (tutorial-bundles.md, tutorial-constraints.md)
+
+# Basic client usage - tutorial
```
mitechie: I'd remove the "Client" and just go with Basic Usage. Maybe CLI usage since it's specific to that client in particular.

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+reading suggestions are included at the end.
+
+## Prerequisites
+
+The following prerequisites are assumed as a starting point for this tutorial:
+
+ - You're using Ubuntu 18.04 LTS (Bionic).
+ - Juju (stable snap channel) is installed. See the [Installing Juju][install]
+   page.
+ - You have chosen a backing cloud and have created a controller for it. Refer
```
mitechie: I don't think we should introduce the term 'backing cloud'. I'd suggest just going with a 'cloud' or if we need to clarify the cloud in some way 'cloud for your model' or something along that line.

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+   to the [Clouds][clouds] page to get started with a cloud and controller.
+ 
+This guide uses LXD (`v.3.9`) as a backing cloud due to its accessibility and
```
mitechie: s/backing//

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+low resource usage. Choose a local LXD cloud if you're not sure what to use
+(see [Using LXD with Juju][clouds-lxd]).
+
+## Command prefixes and aliases
+
+As you gain more experience with the client, you will encounter command
+prefixes whose meanings may not be immediately apparent.
```
mitechie: Starting off this section with a "you're not going to understand these" seems a bit of a harsh intro. I'd start out with the show, list, etc. Only then getting to remove and then noting the destroy vs remove vs kill (and honestly kill is only on a kill-controller so I'd leave it off of here and save it for the removing a controller section)

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+
+The `show-` prefix is used to drill down into an object to reveal details. It
+is akin to the verb "describe".
+
+There are command aliases, that typically employ a prefix, that point to their
+corresponding commands. This practice is for simplifying the life of the
+operator. The `list-` alias/prefix is a good example, where `list-clouds`
+points to `clouds`.
+
+The list of aliases can be obtained in this way:
```
mitechie: I think for aliases I'd just note that list-xxx can typically be shortened to just the entity e.g. list-clouds to clouds. Then note there are a few other aliases and that you can find them in the help commands. I think this paragraph is odd because it again starts off with exceptions vs just stating "there are list commands for the entities in Juju, you can normally shorten the command by leaving off the list because there are aliases setup for them. By the way there are other aliases out there, here's how to find them"

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+
+```no-highlight
+Controller: lxd
+
+Model       Cloud/Region         Status     Machines  Access  Last connection
+controller  localhost/localhost  available         1  admin   just now
+default*    localhost/localhost  available         2  admin   2019-02-02
+```
+
+To see the currently active controller, model, and user:
```
mitechie: if you're going to list out "currently active controller, model, and user" I'd suggest we note the answers to those in that same order so it aligns.

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+default, charms are downloaded from the online [Charm Store][charm-store]
+during deployment.
+
+To deploy a charm, such as 'redis', in the currently active model:
+
+```bash
+juju deploy redis
+```
+
+This causes a similarly-named application ("redis") to be installed on a
```
mitechie: I don't think we want to have this appear happen stance with "similarly-named" but note more explicitly "by default the application will be named after the charm"

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Adding a machine
+
+To request that machine be created that is devoid of an application:
+
+```bash
+juju add-machine -m alpha -n 2
+```
+
+Here we've used the `-m` option to explicitly select a model (i.e. we don't
```
mitechie: we also add in the -n flag without noting it in any way. Is it necessary? If so, should we call out that we're in fact adding two here?

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+juju deploy mongodb
+```
+
+### Scaling up an application
+
+To scale out the 'apache2' application by creating one or more instantiations
+(*units*) of it on new machines:
+
+```bash
+juju add-unit -n 1 apache2
```
mitechie: we don't need the -n of 1 for add-unit. It's just CLI noise imo.

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+0/lxd/0  started  10.72.88.191   juju-ded876-0-lxd-0  bionic      Container started
+1        started  10.243.67.53   juju-ded876-1        bionic      Running
+2        started  10.243.67.141  juju-ded876-2        bionic      Running
+```
+
+The output is broken up into four sections.
+
+The **top** section mentions basic information such as the names of the current
+model and controller ('alpha' and 'lxd' respectively), followed by the cloud
+name ('localhost'), what version is running on the agents in the current model,
```
mitechie: "what version is running on the agents in the current model" I'd suggest "what version of Juju the model is running"

src/en/tutorial-client-basic.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Removing a machine
+
+To remove the machine whose ID is '1':
+
+```bash
+juju remove-machine 1
+```
+
+Note that it is not possible to remove a machine (without the `--force` option)
```
mitechie: This reads that the recommended way to remove a machine with a unit on it is to go the --force route. I'd prefer we noted that if you remove a unit on a machine and it's the last unit on that machine the machine will automatically be cleaned up.

mitechie: which to be fair you note above in remove-application but this section still reads a bit wrong imo. I'd just leave out the note about the error perhaps or call out that if you have a unit on it to use remove-unit or application instead.

src/en/tutorial-client-basic.md
pmatulis: Well, "client" is a concept that is well defined and used throughout the documentation. I also think "Basic usage" is just too vague. I think "CLI" is not a well enough known acronym to the intended audience.

src/en/tutorial-client-basic.md
pmatulis: Ditto.

The reason why "backing cloud" was introduced (years ago) was that just "cloud" was considered too vague. A backing cloud is a cloud that is being used *with* Juju. "Your cloud" just sounds, again, vague in many contexts.

src/en/tutorial-client-basic.md
pmatulis: I didn't say the reader won't understand it. But OK, you want something more upbeat.

Right but there is `kill-controller` as well as `destroy-controller`, and 'destroy-*' was covered.

src/en/tutorial-client-basic.md
pmatulis: OK

src/en/tutorial-client-basic.md
pmatulis: OK, I missed that.

src/en/tutorial-client-basic.md
pmatulis: I thought I'd make it clearer using agents since it's hard to imagine "Juju running in the model". That's an abstraction. Dunno which is more confusing.

src/en/tutorial-client-basic.md
pmatulis: It's in the output?

src/en/tutorial-client-basic.md
pmatulis: I'll reword.

src/en/tutorial-client-basic.md
pmatulis: I'll go with the abstraction.

PR #3384, opened 2019-02-14
Remove media folders

nottrobin:
And link to assets.ubuntu.com for media.


pmatulis:
Do we lose the .svg files? These are useful for doing updates.

nottrobin:
@pmatulis: In discourse, images cannot be stored in a codebase, they need to be hosted somewhere. Whether that's in discourse.jujucharms.com or assets.ubuntu.com, you have no option but to re-upload the image.

The other option you have is to use a data-uri directly in the markdown, but I won't do this in this pr. You can feel free to go through and do that afterwards, or once it's in discourse, it's up to you.

nottrobin:
@pmatulis I've put back the `static/media` folder, as I hope that may allay your concerns. It's messy, but given this codebase will be legacy within the month (hopefully) it doesn't bother me too much,

PR #3459, opened 2019-03-12
Fixing deprecated env vars

joedborg:


pmatulis:
I don't feel we should use JUJU_REPOSITORY. It used to designate a directory for local charms but this has been deprecated since 2.0 afaict. It can still be used as a variable of course but this is confusing. In this context we should use something generic like JUJU_WORK_DIR, if anything at all.

There is also missing one of the more important variables: CHARM_BUILD_DIR.

joedborg:
Hey @pmatulis, thanks for the feedback.  Would `JUJU_WORK_DIR` and `CHARM_BULD_DIR` not need to be the same thing, in this instance?

evilnick:
I know we *say* JUJU_REPOSITORY is deprecated, but it is still referenced in the current docs

https://docs.jujucharms.com/2.5/en/reference-environment-variables#building

which reminds me that any changes here should also be replicated there

pmatulis:
@joedborg CHARM_BUILD_DIR has an effect on the system whereas JUJU_WORK_DIR is benign.

Issue #3461, opened 2019-03-18
GUI docs should state how to install GUI


mdavis-xyz:
I was looking at [this page](https://docs.jujucharms.com/2.5/en/controllers-gui).

> To view the URL and login credentials for the GUI, use the following command:
>
>     juju gui

But when I try to run that I see

```
$ juju gui
ERROR Juju GUI is not available: Juju GUI not found
```

The docs should say how to install the GUI. (Also the error message should say that too.)

The docs say:

> The GUI comes with every controller providing the controller was not created with the `--no-gui` option

I don't think that's true. It obviously didn't come with my controller, and I didn't use `--no-gui`.

mitechie:
What version of Juju are you using? It wasn't baked into 1.25 but from 2.0 onward it's build in and downloaded as part of boostrap.

hatched:
@mitechie is correct, however it's possible that there was an issue downloading the GUI on bootstrap and the error messaging around that might have been lacking.

If you'd like to get a GUI going on that controller you can try the following:
- Download the latest release https://github.com/juju/juju-gui/releases/download/2.14.1/jujugui-2.14.1.tar.bz2
- `juju upgrade-gui path/to/jujugui-2.14.1.tar.bz2`
- Wait a few moments after the upload completes for it to uncompress and start serving.
- `juju gui`

mlda065:
I'm using version 2.5.2

mlda065:
The bootstrap claimed to finish successfully. I saw no errors.

That said, I am unable to deploy anything successfully.

hatched:
Were you able to add a GUI using the above mentioned steps?

mlda065:
Yes. That worked. Now I'm writing a pull request to add those steps to the docs.

hatched:
Glad to hear it.

It sounds like there was an issue downloading the GUI during bootstrap, we'll have to make that more resilient to failure and to output some messaging if it fails.

mlda065:
Maybe I can try it again and paste the logs here?

How do I unbootstrap?

hatched:
https://docs.jujucharms.com/2.5/en/charms-destroy#destroying-controllers

PR #3462, opened 2019-03-19
add instructions on how to install GUI

mdavis-xyz:
Allegedly the GUI is installed by default since version 2.0. 
However that's not true. I've tried installing and reinstalling version 2.5 and the GUI is still absent.

So I've added instructions on how to install it, based on issue #3461

### DIFF
--- a/src/en/controllers-gui.md
+++ b/src/en/controllers-gui.md
@@ -27,8 +27,7 @@ Your login credential is:
   password: 1d191f0ef257a3fc3af6be0814f6f1b0
 ```
 
-From Juju 2.0 onwards, the GUI *should* be installed by default.
-However if you see the error `ERROR Juju GUI is not available: Juju GUI not found`, then:
+If you see the error `ERROR Juju GUI is not available: Juju GUI not found`, then:
 
    * Download the [latest release](https://github.com/juju/juju-gui/releases/) (`.tar.bz2` file)
    * `juju upgrade-gui path/to/release.tar.bz2`
--- a/src/en/controllers-gui.md
+++ b/src/en/controllers-gui.md
@@ -17,7 +17,7 @@ To view the URL and login credentials for the GUI, use the following command:
 juju gui
 ```
 
-This will produce output similar to the following:
+This should produce output similar to the following:
 
 ```no-highlight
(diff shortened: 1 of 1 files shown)


src/en/controllers-gui.md
```diff
@@ -17,7 +17,7 @@ To view the URL and login credentials for the GUI, use the following command:
 juju gui
 ```
 
-This will produce output similar to the following:
+This should produce output similar to the following:
```
hatched: As this is simply a messaging issue on the rare event bootstrap fails to download a GUI. I feel that this should remain as `will` and the issue be resolved in Juju.

src/en/controllers-gui.md
```diff
@@ -27,6 +27,14 @@ Your login credential is:
   password: 1d191f0ef257a3fc3af6be0814f6f1b0
 ```
 
+From Juju 2.0 onwards, the GUI *should* be installed by default.
+However if you see the error `ERROR Juju GUI is not available: Juju GUI not found`, then:
```
hatched: I'd remove line 30 entirely and modify this one to read...

"If the `juju gui` command returns the error `ERROR Juju GUI is not available: Juju GUI not found`, then:"

hatched (changes requested):
Thanks for the PR @mlda065 however as the issue is rare and caused by a bug I feel that this should be worded as a help to people running into the issue and not as a flaw in the tool.

src/en/controllers-gui.md
mlda065: Ok, I just updated that

pmatulis:
@mlda065 You've caught the Juju documentation on the cusp of significant change. Changes, as of today actually, are now made on discourse.jujucharms.com. See this [announcement](https://discourse.jujucharms.com/t/documentation-migrated-to-discourse/1274).

timClicks:
@hatched do you know if this patch is still relevant?

hatched:
The issue that caused this PR to be created is still technically possible however the GUI has now been replaced with the Dashboard in 2.8 so it'll need some further updates to that effect. 

https://github.com/canonical-web-and-design/jaas-dashboard/releases

PR #3497, opened 2019-10-03
Describe the root-disk-source constraint available for OpenStack

timClicks:
The `root-disk-source` constraint is currently undocumented for the OpenStack provider. This commit mentions it in the constraints reference and the openstack-cloud pages.

### DIFF
--- a/1097-using-openstack-with-juju.md
+++ b/1097-using-openstack-with-juju.md
@@ -184,6 +184,14 @@ juju bootstrap <cloud> <controller name> --config network=<network_id> \
 
 For a detailed explanation and examples of the `bootstrap` command see the [Creating a controller](/t/creating-a-controller/1108) page.
 
+## OpenStack-specific features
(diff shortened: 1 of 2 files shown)


babbageclunk:
Hey, this isn't quite right - the openstack provider only supports the literal value "volume" for `root-disk-source` so the disk is created in cinder. You can't specify any existing block device.

mitechie:
The documentation is now built from the discourse docs material. Editing the documentation here won't have any visible effect for the end users. 

It does look like you were tinkering there though so maybe I'm missing a step/gap. https://discourse.jujucharms.com/t/juju-constraints/1160

howbazaar:
@mitechie we are trialing using a git workflow to edit the discourse docs. We are using a tool that I originally wrote for Peter called discli, which gives a CLI to get and update posts. We have a way to update a branch with any changes.

PR #3509, opened 2020-07-01
Fix grammar; remove redundant word

timClicks:
Hat tip @setharnold


timClicks:
https://discourse.juju.is/t/using-lxd-with-juju/1093 updated (revision 9)

